PR #8, opened 2019-10-26
KOGITO-228 - Several fixes and improvements for management console

cristianonicolai:


srambach:
Nice progress on the front end! 

One thing to take a look at is the new Data Toolbar component. I think this is what we'll want to use on the process instance list. It's still "experimental" meaning for now, there could be breaking changes coming, but it would be great to start using it so that we can help to shape its functionality. 
https://www.patternfly.org/v4/documentation/react/experimental/datatoolbar/
It was demoed at the Patternfly community meeting today and when the recording comes out, I'll send it along. 

I also think we'll want to look at the layout on the details page in the near future, but for now this is looking good.
👍

srambach (approved):

Sara4994 (approved):

PR #15, opened 2019-11-05
Remove process diagram panel

cristianonicolai:


cristianonicolai:
jenkins please retest this

packages/management-console/src/components/Templates/ProcessDetailsPage/ProcessDetailsPage.tsx
```diff
@@ -60,13 +60,10 @@ const ProcessDetailsPage = ({match}) => {
...
                             </Breadcrumb>
                         </GridItem>
-                        <GridItem span={8}>
-                                <ProcessDetailsProcessDiagram/>
-                        </GridItem>
                         <GridItem span={4}>
                             <ProcessDetails loading={loading} data={data}/>
                         </GridItem>
-                        <GridItem span={8}>
+                        <GridItem span={4}>
```
srambach: To make this responsive, you should be able to use `className="pf-m-all-12-col pf-m-all-6-col-on-lg pf-m-all-4-col-on-xl"` on the `<Grid>` and then remove the span properties on the `<GridItem>`s with the content (not the title one). It's not perfect but unless the screen is pretty big, the 3 cards are pretty squished so this at least makes that a little better.

cristianonicolai: @srambach thanks for the tip, take a look if I got it right.

Sara4994 (approved):

PR #16, opened 2019-11-05
Styling fixes to process instance list and details pages

srambach:
Removes nav toggle, fixes card titles, adjusts breadcrumb, changes to Red Hat font.
Removes some extraneous inline styles


cristianonicolai:
@srambach we had conflicting changes in another PR, would you mind rebasing this?

srambach:
I rebased, but now some tests are failing and I'm not sure how to fix that.

Also, I learned that the responsive grid classes actually are implemented in the React component. So we can actually go back to how the grid was defined with props but add in additional props to make the grid responsive. I added this in the last commit.

srambach:
I updated the snapshots to make the tests pass; hopefully that was the right thing to do!

cristianonicolai (approved):

PR #18, opened 2019-11-07
KOGITO-548 - Fixed favicon

Sara4994:


packages/management-console/webpack.common.js
```diff
@@ -11,7 +12,10 @@ module.exports = {
   plugins: [
     new HtmlWebpackPlugin({
       template: path.resolve(__dirname, 'src', 'index.html')
```
cristianonicolai: @Sara4994 I found that all we need is to add the follwing:
```suggestion
      template: path.resolve(__dirname, 'src', 'index.html')
      favicon: 'src/favicon.ico'
    })
```

then rename the image to favicon.ico but keep it at the src level.
CopyWebpackPlugin will copy all images to image directory, which is not needed.

cristianonicolai (changes requested):

packages/management-console/webpack.common.js
Sara4994: @cristianonicolai Yes i tried this at first place, but still favicon wasnt showing up. So went with other aproach which worked fine.. anyway will try onemore time.

cristianonicolai: cool, ok, I got it working fine with this approach so let me know if doesnt work for you.

Sara4994: @cristianonicolai Modified.

cristianonicolai:
@Sara4994 I believe this PR could be closed.

PR #29, opened 2019-11-19
KOGITO-618 Changed logic for filtering data

AjayJagan:
- on Click of Apply Filter button the data is re-fetched every time.
- Initially the parentProcessInstanceId whose state is 'ACTIVE' is loaded, instead of loading the entire data

### DIFF
--- a/packages/management-console/server/server.js
+++ b/packages/management-console/server/server.js
@@ -1,5 +1,5 @@
 const express = require("express");
-const {ApolloServer, gql} = require("apollo-server-express");
+const { ApolloServer, gql } = require("apollo-server-express");
 
 // Construct a schema, using GraphQL schema language
 const typeDefs = gql`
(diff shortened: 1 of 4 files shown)


cristianonicolai:
@AjayJagan I noticed that now when I select a filter in the dropdown, it automatically updates the status panel on the right side. I guess that should only be update once I apply the filter? @srambach wdyt?

packages/management-console/server/server.js
```diff
@@ -730,7 +730,12 @@ const resolvers = {
                 console.log("data", datum.parentProcessInstanceId);
                 if (args["filter"].id) {
                   return (datum.id == args["filter"].id);
-                } else {
+                }
+                else if (args["filter"].parentProcessInstanceId[0] == null){
```
cristianonicolai: please align else statements
```suggestion
               } else if (args["filter"].parentProcessInstanceId[0] == null){
```

AjayJagan:
> @AjayJagan I noticed that now when I select a filter in the dropdown, it automatically updates the status panel on the right side. I guess that should only be update once I apply the filter? @srambach wdyt?

sure that can be done.

srambach:
Yes, I agree - the chip group should always indicate what filter is currently being applied to the list. So it wouldn't update until the button is pushed to apply the filter.
@cristianonicolai @AjayJagan

packages/management-console/server/server.js
AjayJagan: Should an else statement be addded?

cristianonicolai: its just the formatting that looks a bit odd.

AjayJagan:
> Yes, I agree - the chip group should always indicate what filter is currently being applied to the list. So it wouldn't update until the button is pushed to apply the filter.
> @cristianonicolai @AjayJagan

Yes. Now the chips would be shown only on click of the Apply Filter button.

packages/management-console/server/server.js
AjayJagan: I guess I will add the prettier soon :) . That should solve most of the formatting issue.

cristianonicolai: you could just fix these ones quite easily, the prettier formatting can be done later

cristianonicolai (approved):

Sara4994 (approved):

PR #30, opened 2019-11-20
KOGITO-580 - Task console bootstrap

Sara4994:


packages/task-console/src/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8" />
+    <title>Kogito - Management Console</title>
```
cristianonicolai:
```suggestion
    <title>Kogito - Task Console</title>
```

cristianonicolai (changes requested):
@Sara4994 please  also take a look in the build error

cristianonicolai:
@Sara4994 I wonder if we could reuse some of these config files. I belive most settings will be common anyway so we could share it. It would avoid inconsistencies in the future.

Jenkinsfile
```diff
@@ -29,7 +29,7 @@ pipeline {
                     script {
                         githubscm.checkoutIfExists('kogito-apps', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
                         wrap([$class: 'Xvnc', takeScreenshot: false, useXauthority: true]) {
-                            sh('yarn run init && yarn build:prod')
+                            sh('yarn run init && yarn build:prod && yarn run lint')
```
cristianonicolai: shouldnt lint be part of build:prod instead?

cristianonicolai (commented):
@Sara4994 could we also share the webpack files? seems like we can do import and these are not really different between apps, wdyt?

Jenkinsfile
Sara4994: @cristianonicolai moved lint as common one for both the apps. there is a script in package.json of lerna, which will check lint errors for the both application.

cristianonicolai (approved):

AjayJagan (approved):

PR #40, opened 2019-12-06
KOGITO-557 & KOGITO5-594 - Integrated datatoolbar from PF4 + refresh button

AjayJagan:
- added DataToolbar from PF4.
- added refresh button to refresh process instances.


packages/management-console/src/components/Molecules/DataToolbarComponent/DataToolbarComponent.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const onRefreshClick = () => {
+    if (isFilterClicked) {
+      filterClick(checkedArray);
+    }
+  };
+  const onStatusToggle = isExpandedItem => {
+    setisExpanded(isExpandedItem);
+  };
+  const statusMenuItems = [
+    <SelectOption key="=ACTIVE" value="ACTIVE" />,
```
cristianonicolai:
```suggestion
    <SelectOption key="ACTIVE" value="ACTIVE" />,
```

cristianonicolai (commented):
looks good, just a minor comment.

srambach:
This is looking great!
Two comments -
- I think that when you clear all the filters, you should see _all_ the items in the list.
- The number of filters updates when you click the checkbox, but if you don't actually apply the filter, then it doesn't match the results. Can you make it only update when the "apply filter" button is clicked?

A heads up - we may ask to change all the statuses to sentence case from all caps (Active vs ACTIVE). @lclay2 is looking at this. Is it better to decide and do this soon?

lclay2:
@srambach - I switched the status text to sentence case in the updated mockup, posted at: https://issues.redhat.com/browse/KOGITO-607. Looks fine, if not better, to me.

AjayJagan:
> This is looking great!
> Two comments -
> 
> * I think that when you clear all the filters, you should see _all_ the items in the list.
> * The number of filters updates when you click the checkbox, but if you don't actually apply the filter, then it doesn't match the results. Can you make it only update when the "apply filter" button is clicked?
> 
> A heads up - we may ask to change all the statuses to sentence case from all caps (Active vs ACTIVE). @lclay2 is looking at this. Is it better to decide and do this soon?

Hey Sarah, 
1) I have added the change to fetch all the instances when clear filters is clicked.
2) The number of filter updates is controlled by the PF component so I guess I have no control over it. If you see this  https://www.patternfly.org/v4/documentation/react/experimental/datatoolbar   it updates in a similar way.
3) I feel that 'Active' looks better than 'ACTIVE', So shall I change all the states to Sentence Cases?

srambach:
@AjayJagan @cristianonicolai So instead of the "Clear all filters" link, let's call it "Reset to default (active instances only)" and that restores the filters to just have Active checked and that's the results that are displayed. Showing nothing isn't going to be useful, so as long as Active is the most common use case, that should be good.

And for the numbers...I agree, that's how the component works. It's a little different because there's no "Apply filter" button there - the filter would apply instantly. But, since we still have the chips in sync with what's shown, I am ok with this as it is for now.

AjayJagan:
> @AjayJagan @cristianonicolai So instead of the "Clear all filters" link, let's call it "Reset to default (active instances only)" and that restores the filters to just have Active checked and that's the results that are displayed. Showing nothing isn't going to be useful, so as long as Active is the most common use case, that should be good.
> 
> And for the numbers...I agree, that's how the component works. It's a little different because there's no "Apply filter" button there - the filter would apply instantly. But, since we still have the chips in sync with what's shown, I am ok with this as it is for now.

Hey Sarah, The 'clear all filters' is a part of the Datatoolbar component. So is it possible for me to change the naming to 'Reset to default'. I was under the assumption that I cannot manipulate the components provided by PF.
datatoolbar

srambach:
Hi, @AjayJagan - Yes, you can change the text using the `clearFiltersButtonText` property. And you provide the callback function in `clearAllFilters` and that would just do whatever to reset back to just Active being shown. It is set up to only show that link if you have a callback and if there is a filter applied. (there is a small bug right now but that will be the behavior very soon!) So it might go away if the user unchecks everything I think, but I don't think it's a showstopper. I have the developer of that component sitting across from me :-) so if you have more questions I can connect you directly!

AjayJagan:
> Hi, @AjayJagan - Yes, you can change the text using the `clearFiltersButtonText` property. And you provide the callback function in `clearAllFilters` and that would just do whatever to reset back to just Active being shown. It is set up to only show that link if you have a callback and if there is a filter applied. (there is a small bug right now but that will be the behavior very soon!) So it might go away if the user unchecks everything I think, but I don't think it's a showstopper. I have the developer of that component sitting across from me :-) so if you have more questions I can connect you directly!

ahh! Sorry. Its my bad :(. I have done the changes accordingly . Please do have a look once the jenkins build is done. :)

cristianonicolai (approved):

Sara4994 (approved):

srambach (approved):
image

PR #41, opened 2019-12-06
Changes path to experimental spinner component

srambach:
I was unable to build because it could not find the spinner component. I am told this should be pointing to the js directory instead of the esm directory - do you agree?

@AjayJagan @cristianonicolai


AjayJagan:
> I was unable to build because it could not find the spinner component. I am told this should be pointing to the js directory instead of the esm directory - do you agree?
> 
> @AjayJagan @cristianonicolai

I guess @cristianonicolai  also faced the same issue. The patternfly version is out of date . So we have to do a "yarn upgrade" to resolve it.

srambach:
The upgrade fixed the problem.

PR #44, opened 2019-12-10
KOGITO-684 Make icon button for parent-child navigation

srambach:
This changes the parent/sub process instance buttons to icon buttons with 2 different icons.
https://issues.redhat.com/browse/KOGITO-684
NOTE: I've put in the tooltip intended for displaying the parent or child's name, but I don't know if this information is available. @AjayJagan @cristianonicolai can you let me know and/or whether it's something I can add here easily or if someone else will need to do some work to show it?

### DIFF
--- a/packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
+++ b/packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
@@ -8,10 +8,12 @@ import {
   FormGroup,
   Text,
   TextVariants,
-  Title
+  Title,
+  Tooltip
 } from '@patternfly/react-core';
 import React from 'react';
 import { Link } from 'react-router-dom';
+import { LevelDownAltIcon, LevelUpAltIcon } from '@patternfly/react-icons';
 
 const ProcessDetails = ({ loading, data }) => {
   return (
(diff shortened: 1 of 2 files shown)


srambach:
image

image

lclay2 (dismissed):
Looks good to me, thanks @srambach

AjayJagan:
> This changes the parent/sub process instance buttons to icon buttons with 2 different icons.
> https://issues.redhat.com/browse/KOGITO-684
> NOTE: I've put in the tooltip intended for displaying the parent or child's name, but I don't know if this information is available. @AjayJagan @cristianonicolai can you let me know and/or whether it's something I can add here easily or if someone else will need to do some work to show it?

To differentiate between the parentprocess and child process - we can use the parentProcessInstanceId(if it's null then it's a parent else it's a child).
This is just a suggestion - What if we can use the process instance name (like travels, hotel booking etc...) and on hover we can display the id. Wouldn't it have a cleaner look. WDYT @cristianonicolai  @srambach :)

cristianonicolai:
@srambach @AjayJagan I like the suggestion, same design but showing the name instead when you hover you can see the id, that is mostly to make sure we can show a difference between processes of same type, as the id is mostly irrevelevant.

lclay2:
good catch @cristianonicolai

AjayJagan (approved):

cristianonicolai (approved):

Sara4994 (approved):

PR #48, opened 2019-12-16
KOGITO-716 Prevent user from searching processes without any filter

AjayJagan:
added changes to show an alert when no status in selected from the drop down

### DIFF
--- a/packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
+++ b/packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
@@ -16,12 +16,25 @@ interface IOwnProps {
   iconType: string;
   title: string;
   body: string;
+  filterClick?: any;
+  setFilters?: any;
+  setCheckedArray?: any;
+  refetch?: any;
 }
 const EmptyStateComponent: React.FC<IOwnProps> = ({
   iconType,
   title,
-  body
+  body,
+  filterClick,
+  setFilters,
+  setCheckedArray,
+  refetch
 }) => {
+  const resetClick = () => {
+    filterClick(['ACTIVE']);
+    setFilters(['ACTIVE']);
+    setCheckedArray(['ACTIVE']);
+  };
   return (
     <Bullseye>
       <EmptyState>
@@ -34,11 +47,24 @@ const EmptyStateComponent: React.FC<IOwnProps> = ({
             className="pf-u-mb-xl"
           />
         )}
+        {iconType === 'warningTriangleIcon1' && (
+          <FontAwesomeIcon
+            icon={faExclamationTriangle}
+            size="3x"
+            color="var(--pf-global--warning-color--100)"
+            className="pf-u-mb-xl"
+          />
+        )}
         <Title size="lg">{title}</Title>
         <EmptyStateBody>{body}</EmptyStateBody>
         {iconType === 'warningTriangleIcon' && (
-          <Button variant="primary" onClick={() => window.location.reload()}>
+          <Button variant="primary" onClick={() => refetch()}>
             Refresh
+          </Button>
+        )}
+        {iconType === 'warningTriangleIcon1' && (
+          <Button variant="link" onClick={resetClick}>
+            Reset to default
           </Button>
         )}
       </EmptyState>
--- a/packages/management-console/src/components/Templates/DataListContainer/DataListContainer.tsx
+++ b/packages/management-console/src/components/Templates/DataListContainer/DataListContainer.tsx
@@ -15,12 +15,15 @@ import DataListTitleComponent from '../../Molecules/DataListTitleComponent/DataL
 import DataToolbarComponent from '../../Molecules/DataToolbarComponent/DataToolbarComponent';
 import './DataList.css';
 import DataListComponent from '../../Organisms/DataListComponent/DataListComponent';
+import EmptyStateComponent from '../../Atoms/EmptyStateComponent/EmptyStateComponent';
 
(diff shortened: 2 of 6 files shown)


packages/management-console/src/components/Templates/DataListContainer/DataListContainer.tsx
```diff
@@ -90,14 +93,23 @@ const DataListContainer: React.FC<{}> = () => {
...
+                  initData={initData}
+                  setInitData={setInitData}
+                  isLoading={isLoading}
+                  setIsError={setIsError}
+                />
+              ) : (
+                <EmptyStateComponent
+                  iconType="warningTriangleIcon1"
+                  title="No status is selected"
+                  body="Try selecting atleast one status to see results"
```
srambach: Typo here, and let's simplify the language slightly. "Select at least one status to see results."

srambach:
One other question - should the user also get the alert message when they've removed all filters by clicking the X on all the chips? It seems like that puts them into the same state - no status selected, so they won't see anything.

AjayJagan:
> One other question - should the user also get the alert message when they've removed all filters by clicking the X on all the chips? It seems like that puts them into the same state - no status selected, so they won't see anything.

I guess if the user clicks on the 'X' button in the chip , then no results found would occur. The 'No status selected' occurs when there are no status checked and the apply filter is clicked. I can change the working if needed :)

srambach:
@AjayJagan Yes - I also showed it to the other UX people yesterday and we think X'ing out of all the chips puts you in the same state as having no filters applied, so you should get the alert in that case also. This contrasts with having one or more of them checked, and not having any results. @cristianonicolai does that sound ok with you as well?

cristianonicolai:
@srambach +1 makes sense to me.

AjayJagan:
> @AjayJagan Yes - I also showed it to the other UX people yesterday and we think X'ing out of all the chips puts you in the same state as having no filters applied, so you should get the alert in that case also. This contrasts with having one or more of them checked, and not having any results. @cristianonicolai does that sound ok with you as well?

Done :+1:

srambach:
Behavior when removing all the chips and getting the alert looks good now. 
I swear I got into a weird state when clicking filters on and off but I can't reproduce it.
image
Although, I can reproduce that if I have "Active" showing, the one there has 4 subprocesses that are "Completed" but they don't show if I select Completed. Is that expected behavior?

AjayJagan:
> Behavior when removing all the chips and getting the alert looks good now.
> I swear I got into a weird state when clicking filters on and off but I can't reproduce it.
> image
> Although, I can reproduce that if I have "Active" showing, the one there has 4 subprocesses that are "Completed" but they don't show if I select Completed. Is that expected behavior?

I think the filters are for the parent process. So the completed ones that we are seeing are the sub processes. So they would not be shown. @cristianonicolai  please verify :)

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
```diff
@@ -34,13 +34,26 @@ const EmptyStateComponent: React.FC<IOwnProps> = ({
...
+        )}
         <Title size="lg">{title}</Title>
         <EmptyStateBody>{body}</EmptyStateBody>
         {iconType === 'warningTriangleIcon' && (
           <Button variant="primary" onClick={() => window.location.reload()}>
             Refresh
           </Button>
         )}
+        {iconType === 'warningTriangleIcon1' && (
+          <Button variant="primary" onClick={() => window.location.reload()}>
```
cristianonicolai: @srambach should we use variant="secondary" or even variant="link" in here?

@AjayJagan can we avoid reloading the entire apge here? I mean, we should be able to trigger a React action to go back to the initial state.

cristianonicolai (changes requested):
@srambach regarding the state of subprocesses, the current behaviour is expected. We only apply the filter to the root process. Sub process can, and will have differemt states that don't match to the filter that is applied. The focus is really on the root process.

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
srambach: Looks like `variant="link"` is the way to go, since there's already a primary button. Good catch.

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
```diff
@@ -34,13 +45,26 @@ const EmptyStateComponent: React.FC<IOwnProps> = ({
...
+            icon={faExclamationTriangle}
+            size="3x"
+            color="var(--pf-global--warning-color--100)"
+            className="pf-u-mb-xl"
+          />
+        )}
         <Title size="lg">{title}</Title>
         <EmptyStateBody>{body}</EmptyStateBody>
         {iconType === 'warningTriangleIcon' && (
           <Button variant="primary" onClick={() => window.location.reload()}>
```
cristianonicolai: Can you also fix the page reload here :)

AjayJagan: I thought since its a server connection issue, we need to reload the entire page.

AjayJagan: I have changed it to refetch() comming from the useQuery hook :)

cristianonicolai (approved):

Sara4994 (approved):

PR #49, opened 2019-12-18
KOGITO-689 Add abort process instance action for any active process

AjayJagan:
- added abort process button for active instances.
- added snapshot tests for the following components : 
     - Datatoolbar
     - DataList
     - DataListContainer
    - DataListItem


packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+          <DropdownItem
+            key={4}
+            onClick={() =>
+              handleAbortActiveInstances(
+                processInstanceData.id,
+                processInstanceData.processId
+              )
+            }
+          >
+            Abort instance
```
cristianonicolai:
```suggestion
            Abort
```

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+            key={3}
+            onClick={() =>
+              handleViewError(
+                processInstanceData.processId,
+                processInstanceData.id,
+                processInstanceData.endpoint
+              )
+            }
+          >
+            View Error
```
srambach: "View error" (lower case e)

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+        setAlertTitle('Error aborting process instances');
+        setAlertType('danger');
+        setAlertMessage(
+          `Aborting process instance failed with error - ${JSON.stringify(
+            error.message
+          )}`
+        );
+        setAlertVisible(true);
+      });
+  };
```
srambach: I think that these should be the inline variant of alert (if they are going to be shown within the list) I can't see how it looks if it's not the top one, so I'm not sure how well putting the alert inline works either.

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+        >
+          View Error
+        </DropdownItem>
+      ];
+    }
+  };
+  const handleAbortActiveInstances = (processId, processInstanceId) => {
+    axios
+      .delete(
+        `http://localhost:4000/management/processes/${processId}/instances/${processInstanceId}`
```
cristianonicolai:
```suggestion
        `${processInstanceData.endpoint}/management/processes/${processId}/instances/${processInstanceId}`
```

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+      )
+      .then(() => {
+        setAlertTitle('Process Instance Aborted');
```
cristianonicolai:
```suggestion
        setAlertTitle('Abort operation');
```

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+        setAlertType('success');
+        setAlertMessage(
+          `Process Instance - ${processId} is aborted successfully`
+        );
+        setAlertVisible(true);
+      })
+      .catch(error => {
+        setAlertTitle('Error aborting process instances');
```
cristianonicolai:
```suggestion
        setAlertTitle('Abort operation');
```

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -205,7 +205,139 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+  const handleAbortActiveInstances = (processId, processInstanceId) => {
+    axios
+      .delete(
+        `http://localhost:4000/management/processes/${processId}/instances/${processInstanceId}`
+      )
+      .then(() => {
+        setAlertTitle('Process Instance Aborted');
+        setAlertType('success');
+        setAlertMessage(
+          `Process Instance - ${processId} is aborted successfully`
```
cristianonicolai:
```suggestion
          `Process ${processId} execution has been aborted`
```
@srambach does that makes sense?

packages/management-console/src/components/Templates/ProcessDetailsPage/tests/ProcessDetailsPage.test.tsx
```diff
@@ -84,7 +84,7 @@ const mocks = [
 ];
 
 describe('Process Details Page component', () => {
-  it('Sample test case', async () => {
+  test.skip('Sample test case', async () => {
```
cristianonicolai: Is that skipping the test execution?

cristianonicolai (changes requested):
@AjayJagan sorry but I missed a few details about the rule to determine if a process instance can be aborted. To start with only processes that contain the addon _process-management_ can be aborted. If so, then processes on states Active, Error and Suspended can be aborted.

packages/management-console/src/components/Templates/ProcessDetailsPage/tests/ProcessDetailsPage.test.tsx
AjayJagan: yes I wanted to discuss this. The tests are working normally in the local. I have updated the snapshots and tested the component(actually ProcessDetailsPage has no relation to this PR).But jenkins failure happens in the ProcessDetailsPage tests . Any idea why this could happen

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
AjayJagan: > @AjayJagan sorry but I missed a few details about the rule to determine if a process instance can be aborted. To start with only processes that contain the addon _process-management_ can be aborted. If so, then processes on states Active, Error and Suspended can be aborted.

So all the processes containing the addon as process-management can be aborted.

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
AjayJagan: abortprocess
This is how inline style alert looks. Is this fine

packages/management-console/src/components/Templates/ProcessDetailsPage/tests/ProcessDetailsPage.test.tsx
AjayJagan: I think I figured out the issue .Will make the necessary changes

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
AjayJagan: so , processes('ACTIVE','ERROR','SUSPENDED') having addon-_process-management_ will have _Retry,Skip,Abort,View error_ in the dropdown. if addon- _process-management_ is not available then the drop down would show only _View error_.
The processes('ABORTED' and 'COMPLETED') would have the dropdown _**disabled**_

AjayJagan: is this assumption correct? Please correct me if I am wrong :)

packages/management-console/src/components/Templates/ProcessDetailsPage/tests/ProcessDetailsPage.test.tsx
AjayJagan: Resolved! :)

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
srambach: After talking with some interaction designers, we think the best thing to do is to use an error modal. (https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/modal and scroll down to Error Reporting)

This is likely the best approach for any error that happens as a result of the user selecting an action on the list items (process instances). We'll want to make sure the name or ID of the process is included in the error message. (no matter where we display it, that's a good idea)

Reasoning:
Inserting an error into the list is not a good idea; and having it show up at the top or above the list separates the error from where the action was taken. Adding some sort of error indication within the process instance row is confusing with the existing state, which can be "error." 

@AjayJagan @cristianonicolai Does that make sense and are there any problems implementing it that way?

cristianonicolai: @srambach yep, that should be fine, although we need to change it for the skip and retry actions as well. Also, when we add the bulk actions, we could have multiple errors, would we display all in the same popup?

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
cristianonicolai: addon contains process-management and state in Active, Error or Suspended

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
srambach: I think as long as the errors are going to happen close enough in time to report together, then yes, one modal for bulk actions. Each of the names/IDs that failed would be listed.

The alternative would be making them toast alerts but there's a lot to getting that right and there is HTML but no React component for that right now.  Maybe that's a future enhancement? https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/alerts-and-notifications

cristianonicolai: Yep, thats fine. Would you look into migrating these inline messages to popup in the refactor of the process instance list or should we create a new JIRA for that?

srambach: @cristianonicolai That's probably worth of its own JIRA.

cristianonicolai: ack, created https://issues.redhat.com/browse/KOGITO-817

AjayJagan: will start working on it

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -122,44 +140,70 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
       const processInstanceId = processInstanceData.id;
       const processId = processInstanceData.processId;
       try {
-        const result = await axios.get(
+        await axios.get(
```
cristianonicolai: this should actually be a post

cristianonicolai (changes requested):
@AjayJagan can you add one mock example where abort works? :) I guess it needs to be handled in the mock server.
Also, would be nice to have at least one process in SUSPENDED state to check.

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
AjayJagan: popupmodal
the pop up modal design as requested

packages/management-console/server/MockData/controllers.js
```diff
@@ -58,5 +58,29 @@ module.exports = controller = {
         res.status(500).send(error[0].skip);
         break;
     }
+  },
+  callAbort: (req, res) => {
```
cristianonicolai, line 62: Can you also change the process state in the mocked server? That way once we do another refresh on the page, the process moved into Aborted state.

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -122,44 +149,70 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
-  const handleModalToggle = () => {
+  const handleAbortActiveInstances = useCallback(
+    (processId, processInstanceId) => {
+      axios
+        .post(
+          `${processInstanceData.endpoint}/management/processes/${processId}/instances/${processInstanceId}`
+        )
+        .then(() => {
+          setModalTitle('Abort operation');
+          setModalContent(`Process - ${processId} execution has been aborted`);
```
cristianonicolai: in the UI, I see the message:

> Process - 8035b580-6ae4-4aa8-9ec0-e18e19809e0b execution has been aborted

Seems to somehow be showing _processInstanceId_ instead of _processId_.

AjayJagan: done :+1:

packages/management-console/server/MockData/controllers.js
AjayJagan: done :+1:

cristianonicolai (approved):

Sara4994 (approved):

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
srambach: A couple of tweaks to this based on talking with @kywalker-rh 
- An "info" type would be more appropriate - so an "i" icon
- Change the title to "Process aborted"
- Change the text to "_processname_ - process execution has been aborted"
- Add an "OK" action button

Also, peeking in at the code, if the abort fails, use the title "Process not aborted"

AjayJagan: @srambach - should the info type("i") be used for both success and failure case?

AjayJagan: I have done the above mentioned changes. Please have a look

srambach: Hi, Ajay. Thanks for the changes. I have a few more tweaks -
I think the color of the i icon should be the lighter blue - like the inline alert here. https://www.patternfly.org/v4/documentation/core/components/alert
Also, the size is too big/it needs to be vertically centered.
Both of these things should work if the structure is correct - likely no custom CSS is needed if we get it built correctly. But I'm not sure how it's being built to help out much there.
Also, replace "processname" with the name of the process - so in this case, it would end up as "travels - process execution has been aborted"
Finally, please us "OK" (all uppercase) on the button.
Thanks!

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -188,35 +238,135 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+                  color="var(--pf-global--danger-color--100)"
+                />
+              )}
+            </div>
+            <div className="pf-l-flex__item">
+              <TextContent>
+                <Text component={TextVariants.h1}>{modalTitle}</Text>
+              </TextContent>
+            </div>
+          </div>
```
AjayJagan: @srambach - this is the part which deals with the css and the size of the icon. Can you please comment on this.

srambach: I think that it's better to do the title of the modal this way, and let's put the icon into the title, using a utility class to separate it.
```suggestion
          </div>        header={
          <Title headingLevel={TitleLevel.h1} size={BaseSizes['2xl']}>
            {alertType === 'success' ? (
              <FontAwesomeIcon
                icon={faInfoCircle}
                size="sm"
                color="var(--pf-global--info-color--100)"
                className="pf-u-mr-md"
              />
            ) : (
              <FontAwesomeIcon
                icon={faTimesCircle}
                size="sm"
                color="var(--pf-global--danger-color--100)"
                className="pf-u-mr-md"
              />
            )}
            {modalTitle}
          </Title>
        }

```

(you'll have to update the imports as well)

srambach: ^^ sorry that's a bit rough to match up but I think you can figure out where it goes!

AjayJagan: @srambach - I have updated the code - Can you please review it once :)

srambach: The modal alert looks good. 👍

srambach:
One thing I notice - all the list items are marked as selected (causing the blue line on the left and the box-shadow on everything). Is this intended?

AjayJagan:
> One thing I notice - all the list items are marked as selected (causing the blue line on the left and the box-shadow on everything). Is this intended?

@srambach - I guess these are new updates by patternfly :)

AjayJagan:
> One thing I notice - all the list items are marked as selected (causing the blue line on the left and the box-shadow on everything). Is this intended?

@srambach - i guess thats how patternfly has done it :( . https://www.patternfly.org/v4/documentation/react/components/datalist

srambach:
> > One thing I notice - all the list items are marked as selected (causing the blue line on the left and the box-shadow on everything). Is this intended?
> 
> @srambach - i guess thats how patternfly has done it :( . https://www.patternfly.org/v4/documentation/react/components/datalist

True - that's how the selected state should be shown - but every line of the datalist is marked as "selected" and I don't think they should be, right?

AjayJagan:
> > > One thing I notice - all the list items are marked as selected (causing the blue line on the left and the box-shadow on everything). Is this intended?
> > 
> > 
> > @srambach - i guess thats how patternfly has done it :( . https://www.patternfly.org/v4/documentation/react/components/datalist
> 
> True - that's how the selected state should be shown - but every line of the datalist is marked as "selected" and I don't think they should be, right?

I guess  that is because - I reuse the datalist component again . So , the blue line appears. I tried the same in patternfly website and got the same results.
datalistblueline

srambach:
> > > > One thing I notice - all the list items are marked as selected (causing the blue line on the left and the box-shadow on everything). Is this intended?

This is a Patternfly bug that has been fixed in https://github.com/patternfly/patternfly-react/issues/3446 and will be released shortly.

cristianonicolai:
@AjayJagan I believe this might be fixed already in the upgrade we done for PatternFly on master branch, please try to rebase your PR, that might be enough.

AjayJagan:
> @AjayJagan I believe this might be fixed already in the upgrade we done for PatternFly on master branch, please try to rebase your PR, that might be enough.

@cristianonicolai - Just rebased and pushed it. The blue lines are still present :(

srambach:
At the bottom of this, you can see the versions that should include the fix for the selected state https://github.com/patternfly/patternfly-react/pull/3447

cristianonicolai:
@AjayJagan there is some minor update that is needed, please compare there versions with the link that Sarah provided.

cristianonicolai:
@AjayJagan this needs updates following these minimal versions: https://github.com/patternfly/patternfly-react/pull/3447#issuecomment-571663699

packages/management-console/package.json
```diff
@@ -33,11 +33,17 @@
...
+    "@patternfly/react-charts": "5.2.21",
+    "@patternfly/react-core": "3.134.2",
+    "@patternfly/react-docs": "^3.25.1",
+    "@patternfly/react-icons": "3.14.39",
+    "@patternfly/react-inline-edit-extension": "2.15.6",
+    "@patternfly/react-styles": "3.6.26",
+    "@patternfly/react-table": "2.25.6",
+    "@patternfly/react-tokens": "2.7.25",
+    "@patternfly/react-topology": "2.12.5",
+    "@patternfly/react-virtualized-extension": "1.3.93",
```
AjayJagan: @cristianonicolai - these are the updated ones :)

cristianonicolai: @AjayJagan there is quite a few new ones here, I guess we dont need those?
Also, should we avoid using ^ completely so we match a specific version always?

cristianonicolai (commented):
@AjayJagan please also check that the build failed.

packages/management-console/package.json
AjayJagan: @cristianonicolai - I guess patternfly would keep updating their versions , so shouldn't we keep the "^" with the versions?

packages/management-console/package.json
```diff
@@ -33,10 +33,14 @@
     "@apollo/react-hooks": "^3.1.3",
     "@n1ru4l/react-time-ago": "^1.1.0",
     "@patternfly/patternfly": "^2.49.0",
-    "@patternfly/react-core": "^3.129.3",
-    "@patternfly/react-icons": "^3.14.28",
+    "@patternfly/react-core": "^3.129.4",
+    "@patternfly/react-catalog-view-extension": "1.1.61",
```
cristianonicolai: @AjayJagan is this dependency really needed?

packages/management-console/package.json
```diff
@@ -33,10 +33,14 @@
...
+    "@patternfly/react-inline-edit-extension": "2.14.21",
```
cristianonicolai: same here, seems to be a new one.

packages/management-console/package.json
```diff
@@ -33,10 +33,14 @@
...
+    "@patternfly/react-icons": "^3.14.29",
     "@patternfly/react-styles": "^3.6.15",
-    "@patternfly/react-table": "^2.24.64",
+    "@patternfly/react-table": "^2.24.65",
+    "@patternfly/react-topology": "2.11.49",
```
cristianonicolai: same here

packages/management-console/package.json
```diff
@@ -33,10 +33,14 @@
...
+    "@patternfly/react-virtualized-extension": "1.3.62",
```
cristianonicolai: and this one

cristianonicolai (commented):
@AjayJagan so, we are not upgrading to https://www.patternfly.org/v4/documentation/react/overview/release-notes#202001-release-notes-2020-01-28?

packages/management-console/package.json
cristianonicolai: @AjayJagan I find that to be quite fragile, as we end up with different versions being installed. Coming from a Maven background I would always prefer to have a fixed/stable release to rely on, but that might be different when considering how things are done for JS development.

cristianonicolai:
@srambach do you mind checking this one again, I'm not sure the upgrade fixed the issue you mentioned.

@AjayJagan I notice that when I expand the second process instance, which has no childs, you get an empty space there. Should we open a new JIRA to handle that and give some message, something like "This process has no related sub processes"?

processes

AjayJagan:
> @srambach do you mind checking this one again, I'm not sure the upgrade fixed the issue you mentioned.
> 
> @AjayJagan I notice that when I expand the second process instance, which has no childs, you get an empty space there. Should we open a new JIRA to handle that and give some message, something like "This process has no related sub processes"?
> 
> processes

@cristianonicolai - yeah, I guess we can have a separate JIRA for that.

packages/management-console/package.json
```diff
@@ -33,10 +33,10 @@
     "@apollo/react-hooks": "^3.1.3",
     "@n1ru4l/react-time-ago": "^1.1.0",
     "@patternfly/patternfly": "^2.49.0",
-    "@patternfly/react-core": "^3.129.3",
-    "@patternfly/react-icons": "^3.14.28",
+    "@patternfly/react-core": "^3.129.4",
```
cristianonicolai, line 37: @AjayJagan can you please also align these versions in the task-console.

AjayJagan: done

cristianonicolai:
@srambach does this one looks good to you?

srambach:
Should the process list update to show that the process has been aborted? You can see in this video that it doesn't update, and I can select Abort over and over, even though it's actually been aborted.
Abort-doesn't-update
Also, though it's not shown in the video, the refresh button doesn't update the list (I'm not sure if it's supposed to work yet though)
@AjayJagan @cristianonicolai

AjayJagan:
> Should the process list update to show that the process has been aborted? You can see in this video that it doesn't update, and I can select Abort over and over, even though it's actually been aborted.
> Abort-doesn't-update
> Also, though it's not shown in the video, the refresh button doesn't update the list (I'm not sure if it's supposed to work yet though)
> @AjayJagan @cristianonicolai

oh.. I actually worte a fix. I forgot to update that .Sorry :(

cristianonicolai:
@AjayJagan that was to do with updating the mock right?

@srambach this is something that might actually happen. On a production setup, we will send a request to abort and get a response if it succeeds or not. The problem is that the data index will be on an intermediate state until it receives a message and updates the graphql.
So we could "manually" change the status label to cancelled, because we know it was, but if users refresh the search they might get a different response.

AjayJagan:
> @AjayJagan that was to do with updating the mock right?
> 
> @srambach this is something that might actually happen. On a production setup, we will send a request to abort and get a response if it succeeds or not. The problem is that the data index will be on an intermediate state until it receives a message and updates the graphql.
> So we could "manually" change the status label to cancelled, because we know it was, but if users refresh the search they might get a different response.

@cristianonicolai - I did a query again with the corresponding states selected and that would update the list if we click the 'x' or 'OK' in the modal. Does this sound right to you?

srambach:
@cristianonicolai I'm not sure what the right answer is on updating the status then. Is the modal brought up as soon as the user requests the abort operation, or is it asynchronous and comes up whenever the abort is confirmed?
I'd say we just need to make it as consistent as possible with the real state and what the user thinks they've done :)

srambach:
As for the highlighting/shadow stuff noted above - I don't think that is unique to this PR and so I'll watch it and make a new jira to fix if needed.

cristianonicolai:
@srambach popup shows after we confirm that the process is aborted. This is done via a REST request to the reuntime. So we know that the process is now aborted. The issue is that it could take some time for the data index ( our backend here ) to actually update to this new status.

srambach:
> @srambach popup shows after we confirm that the process is aborted. This is done via a REST request to the reuntime. So we know that the process is now aborted. The issue is that it could take some time for the data index ( our backend here ) to actually update to this new status.

@cristianonicolai OK, I understand. We just need to make sure we're representing the state as truly as possible. Should we add words in the abort modal saying that it might take a moment to update? Or leave as is?

cristianonicolai:
@srambach we put some wording in there but feel free to review/change it, trying to make as clear as possible at least as interim solution.

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -168,35 +210,158 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+        {!checkedArray.includes('ABORTED') &&
+          <Alert variant="info" title="Status change info" className="pf-u-mt-xl" isInline={true}>
+            <TextContent>
+              <Text component={TextVariants.p}>
+                Please note that the process status has been updated in the result list, as result of that, the list will be inconsiistent with the filters until you refresh it.
+           </Text>
+            </TextContent>
+          </Alert>
+        }
+      </Modal>
```
srambach: Let's remove the inline alert and just use text - edited text below. Also, let's put `<strong>` on the modelContent text.

```
        <TextContent>
          <Text>
            <strong>
              {modalContent}
            </strong>
          </Text>
          {!checkedArray.includes('ABORTED') &&
            <Text>
              Note: The process status has been updated. The list may appear inconsistent until you refresh any applied filters.
            </Text>
        }
        </TextContent>
```

srambach:
The Red Hat font isn't getting applied to the modal because the class `pf-m-redhat-font` is applied one level too low. Can we move it to the body tag, and make sure it is applied universally to every page?
now, Red Hat font not applied to modal:
image

correct way:
image

srambach (approved):
Looks great! Thanks for bearing with me on all my questions and "one more things"!

AjayJagan:
> Looks great! Thanks for bearing with me on all my questions and "one more things"!

haha, No worries. It was fun :D

cristianonicolai:
@AjayJagan great work!

PR #50, opened 2019-12-24
KOGITO-452 - Domain Explorer Component

Sara4994:


packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/Queries.ts
```diff
@@ -0,0 +1,19 @@
+import gql from 'graphql-tag';
```
cristianonicolai: can you move this to src/graphql/queries.tsx, that way we can generate a query handler automacially

packages/task-console/src/components/Atoms/TitleComponent/tests/TitleComponent.test.tsx
```diff
@@ -3,8 +3,8 @@ import { shallow } from 'enzyme';
 import TitleComponent from '../TitleComponent';
 
 describe('Title Component Test cases', () => {
```
cristianonicolai, line 5: @Sara4994 please review these changes to task console, I think they can be avoided, same for some files without any content difference.

packages/management-console/src/components/Templates/DomainExplorerPage/Queries.ts
```diff
@@ -0,0 +1,81 @@
+import gql from 'graphql-tag';
+
+export const SAMPLE_URL = gql`
```
cristianonicolai: can you rename this to something like GET_QUERY_TYPES? If that is the goal of this query.

cristianonicolai (commented):
@Sara4994 when you first select the screen there is an extra checkbox there that needs to be removed. See screenshot:
Selection_029

Also, it needs to cleaning of comment out code and removing the filters for this first version.

Also, when selecting travels, there is no need to add a Category: Travels chip.

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/Queries.ts
Sara4994: done

packages/management-console/src/components/Templates/DomainExplorerPage/Queries.ts
Sara4994: done

packages/task-console/src/index.tsx
```diff
@@ -3,4 +3,7 @@ import ApolloClient from 'apollo-boost';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-ReactDOM.render(<div>Task console</div>, document.getElementById('root'));
+ReactDOM.render(
```
cristianonicolai, line 6: @Sara4994 please review these changes to task console, I think they can be avoided, same for some files without any content difference.

Sara4994: @cristianonicolai This change occurs due to husky, which alters the alignment while commiting changes

packages/task-console/src/components/Atoms/TitleComponent/tests/TitleComponent.test.tsx
Sara4994: @cristianonicolai  Same happens here.. Alignment Changes made by husky while comiting code..

cristianonicolai: @Sara4994 ideally we should avoid changes that are not related to the main JIRA, this can make reviewing the PR and git history more difficult. I would suggest doing a separated PR for code format changes only.

AjayJagan (approved):

cristianonicolai:
@Sara4994 do we still need to review/merge this PR or is everything now into https://github.com/kiegroup/kogito-apps/pull/57 ?

PR #55, opened 2020-01-20
Upgrade TypeScript target to ES6

cristianonicolai:


cristianonicolai:
@Sara4994 @AjayJagan  this changes drops the size of the bundle from ~7mb to ~4mb

PR #57, opened 2020-01-31
KOGITO-452 Domain Explorer Version-1

Sara4994:


packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+  };
+
+  return (
+    <React.Fragment>
+      {displayTable && (
+        <Table
+          cells={finalKeys}
+          rows={values}
+          onSelect={onRowSelect}
+          aria-label="Filterable Table Demo"
```
cristianonicolai:
```suggestion
          aria-label="Domain Explorer Table"
```

packages/management-console/src/components/Templates/DashboardComponent/Dashboard.tsx
```diff
@@ -1,24 +1,69 @@
...
+  SkipToContent,
+  PageSidebar,
+  PageHeader,
+  Nav,
+  NavList,
+  NavItem
+} from '@patternfly/react-core';
+import React, { useState } from 'react';
+import { Redirect, Route, Link } from 'react-router-dom';
+// import HeaderComponent from '../../Organisms/PageHeaderComponent/HeaderComponent';
```
cristianonicolai: please remove comment out code.

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+        >
+          <TableHeader />
```
cristianonicolai, line 90: @Sara4994 can we remote the check box column from this version?

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -0,0 +1,239 @@
...
+  };
+
+  let b;
+  let finalResult: any = [];
+  let a: any;
+
+  const childSelectionItems = (_data, title, ...attr) => {
+    let nestedTitles = '';
+    b = _data.map(group => {
+      const label = title + '>' + attr.join();
```
cristianonicolai:
```suggestion
      const label = title + ' > ' + attr.join();
```
just adding a extra space here, before and after the string.

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+  const getPicker = useGetPickerQuery({
+    variables: { columnPickerType }
+  });
+
+  async function generateQuery() {
+    try {
+      await axios
+        .post(
+          'http://localhost:4000/graphql',
```
cristianonicolai: should we create a new query type for this? That would probably avoid having the data index URL in here as well.

cristianonicolai (commented):
@Sara4994 apart from a few comments, I also notice that _processInstances_ and _userTasks_ attributes inside _metadata_ are not expanded. These are complex types that are know before hand, will always be there (ProcessInstanceMeta and UserTaskInstanceMeta).

packages/management-console/src/components/Templates/DashboardComponent/Dashboard.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: removed

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

cristianonicolai:
@Sara4994 we also need to include the page header and breadcrumb for the Domain Explorer page, similar to the one in the Process Instances.

header

Sara4994:
@cristianonicolai All the above comments are taken care and PR is updated with latest code.

cristianonicolai:
@Sara4994 overall looks good but I'm getting some weird results when switching domain and columns, see screenshot:

domain

It seems that some state is kept when you change domains, if the user does that, we should clear the table completely.

Sara4994:
@cristianonicolai Can you please let me know the selections that you have made. Because i am not getting such result for same selection(lastUpdated)
Screenshot from 2020-02-13 16-06-19

Sara4994:
Screenshot from 2020-02-13 16-06-19

Sara4994:
> @Sara4994 overall looks good but I'm getting some weird results when switching domain and columns, see screenshot:
> 
> domain
> 
> It seems that some state is kept when you change domains, if the user does that, we should clear the table completely.

@cristianonicolai fixed this bug as well.

srambach:
The width of some items are being constrained by the grid and then things start overlapping and wrapping out of their containers. We really don't need to be using the grid layout for this at all, either for the toolbar or for the content section of the page. The DataToolbar should do all the correct spacing.
We talked about moving the column selector to the left, so if we do that you also don't need to worry about how to get it to right align. And can we move the button to the right of the selector?

And as you can see in the Filterable Table demo, just putting the DataToolbar and then the Table, with no grid, is sufficient. https://www.patternfly.org/v4/documentation/react/demos/filtertabledemo

srambach:
Is the empty state part of this PR? (probably better as another since this one has been a lot?)

Sara4994:
@srambach I have just removed grid stuffs and added DataToolBar components, also moved column picker to left as mentioned. Added Empty state as well.. Please have a look if it looks fine to you...

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -0,0 +1,264 @@
...
+      getQueryTypes.data.__schema &&
+      getQueryTypes.data.__schema.queryType.find(item => {
+        if (item.name === option.type.name) {
+          return item;
+        }
+      })
+    );
+  };
+
+  let b;
```
cristianonicolai: perhaps give a more intutive name for variables a and b :)

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <TableBody />
+        </Table>
+      )}
+      {!displayTable && (
+        <Bullseye>
+          <EmptyState>
+            <EmptyStateIcon icon={SearchIcon} />
+            <Title headingLevel="h5" size="lg">
+              Nothing to display
```
cristianonicolai:
```suggestion
              No domain data to display
```

packages/management-console/src/graphql/queries.tsx
```diff
@@ -78,3 +78,109 @@ const GET_PROCESS_INSTANCE = gql`
     }
   }
 `;
+
+const GET_PICKER = gql`
+  query getPicker($columnPickerType: String!) {
```
cristianonicolai: Perhaps give a more meaningful name for the query like getFieldsFromType?

cristianonicolai:
```suggestion
  query getPicker($type: String!) {
```

packages/management-console/src/graphql/queries.tsx
```diff
@@ -78,3 +78,109 @@ const GET_PROCESS_INSTANCE = gql`
...
+            kind
+          }
+        }
+      }
+    }
+  }
+`;
+
+const GET_QUERY = gql`
+  query getQuery {
```
cristianonicolai: getQueryFields?

packages/management-console/src/graphql/queries.tsx
```diff
@@ -78,3 +78,109 @@ const GET_PROCESS_INSTANCE = gql`
...
+            name
+          }
+        }
+      }
+    }
+  }
+`;
+
+const GET_SCHEMA = gql`
+  query getSchema($currentQuery: String!) {
```
cristianonicolai: getInputFieldsFromQuery?

packages/management-console/src/graphql/queries.tsx
```diff
@@ -78,3 +78,109 @@ const GET_PROCESS_INSTANCE = gql`
...
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+const GET_TYPES = gql`
+  query getTypes($currentArgument: String!) {
```
cristianonicolai: getInputFieldsFromType?

cristianonicolai:
```suggestion
  query getTypes($type: String!) {
```

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const getPicker = useGetPickerQuery({
+    variables: { columnPickerType }
+  });
+
+  async function generateQuery() {
+    if (columnPickerType) {
+      try {
+        await axios
+          .post(
```
cristianonicolai: review if apollo client can be used instead.

cristianonicolai (commented):
@Sara4994 it's looking really nice. Another enhancement we need, filter out Job type from the domain list. We've recently added it to data index, and it is another technical type.

Seems that boolean values aren't shown in the result table.
When I select in the column picker, an attribute that doesn't return any value, the column is not added to the result table.

cristianonicolai:
@Sara4994 when we have a domain with multiple attributes in the root level, there is quite a big space between then, see:

image

packages/management-console/src/graphql/queries.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

Sara4994:
> @Sara4994 when we have a domain with multiple attributes in the root level, there is quite a big space between then, see:
> 
> image



> @Sara4994 when we have a domain with multiple attributes in the root level, there is quite a big space between then, see:
> 
> image

@cristianonicolai ofcourse yes... because we have just given empty string for SelectGroup.. We dont have other choice right now in PF. Even @srambach suggested me same solution, to provide empty string.

packages/management-console/src/graphql/queries.tsx
Sara4994: done

packages/management-console/src/graphql/queries.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

cristianonicolai:
@Sara4994 sure, but can't have a single SelectGroup with empty string? It seems that we're creating one for enry single property.

packages/management-console/src/graphql/queries.tsx
Sara4994: done

packages/management-console/src/graphql/queries.tsx
Sara4994: done

packages/management-console/src/graphql/queries.tsx
Sara4994: done

srambach:
> @Sara4994 sure, but can't have a single SelectGroup with empty string? It seems that we're creating one for enry single property.

Exactly what I was going to suggest. That should provide good spacing.

srambach:
I see a few things:
- When switching to a different domain and back again, would it be possible to remember what you had selected previously for that domain?
- When too many columns are selected, the table overflows and there is no horizontal scroll. I'll look at where we want to put the overflow. We might want to wrap the column in one more div with `overflow-x: auto` for that.
- After adding many columns and the removing some, I see columns with no headers, and they are not getting enough space.
image
- Here I had removed bookingNumber but it and another field still appeared (without a column header)
image
- I selected a few fields, applied it, and then unchecked all of them. "Apply Columns" doesn't remove them to go back to the empty state.
- I'm seeing the removed column data left over even when the column is removed sometimes. Here I added and then removed columns. Also, there used to be 3 rows and suddenly there are 4. It's hard to tell what's going on since most of the fields are identical.
image
Sorry, I know this is a lot for one comment. Let me know if you prefer that I put each thing in a different comment going forward.

srambach:
One more - On Travels, there are some shared names between `traveller` and `visaApplication` and I'm seeing the checkbox check both at the same time.

srambach:
Another thought - @kywalker-rh and I were talking about using `/` instead of `>` to separate the attributes. So `flight / arrival`? @cristianonicolai what do you think? Also, with or without spaces? I'm thinking with spaces to make it easier to read.

cristianonicolai:
@srambach agree to use `/` with space.

@srambach regarding your point for remembering the selection across domains, I think we can postpone that, one because we will introduce a landing page in between any domain change and second that we also should look into how to save searches/filters and columns, similar to v7 approach. Apart from that, it all seems reasonable issues that we need to look into.

Sara4994:
> @Sara4994 when we have a domain with multiple attributes in the root level, there is quite a big space between then, see:
> 
> image

This is fixed

Sara4994:
> Another thought - @kywalker-rh and I were talking about using `/` instead of `>` to separate the attributes. So `flight / arrival`? @cristianonicolai what do you think? Also, with or without spaces? I'm thinking with spaces to make it easier to read.

done

Sara4994:
> One more - On Travels, there are some shared names between `traveller` and `visaApplication` and I'm seeing the checkbox check both at the same time.

Fixed

srambach:
@Sara4994 When you uncheck a column, the list changes immediately to the empty state until the button is pressed. Checking an additional column doesn't do this. Can we have the list stay in place until the user clicks "Apply columns" again?

Uncheck-goes-to-empty

srambach:
@Sara4994 At the narrower screen width where the filters collapse, clicking on the filter goes to a blank screen.

SelectDomain-emptyscreen

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <EmptyState>
+            <EmptyStateIcon icon={SearchIcon} />
+            <Title headingLevel="h5" size="lg">
+              No domain data to display
+            </Title>
+            <EmptyStateBody>
+              Select a Domain from dropdown to see content
+            </EmptyStateBody>
+          </EmptyState>
+        </Bullseye>
```
srambach: I suggest adding a Card component around the Bullseye, which will make the white background to match the table. Also, I suggest some wording changes:
```
        <Card component={'div'}>
          <CardBody>
            <Bullseye>
              <EmptyState>
                <EmptyStateIcon icon={SearchIcon} />
                <Title headingLevel="h5" size="lg">
                  No domain data to display
                </Title>
                <EmptyStateBody>
                  To see content, select a domain from the dropdown, and then select the columns to view.
                </EmptyStateBody>
              </EmptyState>
            </Bullseye>
          </CardBody>
        </Card>
```

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,112 @@
+import React, { useState, useEffect } from 'react';
+import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import {
+  Title,
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateBody,
+  EmptyStateSecondaryActions,
```
srambach: I think this is unused, so it can be removed

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+  Bullseye,
+  Button
```
srambach: This also looks like it's not used here, so it can be removed

packages/management-console/src/components/Templates/DomainExplorerPage/DomainExplorerPage.css
```diff
@@ -0,0 +1,5 @@
+.table-OverFlow {
+    overflow-y: auto;
```
srambach: Let's start using a modified BEM naming system similar to Patternfly's. (let's discuss specifics in chat)
So for now, let's name this --kogito-runtime-tooling--table-container
And, I think we need overflow-x here, not y, so that we ensure a horizontal scroll when needed.

packages/management-console/src/components/Templates/DomainExplorerPage/DomainExplorerPage.css
```diff
@@ -0,0 +1,5 @@
...
+}
+
+table.pf-c-table td {min-width: 10em;}
```
srambach: This will apply to every PF table on the page, which right now is just one, but this isn't really good practice to rely on. So a better way would be to add another class to that table, say `.kogito-runtime-tooling--domain-explorer__table` and then we can apply this style just to that class.
Also, even though it's just one attribute, I'd put the `min-width:10em;` on its own line.
```
.kogito-runtime-tooling--domain-explorer__table td {
  min-width: 10em;
}
```

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  return (
+    <React.Fragment>
+      {displayTable && (
+        <Table
+          cells={finalKeys}
+          rows={values}
+          aria-label="Domain Explorer Table"
+        >
```
srambach, line 89: (See comment on the CSS file)
Add `className="kogito-runtime-tooling--domain-explorer__table"` to define a new class for styling the cells inside.

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

packages/management-console/src/components/Templates/DomainExplorerPage/DomainExplorerPage.css
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: Removed Button and EmptyStateSecondaryActions... Other things are used

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: Removed EmptyStateSecondaryActions, other things are used

Sara4994:
> I see a few things:
> 
> * When switching to a different domain and back again, would it be possible to remember what you had selected previously for that domain?
> * When too many columns are selected, the table overflows and there is no horizontal scroll. I'll look at where we want to put the overflow. We might want to wrap the column in one more div with `overflow-x: auto` for that. -----> Fixed
> * After adding many columns and the removing some, I see columns with no headers, and they are not getting enough space.  ----> This is fixed
>   image
> * Here I had removed bookingNumber but it and another field still appeared (without a column header) ------> This 
>   image
> * I selected a few fields, applied it, and then unchecked all of them. "Apply Columns" doesn't remove them to go back to the empty state.
> * I'm seeing the removed column data left over even when the column is removed sometimes. Here I added and then removed columns. Also, there used to be 3 rows and suddenly there are 4. It's hard to tell what's going on since most of the fields are identical.
>   image
>   Sorry, I know this is a lot for one comment. Let me know if you prefer that I put each thing in a different comment going forward.



> @Sara4994 When you uncheck a column, the list changes immediately to the empty state until the button is pressed. Checking an additional column doesn't do this. Can we have the list stay in place until the user clicks "Apply columns" again?

@srambach Actually i see a bug in the table... guess you shared me a screenshot in an older comment, when you deselect the columns in the dropdown and hit apply filter the table data collapses...
this happens only for element "id". When you select and deselect Id, the older data still stays on the table.. this doesnot happen for anyother element in root level... to test this i added some new element at root level, when i select and deselect the new element i dont see such collapse of table...
To overcome this bug, i had to refresh the table content while deselecting the columns... That the reason it appears like empty state comes  up when you deselect columns

Sara4994:
> @Sara4994 When you uncheck a column, the list changes immediately to the empty state until the button is pressed. Checking an additional column doesn't do this. Can we have the list stay in place until the user clicks "Apply columns" again?

@srambach Actually i see a bug in the table... guess you shared me a screenshot in an older comment, when you deselect the columns in the dropdown and hit apply filter the table data collapses...
this happens only for element "id". When you select and deselect Id, the older data still stays on the table.. this doesnot happen for anyother element in root level... to test this i added some new element at root level, when i select and deselect the new element i dont see such collapse of table...
To overcome this bug, i had to refresh the table content while deselecting the columns... That the reason it appears like empty state comes  up when you deselect columns

srambach:
Thanks for the CSS changes and bug fixes. I see one more problem - now that the filter can open at narrow widths, I see that the column list is cleared. 
1. Select several fields *including id*
2. Make screen narrow. Open Filter.
Column selection is reset to nothing.
3. Make screen wide. Column picker still has nothing selected. Open it and select a column. Click Apply Columns.
Previous bug is there - columns are out of sync. 
image

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const firstKey = Object.keys(columnFilters)[0];
+  const tableContent = columnFilters[firstKey];
+
+  const keys = [];
+  const values = [];
+  if (tableContent) {
+    const finalResult = tableContent.map(item => getKeys(item));
+    finalResult.map(result => {
+      keys.push(result.tempKeys);
+      values.push({ cells: result.tempValue, id: Math.random().toString() });
```
srambach: Change to 
`      values.push({ cells: result.tempValue, rowKey: Math.random().toString() });
`

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+    <React.Fragment>
+      {displayTable && (
+        <Table
+          cells={finalKeys}
+          rows={values}
+          aria-label="Domain Explorer Table"
+          className="kogito-management-console--domain-explorer__table"
+        >
+          <TableHeader />
+          <TableBody rowKey={'iqu911'} />
```
srambach: Change to
`          <TableBody rowKey="rowKey" />
`

Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

srambach:
OK - those last two lines of code to change are because there was some problem where `id` was conflicting with something else.
I'm assuming the `Math.random()` is just for sample data :-) and once there's real data, it will be a real unique value from the data.
Huge thanks to @mturley for his help on this.

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -0,0 +1,277 @@
...
+            isExpanded={isExpanded}
+            placeholderText="Pick Columns"
+            ariaLabelledBy="Column Picker dropdown"
+            isGrouped
+            maxHeight="60vh"
+          >
+            {getAllChilds(finalResult, 'props')}
+          </Select>
+          <Button variant="primary" onClick={generateQuery}>
+            Apply Columns
```
srambach: Sentence case - use `Apply columns` :-D

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+      });
+    });
+    const rowObject: any = {};
+    if (tableLoading) {
+      rowObject.cells = [
+        {
+          props: { colSpan: 8 },
+          title: (
+            <Bullseye>
+              <SpinnerComponent spinnerText="Loading Domain Explorer..." />
```
srambach: Sentence case `Loading domain explorer`

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+        <Card component={'div'}>
+          <CardBody>
+            <Bullseye>
+              <EmptyState>
+                <EmptyStateIcon icon={SearchIcon} />
+                <Title headingLevel="h5" size="lg">
+                  No domain data to display
+                </Title>
+                <EmptyStateBody>
+                  Select a Domain from dropdown to see content
```
srambach: Sentence case, and "the" 😸   `Select a domain from the dropdown to see content`

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -0,0 +1,277 @@
...
+      {!getPicker.loading && columnPickerType && (
+        <>
+          <Select
+            variant={SelectVariant.checkbox}
+            aria-label="Select Input"
+            onToggle={onToggle}
+            onSelect={onSelect}
+            selections={selected}
+            isExpanded={isExpanded}
+            placeholderText="Pick Columns"
```
srambach: Sentence case "Pick columns"

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

srambach:
bitmoji Just a couple of comments cleaning up the text. Looks good!

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
Sara4994: done

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
Sara4994: done

Sara4994:
> bitmoji Just a couple of comments cleaning up the text. Looks good!

@srambach done

srambach (approved):
👍

cristianonicolai (approved):
@Sara4994 great work!

AjayJagan (approved):
looks good !

PR #58, opened 2020-02-05
KOGITO-593 Add pagination controls to datalist

AjayJagan:
- added pagination to data list.

### DIFF
--- a/packages/management-console/server/MockData/graphql.js
+++ b/packages/management-console/server/MockData/graphql.js
@@ -782,5 +782,1357 @@ module.exports = data = [
       }
     ],
     childProcessInstances: []
+  },
+  {
+    id: '8035b580-6ae4-4aa8-9ec0-e18e19809e0b1',
+    processId: 'travels',
+    parentProcessInstanceId: null,
+    processName: 'travels',
+    roles: [],
+    state: 'ACTIVE',
+    rootProcessInstanceId: null,
+    endpoint: 'http://localhost:4000/',
+    addons: [],
+    error: {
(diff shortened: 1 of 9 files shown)


cristianonicolai:
@AjayJagan this PR needs rebase now.

AjayJagan:
> @AjayJagan this PR needs rebase now.

done :)

cristianonicolai:
@AjayJagan @srambach does the pagination component supports not having a total count? We had similar issue for v7 and we want to avoid having to do extra query just for the sake of having a count number. If this is something that isn't really supported by the component we might need to check an alternative way of doing the pagination.

srambach:
@cristianonicolai One of our designers suggested something along the lines of "1 to 20 of 300ish" Would something like that help here? Would you be querying often enough to be close, if not exact? Also I'm not sure how well that internationalizes ;-P

cristianonicolai:
@srambach I don't think there is much point on showing some data that will be inaccurate anyway ( how would that help users?), not to mention the internationalization issue :P

srambach:
I think the Pagination controls should show up in the toolbar, whether we use the Patternfly component or our own.
Like you see here: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/pagination#main-content

It looks really good... @cristianonicolai how about making it say "1 - 10 of **about** 13" 🥺😁

cristianonicolai:
@srambach it looks nice indeed. 
Well, how do we come up with the **13** ?? :P The point is to remove any need for doing another query, simply to do a count :/

PR #59, opened 2020-02-07
KOGITO-1105 Management console branding

AjayJagan:
- changed the management console logo
- added test case for DashboardComponent and BrandComponent


packages/management-console/src/components/Atoms/BrandComponent/BrandComponent.tsx
```diff
@@ -2,15 +2,23 @@ import React from 'react';
...
     history.push('/ProcessInstances');
   };
-  return <Brand src={kogitoLogoRGB} alt="Kogito Logo" onClick={onLogoClick} />;
+  return (
+    <Brand
+      src={managementConsoleLogo}
+      alt="Management Console Logo"
+      onClick={onLogoClick}
+      className="logo_dimensions"
+    />
```
AjayJagan: @srambach - When I try to add the new logo svg in the Brand component, by default then with of the svg becomes zero. If I add a css class to add some width then the svg image appears.
The same doesnot happen for the about page.
Is it because the svg size is too big that the brand component is hiding it. Can you please guide me in this

srambach: Yes! I just found the problem/solution elsewhere yesterday. The solution is not in the CSS, but in the SVG file. I see that inside the SVG file, the viewbox is set to "0 0 339.93 36.37". Right after that, add `width="399.93px" height="36.37px"` That should fix the problem.

AjayJagan: Thanks. That resolved the issue :)

srambach:
Great! This is almost there. I noticed there was extra space to the left, and I think we need to tweak the size parameters inside the SVG.  I'm going to email you a new SVG, and then we need one line of CSS added at the Page level that will adjust the height for within the banner.
`--pf-c-page__header-brand-link--c-brand--MaxHeight: 42px;`

AjayJagan:
> Great! This is almost there. I noticed there was extra space to the left, and I think we need to tweak the size parameters inside the SVG. I'm going to email you a new SVG, and then we need one line of CSS added at the Page level that will adjust the height for within the banner.
> `--pf-c-page__header-brand-link--c-brand--MaxHeight: 42px;`

@srambach I have added the css at in the Dashboard component- which is the kind of the base component. Can you please review if I have added it properly

packages/management-console/src/components/Templates/DashboardComponent/Dashboard.css
```diff
@@ -0,0 +1,3 @@
+.dashboard {
```
srambach: Let's start using a modified BEM naming for CSS classes. I'd suggest `.kogito-management-console--dashboard-page` for this.

packages/management-console/src/components/Templates/DashboardComponent/Dashboard.tsx
```diff
@@ -18,7 +18,7 @@ const Dashboard: React.FC = () => {
         header={<HeaderComponent />}
         skipToContent={PageSkipToContent}
         mainContainerId={pageId}
-        className="page"
+        className="dashboard"
```
srambach: If we are using a modified BEM naming for CSS classes. I'd suggest `kogito-management-console--dashboard-page` for this.

srambach:
Aside from the requested CSS class name change, this looks good!

packages/management-console/src/components/Templates/DashboardComponent/Dashboard.tsx
cristianonicolai: +1

cristianonicolai (approved):

srambach:
Looks good! Thanks! 👍

Sara4994:
@AjayJagan Build failure due to test case

Sara4994 (approved):

PR #61, opened 2020-02-13
KOGITO-1178 Add empty state to child process if no child is available

AjayJagan:
- added empty state if no child is available
- addded test case for EmptyStateComponent

### DIFF
--- a/packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
+++ b/packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
@@ -5,11 +5,14 @@ import {
   EmptyStateIcon,
   EmptyStateBody,
   Bullseye,
-  Button
+  Button,
+  EmptyStateVariant
 } from '@patternfly/react-core';
-import { SearchIcon } from '@patternfly/react-icons';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
+import {
+  SearchIcon,
+  ExclamationTriangleIcon,
+  InfoCircleIcon
+} from '@patternfly/react-icons';
 import '@patternfly/patternfly/patternfly-addons.css';
 
 interface IOwnProps {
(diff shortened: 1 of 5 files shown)


AjayJagan:
@srambach @cristianonicolai  - can you please review the design and suggest ideas on the design :)

cristianonicolai (approved):
looks good

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
```diff
@@ -9,7 +9,10 @@ import {
 } from '@patternfly/react-core';
 import { SearchIcon } from '@patternfly/react-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
+import {
+  faExclamationTriangle,
+  faInfoCircle
+} from '@fortawesome/free-solid-svg-icons';
```
srambach: We should be importing icons from Patternfly -  @patternfly/react-icons probably?

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
```diff
@@ -55,6 +58,14 @@ const EmptyStateComponent: React.FC<IOwnProps> = ({
...
           />
         )}
+        {iconType === 'infoCircleIcon' && (
+          <FontAwesomeIcon
+            icon={faInfoCircle}
+            size="2x"
+            color="var(--pf-global--info-color--100)"
+            className="pf-u-mb-xl"
+          />
+        )}
```
srambach, line 61: We should be able to use the empty state to put the icon in and size and position it correctly.
<EmptyStateIcon icon={InfoCircleIcon} />

srambach:
I started to comment in the code but realized I need a more general comment. I'm a little confused as to why we need to manually adjust so many things to do with the icon in the EmptyStateComponent. And it looks like there are some deprecated props that we are using. (there might be newly deprecated, I'm not sure) https://www.patternfly.org/v4/documentation/react/components/emptystate

Also, I assume we need to do this because we can't know for sure until loading that there are no child processes? Otherwise we just wouldn't put the expansion on, right?

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
AjayJagan: > I started to comment in the code but realized I need a more general comment. I'm a little confused as to why we need to manually adjust so many things to do with the icon in the EmptyStateComponent. And it looks like there are some deprecated props that we are using. (there might be newly deprecated, I'm not sure) https://www.patternfly.org/v4/documentation/react/components/emptystate
> 
> Also, I assume we need to do this because we can't know for sure until loading that there are no child processes? Otherwise we just wouldn't put the expansion on, right?

Yes Sarah, so I thought we should show the carets uniformly , and only when we expand it ,we would know if the child is present or not.

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
```diff
@@ -37,31 +40,38 @@ const EmptyStateComponent: React.FC<IOwnProps> = ({
...
     <Bullseye>
-      <EmptyState>
-        {iconType === 'searchIcon' && <EmptyStateIcon icon={SearchIcon} />}
-        {iconType === 'warningTriangleIcon' && (
-          <FontAwesomeIcon
-            icon={faExclamationTriangle}
-            size="3x"
+      <EmptyState variant={EmptyStateVariant.full}>
+        {iconType === 'searchIcon' && (
+          <EmptyStateIcon icon={SearchIcon} size="sm" />
```
AjayJagan, line 45: @srambach - if the size and color prop of empty state icon are deprecated, then how do we define the color and size of the icons

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
srambach: > Yes Sarah, so I thought we should show the carets uniformly , and only when we expand it ,we would know if the child is present or not.

I talked with @kywalker-rh about this - we agree that normally if there is nothing in the expansion,  we just would not show the expansion. So I just want to ask the question of whether we need this empty state, or if we just don't have the expansion caret. I'm imagining maybe since this list is highly dynamic, when the list is rendered there are no children, but maybe then they would appear, so we want to be able to expand anytime? It's a little bit of a stretch so I wanted to ask. @cristianonicolai any thoughts?

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
srambach: I've sparked a little conversation over on patternfly-react about what should be configurable on an empty state. Let's leave the props alone and maybe revisit empty states another day.

AjayJagan: okay :)

Sara4994 (approved):

packages/management-console/src/components/Atoms/EmptyStateComponent/EmptyStateComponent.tsx
AjayJagan: @srambach - The children are not loaded initially. So at this point we wouldn't know if the parent has the children or not. Only when we click on the caret and expand it, we would run a query for fetching the children and at this point if the children are empty then we show the no process available. @cristianonicolai - am I correct?

cristianonicolai: @AjayJagan correct.

srambach (approved):
Looks good!

PR #63, opened 2020-02-14
Fix environment variables

cristianonicolai:
- Missing single comma in data index url
- Use defaults if not defined


nmirasch (approved):
works fine!

PR #72, opened 2020-02-24
KOGITO-1229 list redesign

AjayJagan:
- added icons to states
 - added external link icon
 - added link to process details page in process name
 - added popover for error instances

### DIFF
--- a/packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
+++ b/packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
@@ -1,5 +1,5 @@
 import Moment from 'react-moment';
-import React, { useCallback, useState, useEffect } from 'react';
+import React, { useCallback, useState, useEffect, useRef } from 'react';
 import axios from 'axios';
 import {
   Button,
@@ -17,23 +17,32 @@ import {
   Bullseye,
   KebabToggle,
   Modal,
-  Alert,
   TextContent,
-  TextVariants,
   Text,
-  Flex,
-  FlexItem,
   Title,
   TitleLevel,
-  BaseSizes
+  BaseSizes,
+  DataList
 } from '@patternfly/react-core';
 import { Link } from 'react-router-dom';
 import SpinnerComponent from '../../Atoms/SpinnerComponent/SpinnerComponent';
-import { useGetChildInstancesLazyQuery } from '../../../graphql/types';
+import {
+  useGetChildInstancesLazyQuery,
+  ProcessInstanceState
+} from '../../../graphql/types';
 import EmptyStateComponent from '../../Atoms/EmptyStateComponent/EmptyStateComponent';
-
+import {
+  OnRunningIcon,
+  CheckCircleIcon,
+  BanIcon,
+  PausedIcon,
+  ErrorCircleOIcon,
+  ExternalLinkAltIcon,
+  HistoryIcon
+} from '@patternfly/react-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faInfoCircle, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
+import ErrorPopover from '../../Atoms/ErrorPopoverComponent/ErrorPopoverComponent';
 /* tslint:disable:no-string-literal */
 
 export interface IProcessInstanceError {
@@ -45,6 +54,7 @@ interface IProcessInstance {
   id: string;
   processId: string;
   parentProcessInstanceId: string | null;
+  rootProcessInstanceId: string | null;
   processName: string;
   start: string;
   state: string;
@@ -55,7 +65,7 @@ interface IProcessInstance {
 export interface IOwnProps {
   id: number;
   processInstanceData: IProcessInstance;
-  checkedArray: string[]
+  checkedArray: string[];
 }
 
 const DataListItemComponent: React.FC<IOwnProps> = ({
@@ -67,12 +77,11 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
   const [isLoaded, setisLoaded] = useState(false);
   const [isChecked, setisChecked] = useState(false);
   const [childList, setChildList] = useState({});
-  const [openModal, setOpenModal] = useState(false);
   const [isModalOpen, setIsModalOpen] = useState(false);
   const [modalTitle, setModalTitle] = useState('');
   const [modalContent, setModalContent] = useState('');
   const [alertType, setAlertType] = useState(null);
-
+  const [isPopoverOpen, setIsPopoverOpen] = useState<boolean>(false);
   const [getChildInstances, { loading, data }] = useGetChildInstancesLazyQuery({
     fetchPolicy: 'network-only'
   });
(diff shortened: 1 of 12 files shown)


cristianonicolai (approved):

Sara4994 (approved):

packages/management-console/server/MockData/types.js
```diff
@@ -88,12 +88,12 @@ module.exports = typeDefs = gql`
   }
 
   enum ProcessInstanceState {
-    PENDING
```
cristianonicolai: The actual enum is upper case on Data Index, changing here will probably break the query, see https://github.com/kiegroup/kogito-apps/blob/master/data-index/data-index-service/src/main/resources/schema.graphqls#L73

srambach: I have CSS Magic... we can still display it as "Pending" even if it comes in as "PENDING" by using `text-transform: capitalize;` on that element when we put it into the list. 🧙‍♀️

packages/management-console/src/components/Molecules/DataToolbarComponent/Datatoolbar.css
```diff
@@ -0,0 +1,14 @@
...
+  text-transform: lowercase;
+}
+
+.pf-c-check__label::first-letter {
+  text-transform: uppercase;
+}
+
+.pf-c-check__label {
+  text-transform: lowercase;
+}
```
AjayJagan, line 14: Hey @srambach - I tried adding text-transform:capitalize , but it was still showing every letter in caps. So I did this work around. I am still trying to figure it out, if I am not successful ,please help me out :)

srambach:
@AjayJagan it's coming into a better shape :) @cristianonicolai 
- All status icons should be default (black) color except for completed (green) and error (red) as those are.
- Spacing of the Apply Filter button is off (I'll comment on the line to fix)
- The name linked to details beside the external icon linked to the endpoint application needs to be addressed. My comments are in the PR I made to your branch. Hopefully there is a name of the endpoint that we could use as text? If not, another option would be to move that to the kebab. But we need to know a little more about how it's used to say for sure - but the details link next to just the external link icon going to the endpoint cannot stay. :-)  https://github.com/AjayJagan/kogito-apps/pull/2

srambach:
OK, I guess it won't let me comment on the line - in DataToolbarComponent.tsx, you can remove the `variant="filter-group"` on line 134:
```
 <DataToolbarGroup variant="filter-group">
```
That's intended to remove the space between multiple filter dropdowns, but the button should be spaced normally.

AjayJagan:
> @AjayJagan it's coming into a better shape :) @cristianonicolai
> 
> * All status icons should be default (black) color except for completed (green) and error (red) as those are.
> * Spacing of the Apply Filter button is off (I'll comment on the line to fix)
> * The name linked to details beside the external icon linked to the endpoint application needs to be addressed. My comments are in the PR I made to your branch. Hopefully there is a name of the endpoint that we could use as text? If not, another option would be to move that to the kebab. But we need to know a little more about how it's used to say for sure - but the details link next to just the external link icon going to the endpoint cannot stay. :-)  [AjayJagan#2](https://github.com/AjayJagan/kogito-apps/pull/2)
I have removed the variant = filter groups and changed the color's of the status.
The only thing that's pending is the endpoint. @cristianonicolai  can you please tell us what should be done regarding that
image

packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
```diff
@@ -389,136 +390,204 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
...
+                </Link>
+                <a
+                  href={`${processInstanceData.endpoint}/management/processes/${processInstanceData.processId}/instances/${processInstanceData.id}`}
+                  target="_blank"
+                >
+                  <Button variant="link" className="pf-u-pt-0">
+                    <ExternalLinkAltIcon />
+                  </Button>
+                </a>
+              </DataListCell>,
```
srambach: How about using this to solve the problem of the external icon by itself? And then we also talked about only putting it on the parent, can you do that?
```
              <DataListCell key={1}>
                <Link to={'/ProcessInstances/' + processInstanceData.id} className="pf-u-mr-sm">
                    <strong>{processInstanceData.processName} </strong>
                </Link>
                <Link to={`${processInstanceData.endpoint}/management/processes/${processInstanceData.processId}/instances/${processInstanceData.id}`} target="_blank"  aria-label={`Endpoint for ${processInstanceData.processName}`}>
                  Endpoint <ExternalLinkAltIcon />
                </Link>
              </DataListCell>,
```
@AjayJagan @cristianonicolai

AjayJagan: yep, displaying it only on the parent is possible. awaiting @cristianonicolai response.

srambach (approved):
I like the endpoint link. Can you make it go to a new tab? @AjayJagan  @cristianonicolai 
Otherwise, approved! Thanks.

AjayJagan:
> I like the endpoint link. Can you make it go to a new tab? @AjayJagan @cristianonicolai
> Otherwise, approved! Thanks.

Done :+1:

cristianonicolai:
@AjayJagan noticed a few things:

- Can you add a mock process with status complete so we can check the UI?
- @srambach  Should the popover close when we click elsewhere? Otherwise we endup with an overlay like in the screenshot below.
- Should buttons in the popover be aligned to the right side? Do we need a close button there?
Selection_032
- In the process details page, error state is not showing up
Selection_033

packages/management-console/server/server.js
```diff
@@ -50,7 +50,7 @@ app.post(
   controller.callRetrigger
 );
 app.post(
-  '/management/processes/:processId/instances/:processInstanceId',
```
cristianonicolai: the endpoint was correct, method should be DELETE.

packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
```diff
@@ -12,14 +12,68 @@ import {
...
 import { Link } from 'react-router-dom';
 
 interface IOwnProps {
   loading: boolean;
   data: any;
 }
 const ProcessDetails: React.FC<IOwnProps> = ({ data, loading }) => {
+  const stateIconCreator = state => {
+    switch (state) {
+      case 'ACTIVE':
```
cristianonicolai: would it help to use the TypseScript enum here?

cristianonicolai (changes requested):
see previous comments

packages/management-console/server/server.js
AjayJagan: This change is made in KOGITO 789

cristianonicolai: perhaps just revert the change here then.

AjayJagan: done :)

cristianonicolai: sorry but still showing _/abort_, I guess the idea was to just revert the change on this line.

cristianonicolai (commented):
@AjayJagan we also need to show the error popup on the process details page, when status is in error.

AjayJagan:
> @AjayJagan we also need to show the error popup on the process details page, when status is in error.

@cristianonicolai  we were adding a view error along with the process management in process details page, So I thought we had to add the pop up instead of the view error. (because the pop contains Skip and Retry which are related to process management)

cristianonicolai:
@AjayJagan we can review the details page in another commit, just revert the change in the abort endpoint and we should be good ti go with this PR.

cristianonicolai (approved):

PR #84, opened 2020-02-28
KOGITO-1313: Override netty version to avoid BufferOverflow in infinispan client

evacchi:
depends on https://github.com/kiegroup/kogito-runtimes/pull/346


evacchi:
looks like the fix is more involved than just bumping versions around, so I'll just do it on the quarkus extension for now

PR #89, opened 2020-03-04
Configure cypress as in Monorepo

jstastny-cz:
Please see how Cypress fits into the repo structure.
Raise any questions you may have before merging.


jstastny-cz:
@AjayJagan @Sara4994 any thoughts?

AjayJagan:
> @AjayJagan @Sara4994 any thoughts?

seems fine to me

Sara4994 (approved):

Sara4994:
> @AjayJagan @Sara4994 any thoughts?

Looks good to me

cristianonicolai (commented):
looks good, just waiting for a green build.

jstastny-cz:
I seem to be struggling with webpack config (I guess), if I run 'yarn run test' then the tests pass, but when I build using 'yarn run webpack --config webpack.prod.js' then the build fails (and that's what happens during our build).

My wild guess is, that the tsconfig.js 'extends' and the 'types' compilerOptions are not passed to webpack during build.

@AjayJagan @Sara4994 @cristianonicolai I need your help here. Can you:
a) checkout locally and try yarn test on management-console (I mean with my changes), to check that it really works
b) try running the webpack command
c) if it fails - help me fix it :-D

cristianonicolai:
@jstastny-cz good news is that I can reproduce the same error locally. I guess we might need some help from either @AjayJagan or @Sara4994 about how this can be tweaked.

Sara4994:
@jstastny-cz @cristianonicolai  add the below line of code in tsconfig.json file present inside cypress folder. this would resolve the error  TS2339: Property 'toMatchSnapshot' does not exist on type 'Assertion

"exclude": ["src/**/*.ts",
        "src/**/*.tsx",
        "src/**/*.jsx",
        "src/**/*.js"]

The problem here is both cypress and jest has configurations which collapses on one another. by telling typescript compiler to limit itself to "src" folder you resolve the conflict between global expect from Jest and cypress solves the issue.

in case cypress requires react components(.tsx files inside src folder) to write test cases, you can only ignore  test.tsx files(test files of jest) alone inside src like this in cypress/tsconfig.json
"exclude": ["src/**/*.test.tsx"]

jstastny-cz:
I've made changes that move me closer to the solution .... instead of excluding in cypress module, I am separating what I import in the packages ... basically only things under src.
When doing this, for some reason I needed to create the custom.d.ts and include svg and png data types there.
Locally I've had snapshot failure, but I suspect that might be due my local env, trying to run the build here.

Sara4994:
@jstastny-cz @cristianonicolai 
Below changes helped me to resolve the assertion error(both while running yarn run build:prod and yarn run test)

We need to removed include block from global tsconfig.ts, reason being tsconfig.json inside cypress folder extends the global tsconfig which include block due to which types of jest and cypress colloids. So here we need to restrict our typescript under src folder, still cypress also will handle typescript since we already have include block in cypress/tsconfig.json :

**1) To get rid of error "TS2339: Property 'toMatchSnapshot' does not exist on type 'Assertion'.":** remove include from global tsconfig.ts and move to tsconfig.ts files present inside management-console and task-console like below

 "include": [
    "src/**/*.ts",
    "src/**/*.tsx",
    "src/**/*.jsx",
    "src/**/*.js"
  ]

**2) To get rid of error"TS2307: Cannot find module '../../../static/KogitoAbout.png'." :** 
 move declare.d.ts file <root>/config/declare.d.ts to both management-console/src and task-console/src folders.

**3) To clear snapshot failure:**
 run script yarn run update-snapshot that would clear the snapshot failure.

jstastny-cz:
@Sara4994 that's basically what I have put in place (except for the snapshot update), except of the structure. I'll try to keep the declare.d.ts where it is and use explicit include for the file only.

I'll update the snapshot and ask for a new round of review.

jstastny-cz:
@Sara4994 Thanks to your help the issues are resolved and build is green. Please review and let me know if any changes are needed.
@cristianonicolai @AjayJagan I'd like to ask you the same - please review.

cristianonicolai (approved):

Sara4994:
> @Sara4994 that's basically what I have put in place (except for the snapshot update), except of the structure. I'll try to keep the declare.d.ts where it is and use explicit include for the file only.
> 
> I'll update the snapshot and ask for a new round of review.

All good :)

Sara4994 (approved):

tomasdavidorg (approved):

PR #99, opened 2020-03-12
KOGITO-1293: 404 Error page

Sara4994:


cristianonicolai (changes requested):
@Sara4994 I think a 404 should be shown when users hit a url that we can't handle. Ex: http://localhost:9000/MyProcess
In the case tha we don't find the process, I think we should show a more specific message. Saying process instance with id xxxxxxx not found.

Sara4994:
> @Sara4994 I think a 404 should be shown when users hit a url that we can't handle. Ex: http://localhost:9000/MyProcess
> In the case tha we don't find the process, I think we should show a more specific message. Saying process instance with id xxxxxxx not found.

Done

cristianonicolai:
@Sara4994 looks better but I think we still can not say _404 Error: page not found_ for when a process is not found. Would suggest using _Process not found_ and also the smaller text that is already there: _Process instance with the id ss not found_.

@Sara4994 @srambach @kywalker-rh this is a case where an id is passed as paramenter in the url and the backend returns that such process does not exist. I'm wondering if that would fall under the no data usage of empty state: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state#no-data-to-show.

kywalker-rh:
@cristianonicolai I haven't been focused on Runtimes, so I don't understand the use case completely, but it does sound like the error you are mentioning should be different than a 404, and provide more data. The empty state is definitely a good guideline. If you can give a reason why the error occurred (possibly with details as an expandable section) and how to resolve it that is best. I agree, a 404 error sounds like it would be misleading in this case.

cristianonicolai:
@kywalker-rh thanks, yeah in this situation is really unclear how the user could get into this problem, but it doesn't feel like an error.

@Sara4994 can you update to use this empty state: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state#no-data-to-show. For when the process is not found?

packages/management-console/src/components/Molecules/ErrorComponent/ErrorComponent.tsx
```diff
@@ -1,48 +1,52 @@
...
+            <EmptyStateIcon
+              icon={ExclamationCircleIcon}
+              size="md"
+              color="var(--pf-global--danger-color--100)" />
+            <Title headingLevel="h1" size="4xl">404 Error: page not found</Title>
+            <EmptyStateBody>
+            {props.location.state ? props.location.state.description : 'This page could not be found.'}
+                  </EmptyStateBody>
+            <Button variant="primary" onClick={redirectHandler}>
+              Take me home
```
cristianonicolai:
```suggestion
              Navigate to homepage 
```

PR #100, opened 2020-03-12
KOGITO-1323: Expand side bar navigation menu by default

Sara4994:


cristianonicolai (approved):

AjayJagan (approved):
:+1:

srambach:
The initial load looks good. 
When I manually open or close the menu and then resize the window, it seems like it should remain open or closed, but sometimes it changes and sometimes not. (I can't recognize a pattern.) I don't think this is a huge problem but feels a little off. To me, if it's an easy fix then it's worth it but if not I can live with it. @cristianonicolai @Sara4994 @kywalker-rh

cristianonicolai:
is probably worth a quick look for sure, otherwise we can create a JIRA for that.

cristianonicolai:
@srambach would you mind creating JIRA to track this error separately?

PR #96, opened 2020-03-09
KOGITO-1269 add businesskey to process instance list

AjayJagan:
- added search bar for businesskey wise search

PR #103, opened 2020-03-16
[Work in progress] KOGITO-1446 Maintain search results in process instance list when navigating to process Details page

AjayJagan:


cristianonicolai:
@AjayJagan you probably need https://github.com/kiegroup/kogito-apps/pull/96 to be merged so the business key is also kept.

AjayJagan:
> @AjayJagan you probably need #96 to be merged so the business key is also kept.

@cristianonicolai - will complete it once #96  is merged :)

PR #109, opened 2020-03-18
KOGITO-1231 add select all capability to process instance list

AjayJagan:
- added select all button in datatoolbar

### DIFF
--- a/packages/management-console/src/components/Molecules/DataToolbarComponent/DataToolbarComponent.tsx
+++ b/packages/management-console/src/components/Molecules/DataToolbarComponent/DataToolbarComponent.tsx
@@ -12,7 +12,13 @@ import {
   SelectVariant,
   InputGroup,
   TextInput,
-  ButtonVariant
+  ButtonVariant,
+  Dropdown,
+  DropdownToggle,
+  DropdownToggleCheckbox,
+  DropdownItem,
+  DropdownPosition,
+  Badge
 } from '@patternfly/react-core';
 import { FilterIcon, SyncIcon, SearchIcon } from '@patternfly/react-icons';
 import _ from 'lodash';
@@ -37,6 +43,10 @@ interface IOwnProps {
   pageSize: number;
   setSearchWord: (searchWord: string) => void;
   searchWord: string;
+  isAllChecked: boolean;
+  setIsAllChecked: (isAllChecked: boolean) => void;
+  setSelectedNumber: (selectedNumber: number) => void;
+  selectedNumber: number;
 }
 const DataToolbarComponent: React.FC<IOwnProps> = ({
   checkedArray,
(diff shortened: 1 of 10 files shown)


srambach:
Is there any chance of having it work as shown in this demo https://www.patternfly.org/v4/documentation/react/demos/bulkselecttable#component-title
and in the design documentation?
https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/bulk-selection#main-content
I.e. Drop down allows the user to select none, this page, or all. I don't know if "all" is reasonable given the dynamic list. Also, it should be updating to show a total selected as individual ones are checked off and when the overall checkbox is checked.
@cristianonicolai @kywalker-rh @AjayJagan

cristianonicolai:
@srambach perhaps we could use select none and select all. Select all could have a variation: select all parent and select all parent and child?
Select page, doesn't seem to apply here.

cristianonicolai:
@AjayJagan this needs rebase now

srambach:
> @srambach perhaps we could use select none and select all. Select all could have a variation: select all parent and select all parent and child?

@cristianonicolai if that makes sense to you, then it sounds good to me.

@AjayJagan FYI at the moment, I see a bug where if you use the checkbox to select all, then change the filter, the checkbox remains checked even though only some or nothing is now checked in the list.

cristianonicolai:
@srambach +1 from me

srambach:
Can we change the dropdown to:
```
Select none
Select all parent processes
Select all processes
```
And the bulk selector would either say nothing (if nothing is selected), "All parent processes", or "All processes"
The text in the selector doesn't update if I change the checkboxes on the list.
If you want to finish this for tomorrow's release, the above would be enough for me. However, do we also want to consider the things below for the next release?
Do we want to use the `-` state of the checkbox to indicate partial selection - like when you've clicked only some, and maybe when only parents are selected?
I'm looking at design guidelines in https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/bulk-selection#main-content and https://www.patternfly.org/v4/documentation/react/demos/bulkselecttable#component-title to match how it works.
WDYT? @cristianonicolai

cristianonicolai:
@srambach regarding the text for the dropdown +1
As for the text once selection, I'm not sure, wouldn't be enough to be consistent and only show the number of items selected? That seems to be what is used in PatternFly anyway.
As for the improvements, that makes sense, can you open a JIRA for that please :)

cristianonicolai (approved):

cristianonicolai:
@AjayJagan needs rebase

packages/management-console/src/components/Molecules/DataToolbarComponent/DataToolbarComponent.tsx
```diff
@@ -192,12 +367,38 @@ const DataToolbarComponent: React.FC<IOwnProps> = ({
...
+                    key="split-checkbox"
+                    aria-label="Select all"
+                    isChecked={isAllChecked}
+                    onChange={() =>
+                      handleCheckboxSelectClick('parent&child', true)
+                    }
+                  />
+                ]}
+              >
+                <Badge isRead>{selectedNumber}</Badge>
```
srambach: Instead of the badge here, can we just use text please? If it is zero, then no text.
```
                {selectedNumber} selected
```

AjayJagan: yeah can do that

srambach:
Great that we can display the count! I see a problem thought in the number of selected processes. If I select all, then it doesn't count the collapsed children - but if I expand one then it's selected. In this example, there were 3, but when I expanded one, one of the children was selected so now there are at least 4 selected but the number says 3. 
SelectAll3-4

cristianonicolai:
@srambach the child processes are loaded asynchronous, that means they are only loaded if you expand. To me, it works as expected, by selecting only what has been loaded in the UI.
I think one addition we should have, is a expand/collapse all option. That would cover your point.

Sara4994 (approved):

AjayJagan:
> @srambach the child processes are loaded asynchronous, that means they are only loaded if you expand. To me, it works as expected, by selecting only what has been loaded in the UI.
> I think one addition we should have, is a expand/collapse all option. That would cover your point.

@cristianonicolai - there was a bug which was selecting the child process if the parent in also selected on the first load. Now with this push I have made them independent. I guess the number works properly now. :) .have a look and tell me if it needs any change

srambach:
👍 Selection numbers match now, thanks!

srambach (approved):

PR #113, opened 2020-03-19
KOGITO-1405 - Use manual acknowledgment for processing messages

cristianonicolai:


sutaakar (approved):

tiagodolphine (approved):
LGTM, @radtriste let's test the jar from the PR to see if it solves the issue.

radtriste:
ok so ... first test with own built image:

1. I came into an issue that the data-index was trying to connect to localhost:9092 instead of the Kafka installed. 
Checking the environment, all `MP_INCOMING_*` env were set correctly to installed Kafka, so I would expect it tries to connect to correct Kafka. _After investigation, I needed to add `KAFKA_BOOTSTRAP_SERVERS` to correct url so it could work. 
My guess is that it comes from 'application.properties' file [configuration](https://github.com/kiegroup/kogito-apps/blob/3245f49aeb37f99def55c6c90db99de5cdf1b888/data-index/data-index-service/src/main/resources/application.properties#L19) where both are defined and Quarkus is taking the kafa.bootstrap-servers as priority... `QUARKUS_KAFKA_BOOTSTRAP_SERVERS` was set by the operator but seems it is not taken into account anymore ..._ 
**This should be somehow corrected and checked with Quarkus team. I can open an issue for that if needed.**
But this is not what I was testing today.
2. I still have the problem and will investigate further on Monday ...

sutaakar:
@radtriste The issue is already opened by @ricardozanini - https://github.com/quarkusio/quarkus/issues/7791

radtriste:
Investigation done, see my comments on [JIRA](https://issues.redhat.com/browse/KOGITO-1405?focusedCommentId=14007550&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14007550)

and see my [proposition](https://issues.redhat.com/browse/KOGITO-1405?focusedCommentId=14007575&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14007575) for next actions
what do you think ?

Also IMO, this PR can be close ...

cristianonicolai:
@sutaakar @tiagodolphine @radtriste closing this PR, please take a look on https://github.com/kiegroup/kogito-apps/pull/153

PR #121, opened 2020-03-25
KOGITO-1465: Resource not found error when navigating directly to /Pr…

nmirasch:
…ocessInstances
react component managing served endpoints

### DIFF
--- a/management-console/src/main/java/org/kie/kogito/mgmt/VertxRouter.java
+++ b/management-console/src/main/java/org/kie/kogito/mgmt/VertxRouter.java
@@ -2,17 +2,18 @@ package org.kie.kogito.mgmt;
 
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
-import io.quarkus.vertx.web.Route;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpHeaders;
+import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
+import io.vertx.ext.web.handler.StaticHandler;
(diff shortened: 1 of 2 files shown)


cristianonicolai:
@nmirasch I'm getting the following error: Uncaught SyntaxError: Unexpected token '<' then nothing loads. Am I missing something? I downloaded the artifact from Jenkins and also build locally.

nmirasch:
@cristianonicolai I changed this a bit adding the exact path to be rerouted. Now it works better, it was a matter of cleaning caches. Could you take a look to this please? Thanks!

cristianonicolai:
@nmirasch I tried to clean the cache, maybe I missed something?

This current approach works, but I guess will be hard to keep covering all the cases, for instance:

http://localhost:8280/DomainExplorer/Travels
http://localhost:8280/ProcessInstances/867c4400-8f72-487d-a734-7ecc21aced34

Should also work, that's why I'm wondering.

nmirasch:
@cristianonicolai Adding this '*' to the path pattern, cover those mentioned cases, the only thing i to take this into account when new friendly url is added. Do you think this could work?

cristianonicolai (approved):

management-console/src/main/java/org/kie/kogito/mgmt/VertxRouter.java
```diff
@@ -37,7 +38,13 @@ public void init() {
                 .replace("__DATA_INDEX_ENDPOINT__", "\"" + dataIndexHttpURL + "/graphql\"");
     }
 
-    @Route(path = "/", methods = GET)
+    void setupRouter(@Observes Router router) {
+        router.route("/").handler(ctx -> ctx.response().putHeader("location", "/ProcessInstances/").setStatusCode(302).end());
```
cristianonicolai, line 42: @nmirasch this redirect should be handle in React itself, just wondering if it is necessary.

PR #129, opened 2020-03-28
KOGITO-1466: Invalid version and Data index url on about page

nmirasch:


nmirasch:
@cristianonicolai @AjayJagan @Sara4994  could you take a look? Just comment that I removed the dataindex url link, regarding the only thing that I think that make sense is to put there a link to the graphiql. Wdyt about let this as a text?

cristianonicolai (approved):

cristianonicolai:
@nmirasch it makes sense as it is, graphiql is not likely to be enabled on production build.

PR #135, opened 2020-04-01
KOGITO-1131 Add management operations to process instance details page

AjayJagan:
- added abort process support to process details page

### DIFF
--- a/packages/management-console/src/components/Templates/ProcessDetailsPage/ProcessDetailsPage.tsx
+++ b/packages/management-console/src/components/Templates/ProcessDetailsPage/ProcessDetailsPage.tsx
@@ -7,27 +7,106 @@ import {
   PageSection,
   Title,
   Card,
-  Bullseye
+  Bullseye,
+  Button,
+  Split,
+  SplitItem,
+  OverflowMenu,
+  OverflowMenuContent,
+  OverflowMenuGroup
 } from '@patternfly/react-core';
-import React from 'react';
+import React, { useState } from 'react';
 import { Link, Redirect } from 'react-router-dom';
 import ProcessDetails from '../../Organisms/ProcessDetails/ProcessDetails';
 import ProcessDetailsProcessDiagram from '../../Organisms/ProcessDetailsProcessDiagram/ProcessDetailsProcessDiagram';
 import ProcessDetailsProcessVariables from '../../Organisms/ProcessDetailsProcessVariables/ProcessDetailsProcessVariables';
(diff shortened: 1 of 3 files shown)


cristianonicolai (approved):

packages/management-console/src/components/Templates/ProcessDetailsPage/ProcessDetailsPage.tsx
```diff
@@ -67,18 +165,32 @@ const ProcessDetailsPage = ({ match }) => {
...
+        {!loading ? (
           <Grid gutter="md" span={12} lg={6} xl={4}>
-          <GridItem span={12}>
-              <Title headingLevel="h1" size="4xl">
-                <ProcessDescriptor
-                  processInstanceData={data.ProcessInstances[0]}
-                />
-              </Title>
+            <GridItem span={12}>
+              <Split gutter={'md'} component={'div'}>
```
srambach: You can add `className="pf-u-align-items-center"` to align the button a little better to the title vertically.

packages/management-console/src/components/Templates/ProcessDetailsPage/ProcessDetailsPage.tsx
```diff
@@ -67,18 +165,32 @@ const ProcessDetailsPage = ({ match }) => {
...
+                <SplitItem isFilled={true}>
+                  <Title headingLevel="h1" size="4xl">
+                    <ProcessDescriptor
+                      processInstanceData={data.ProcessInstances[0]}
+                    />
+                  </Title>
+                </SplitItem>
+                <SplitItem>
+                  <OverflowMenu breakpoint="lg">
```
srambach, line 189: Now that there's just one button, I don't think we actually need the OverflowMenu and it could just be the Button. But it's not harmful if there will be other buttons there some day.

srambach (approved):
Left a couple of minor comments, but otherwise looks great!

packages/management-console/src/components/Templates/ProcessDetailsPage/ProcessDetailsPage.tsx
AjayJagan: @srambach - I asked @cristianonicolai  and he suggested that we might add some buttons in the future . So I guess we can keep the `<OverflowMenu>` :P

nmirasch (approved):

Sara4994 (approved):

PR #152, opened 2020-04-06
KOGITO-1412: Added refresh button on Domain explorer

Sara4994:


packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -247,6 +253,13 @@ const DomainExplorerColumnPicker: React.FC<IOwnProps> = ({
 
     return unique;
   }
+
+  const onRefresh = () => {
+    if(enableRefresh && parameters.length > 1) {
+    // tslint:disable-next-line: no-floating-promises
```
cristianonicolai: can we avoid disabling the tslint rule?

cristianonicolai (approved):

AjayJagan (approved):
looks good ! :+1:

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
Sara4994: @cristianonicolai done

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -363,6 +370,9 @@ const DomainExplorerColumnPicker: React.FC<IOwnProps> = ({
           <Button variant="primary" onClick={generateQuery}>
             Apply columns
           </Button>
+          <Button variant="plain" onClick={onRefresh} className="pf-u-m-md">
```
srambach: We need an aria-label for screenreaders. You can use this:
```
          <Button variant="plain" onClick={onRefresh} className="pf-u-m-md" aria-label={"Refresh list"}>
```

Sara4994: @srambach  done

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.css
```diff
@@ -6,4 +6,8 @@
     --pf-c-table--nested--first-last-child--PaddingLeft: 0;
     margin-bottom: var(--pf-global--spacer--lg);
     max-width: 80vw;
+}
+
+.kogito-management-console--domain-explorer__table .pf-c-table__toggle{
```
srambach, line 11: I'm not sure why this is here? We should always use the spacer variables, but I think the spacing was fine here without any adjustment.

cristianonicolai: @srambach I think the problem is that we don't add any content when value is null, we can review that if we need another placeholder to be displayed which would avoid this fix.

PR #173, opened 2020-04-13
KOGITO-1780: Remembers selection while navigating to details page

Sara4994:


cristianonicolai:
@Sara4994 please rebase

Sara4994:
> @Sara4994 please rebase

@cristianonicolai done :+1:

cristianonicolai:
@Sara4994 it has conflicts now :/

Sara4994:
> @Sara4994 it has conflicts now :/

resolved that too. :+1:

AjayJagan (approved):
:+1:

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -97,7 +99,11 @@ const DomainExplorerColumnPicker: React.FC<IOwnProps> = ({
   };
 
   useEffect(() => {
-    parameters.length !== 1 && generateQuery();
+    if (rememberedParams.length === 0 && parameters.length !== 1) {
```
nmirasch: @Sara4994  This condition could be replaced by (rememberedParams.length === 0 && parameters.length !== 1)  || (rememberedParams.length > 0) to avoid the "else if" executing the same code ?

nmirasch: @Sara4994 it looks good and works fine, just left this minor comment..

cristianonicolai: +1

nmirasch (approved):
looks good

srambach:
I'm not sure of the behavior when I click a process that doesn't exist from the domain explorer and end up at the "Process not found page." The back button returns to the state I was in (expected), however, the "Go to the domain explorer" button returns me to the default state. (I expected to return to it in the state I was in)

Sara4994:
> I'm not sure of the behavior when I click a process that doesn't exist from the domain explorer and end up at the "Process not found page." The back button returns to the state I was in (expected), however, the "Go to the domain explorer" button returns me to the default state. (I expected to return to it in the state I was in)

@srambach this is fixed.. now button click redirects to DE table

srambach (approved):
To error page and back with the button or back button looks good now. 👍

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -193,7 +204,7 @@ const DomainExplorerColumnPicker: React.FC<IOwnProps> = ({
             query: gql`
               ${Query.query}
             `,
-            fetchPolicy: 'no-cache'
+            fetchPolicy: enableCache ? undefined : 'network-only'
```
cristianonicolai:
```suggestion
            fetchPolicy: enableCache ? 'cache-first' : 'network-only'
```
'cache-first' is the default anyway, just to make the code clear.

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -26,6 +26,9 @@ export interface IOwnProps {
   getPicker: any;
   setError: any;
   setDisplayEmptyState: any;
+  rememberedParams: any;
+  enableCache: any;
```
cristianonicolai:
```suggestion
  enableCache: boolean;
```

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -26,6 +26,9 @@ export interface IOwnProps {
...
+  setEnableCache: any;
```
cristianonicolai, line 31: do we also need to pass the state function here? Wouldnt that make on component dependend on the other?

cristianonicolai (commented):
@Sara4994 I'm seeing two queries executed when navigating back to the domain list, both in the _DomainExplorerDashboard.tsx_. 
- getInputFieldsFromQuery
- getColumnPickerAttributes

both returning no data. wondering if these make sense.

For https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L75, should this one be moved to _DomainExplorerColumnPicker.tsx_ instead? Then same logic applies for cache, no cache.

For https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L160. Not sure if I'm missing something but the result is not used? Should we keep this query if that is the case?

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
Sara4994: @cristianonicolai  i would say yes, right now it is dependent. i am passing this function here because i have set enableCache as false when "apply columns" button is clicked, so that the query fetched data from network.

cristianonicolai: sure, but does the parent component needs to be notified/changed because of that? I was wondering if this would be the internal state of the child component.

Sara4994: @cristianonicolai yep, the function is used to set enableCache as true here in parent component, https://github.com/kiegroup/kogito-apps/blob/1acc9223f61a079f3ab077dc64a45e95ae007c4e/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L136

it would be set as true when moving process details -> Domain list so that the query will run with cache, but then when we click on "apply columns" it should trigger query freshly without cache so i am setting at as false in the child component using same function "setEnableCache"

Sara4994:
> @Sara4994 I'm seeing two queries executed when navigating back to the domain list, both in the _DomainExplorerDashboard.tsx_.
> 
> * getInputFieldsFromQuery
> * getColumnPickerAttributes
> 
> both returning no data. wondering if these make sense.
> 
> For https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L75, should this one be moved to _DomainExplorerColumnPicker.tsx_ instead? Then same logic applies for cache, no cache.
> 
> For https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L160. Not sure if I'm missing something but the result is not used? Should we keep this query if that is the case?



> @Sara4994 I'm seeing two queries executed when navigating back to the domain list, both in the _DomainExplorerDashboard.tsx_.
> 
> * getInputFieldsFromQuery
> * getColumnPickerAttributes
> 
> both returning no data. wondering if these make sense.
> 
> For https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L75, should this one be moved to _DomainExplorerColumnPicker.tsx_ instead? Then same logic applies for cache, no cache.
> 
> For https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L160. Not sure if I'm missing something but the result is not used? Should we keep this query if that is the case?

getInputFieldsFromQuery : this one is removed, as this was added for filter options on DE. but yeah this is removed now.

getColumnPickerAttributes: this query is used to populate options into column picker dropdown with first 5 columns are selected by default on page load. the results of this query used on both DEDashboard.tsx and DEColumnPicker.tsx thats the reason i kept it on top level.
you can find some code written on DEDashboard.tsx based on the results of this query,

https://github.com/Sara4994/kogito-apps/blob/Kogito-1780-new/packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx#L98

cristianonicolai (approved):

PR #174, opened 2020-04-13
KOGITO-1378 Periodic job already running is not canceled

tiagodolphine:
A periodic job already running can be canceled properly.

### DIFF
--- a/jobs-service/src/test/java/org/kie/kogito/jobs/service/resource/JobResourceTest.java
+++ b/jobs-service/src/test/java/org/kie/kogito/jobs/service/resource/JobResourceTest.java
@@ -18,6 +18,7 @@ package org.kie.kogito.jobs.service.resource;
 
 import java.io.IOException;
 import java.time.temporal.ChronoUnit;
+import java.util.UUID;
 
 import javax.inject.Inject;
 
@@ -106,17 +107,7 @@ public class JobResourceTest {
         final String id = "3";
         final Job job = getJob(id);
         create(jobToJson(job));
-        final ScheduledJob scheduledJob = given()
-                .pathParam("id", id)
-                .when()
-                .get(JobResource.JOBS_PATH + "/{id}")
-                .then()
-                .statusCode(200)
-                .contentType(ContentType.JSON)
-                .assertThat()
-                .extract()
-                .as(ScheduledJob.class);
-        assertEquals(scheduledJob.getId(), job.getId());
+        assertGetScheduledJob(id);
     }
 
     @Test
@@ -124,16 +115,7 @@ public class JobResourceTest {
         final String id = "4";
         final Job job = getJob(id);
         create(jobToJson(job));
-        final ScheduledJob scheduledJob = given()
-                .pathParam("id", id)
-                .when()
-                .get(JobResource.JOBS_PATH + "/{id}")
-                .then()
-                .statusCode(200)
-                .contentType(ContentType.JSON)
-                .assertThat()
-                .extract()
-                .as(ScheduledJob.class);
+        final ScheduledJob scheduledJob = assertGetScheduledJob(id);
         assertEquals(scheduledJob.getId(), job.getId());
         assertEquals(0, scheduledJob.getRetries());
         assertEquals(JobStatus.SCHEDULED, scheduledJob.getStatus());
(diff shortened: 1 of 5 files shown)


cristianonicolai (approved):

radtriste:
Do we have a reproducer to test this ?

sutaakar:
+1
Which of the test changes actually simulate the issue described in JIRA (invoking the job cancel on periodic job)?
IMHO:
- https://github.com/kiegroup/kogito-apps/pull/174/files#diff-6bfd4f14485feb2bf4797ed88129743bR168 simulates submitting of new job when a job with same id already exists
- https://github.com/kiegroup/kogito-apps/pull/174/files#diff-6bfd4f14485feb2bf4797ed88129743bR293 covers cancelling of non periodic scheduled job. Should this be considered as a reproducer for the JIRA?
- https://github.com/kiegroup/kogito-apps/pull/174/files#diff-6bfd4f14485feb2bf4797ed88129743bR303 covers cancellation of not scheduled job.

jobs-service/src/test/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobSchedulerTest.java
```diff
@@ -165,6 +165,11 @@ private void testExistingJob(boolean expired, JobStatus jobStatus) {
         verify(tested(), expired ? never() : times(1)).doSchedule(delayCaptor.capture(), eq(job));
         verify(jobRepository, expired ? never() : times(1)).save(scheduleCaptor.capture());
 
+        //assert always a scheduled job is canceled (periodic or not)
+        Optional.ofNullable(jobStatus)
+                .filter(SCHEDULED::equals)
+                .ifPresent(s -> verify(tested()).cancel(scheduleCaptorFuture.capture()));
```
MarianMacik: Value of this captor is never retrieved.

tiagodolphine: added the missing assertions.

tiagodolphine:
@radtriste @sutaakar I added a new integration test `JobResourceTest#cancelRunningPeriodicJobTest` that basically reproduces the error without the changes on this PR and passes with the modification on this PR.
The scenario is not easy to test without the integration running on vertx because it is asynchronous. Basically, I create a new periodic job, wait until it is running and after that cancel the running job asserting it was removed on the API perspective and on Vertx, the problem was it was not being removed from Vertx scheduling.
Thanks.

sutaakar:
If I understand it correctly the fix for issue described in JIRA is addressed in https://github.com/kiegroup/kogito-apps/pull/174/files#diff-96206ab3bf634f8422aa73b8f060b41dL334-R334
The issue was related to the fact that the reactive stream created in https://github.com/kiegroup/kogito-apps/pull/174/files#diff-96206ab3bf634f8422aa73b8f060b41dL338 wasn't executed.

The issue also affects non periodic jobs which are already scheduled for execution.

@tiagodolphine Can you please confirm/correct me?

tiagodolphine:
Hi @sutaakar, exactly this is what was causing the issue (on my analysis), and you are right, it was affecting periodic jobs and non-periodic (at least looking at the code), but I'll add a new integration test considering a non-periodic job, in this case, we can make sure and confirm it, this running the test against the code without the modifications on this PR.

tiagodolphine:
@sutaakar added one more test to cancel non-periodic, and it has the issue, the same as periodic.

jobs-service/src/test/java/org/kie/kogito/jobs/service/resource/JobResourceTest.java
```diff
@@ -100,25 +107,115 @@ void getAfterCreate() throws Exception {
         final String id = "3";
         final Job job = getJob(id);
         create(jobToJson(job));
-        final ScheduledJob scheduledJob = given()
+        assertGetScheduledJob(id);
+    }
+
+    @Test
+    void executeTest() throws Exception {
```
sutaakar, line 114: This test seems to be same as getAfterCreate(), it just contains additional assertions, can they be merged together?

sutaakar (approved):

sutaakar:
Thanks

tiagodolphine:
@MarianMacik all good for you or it is missing anything for the approval?
@cristianonicolai I think this would be an important fix to be on the 0.9.1, wdyt?
Thanks.

tiagodolphine:
@cristianonicolai It would be great to include this in **0.10.0**.

cristianonicolai:
@tiagodolphine I created the followup JIRA https://issues.redhat.com/browse/KOGITO-1941 as tests are randomly failing in Jenkins, will disable test for now.

cristianonicolai:
PR https://github.com/kiegroup/kogito-apps/pull/191

PR #177, opened 2020-04-14
KOGITO-1840 enable endpoint for child process

AjayJagan:
- enabled endpoint for child process instances

PR #185, opened 2020-04-16
KOGITO-1842 adding endpoint link to process details page

AjayJagan:
- added link support for endpoint in process details page

### DIFF
--- a/packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
+++ b/packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
@@ -14,68 +14,32 @@ import {
 import React from 'react';
 import {
   LevelDownAltIcon,
-  LevelUpAltIcon,
-  OnRunningIcon,
-  CheckCircleIcon,
-  BanIcon,
-  PausedIcon,
-  ErrorCircleOIcon
+  LevelUpAltIcon
 } from '@patternfly/react-icons';
 import { Link } from 'react-router-dom';
-import { ProcessInstanceState } from '../../../graphql/types';
 import ProcessDescriptor from '../../Molecules/ProcessDescriptor/ProcessDescriptor';
+import { stateIconCreator } from '../../../utils/Utils';
+import { ProcessInstance } from '../../../graphql/types';
+import EndpointLink from '../../Molecules/EndpointLink/EndpointLink';
 
 interface IOwnProps {
-  data: any;
+  data: {
+    ProcessInstances?: Pick<
+      ProcessInstance,
+      | 'id'
+      | 'processName'
+      | 'businessKey'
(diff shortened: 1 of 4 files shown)


cristianonicolai:
@AjayJagan this needs rebase

packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
```diff
@@ -108,9 +72,18 @@ const ProcessDetails: React.FC<IOwnProps> = ({ data, from }) => {
             </Text>
           </FormGroup>
           <FormGroup label="Endpoint" fieldId="endpoint">
-            {data.ProcessInstances[0].endpoint ? (
+            {data.ProcessInstances[0].serviceUrl ? (
```
cristianonicolai: @AjayJagan should we reuse `EndpointLink` from https://github.com/kiegroup/kogito-apps/pull/177 in here?

srambach: I agree, I think this should be consistent between the two and use the same component. If it doesn't make sense to display the URL on the listing, then don't do it here either. If it does, then put it on the listing as well.

packages/management-console/src/components/Organisms/ProcessDetails/ProcessDetails.tsx
```diff
@@ -108,9 +72,18 @@ const ProcessDetails: React.FC<IOwnProps> = ({ data, from }) => {
...
               <Text component={TextVariants.p}>
-                {data.ProcessInstances[0].endpoint}
+                <Button
```
srambach: Button will give this `white-space:nowrap`. So I guess we need to set `white-space:normal` here.
And to make a long URL, and anything else (like a long id) wrap, then we also need to set `word-break:break-all`. I think this is best done maybe on the `form` tag on line 51.
Ideally, we should have a top level CSS (or SCSS) file to place some utility-type classes. If we don't want to do that in this PR, let's make an issue and add the styles inline here for the moment. You can do it inline like this: 
```
<Form style={{
                // @ts-ignore
                'word-break': 'break-all'}}>
```

nmirasch (approved):

srambach (approved):
I created KOGITO-1967 to add a utility class for wrapping long, unbroken text like the URL. So this is good to go. 👍

cristianonicolai (approved):

cristianonicolai:
@Sara4994 looks ok to you?

cristianonicolai:
@AjayJagan seems like you need to update the snapshot tests

Sara4994 (approved):

Sara4994:
@cristianonicolai @AjayJagan Yep, looks good to me

PR #188, opened 2020-04-16
KOGITO-1108 [Jobs Service] API to allow updating the callback for a Job

tiagodolphine:


cristianonicolai (approved):

jobs-service/src/main/java/org/kie/kogito/jobs/service/resource/JobResource.java
```diff
@@ -64,6 +65,17 @@
                 .thenApply(j -> j.orElseThrow(() -> new RuntimeException("Failed to schedule job " + job)));
     }
 
+    @PATCH
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public CompletionStage<ScheduledJob> patch(Job job) {
```
radtriste: PATCH (like PUT) should target a specific ID, present as a path param.http://restcookbook.com/HTTP%20Methods/patch/

radtriste (changes requested):
Testing was good

One small change to follow REST API guidelines

jobs-service/src/main/java/org/kie/kogito/jobs/service/resource/JobResource.java
tiagodolphine: yep, I agree, added under the path "/{id}"

jobs-service/src/main/java/org/kie/kogito/jobs/service/resource/JobResource.java
```diff
@@ -64,6 +66,25 @@
...
 
+    @PATCH
+    @Path("/{id}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public CompletionStage<ScheduledJob> patch(@PathParam("id") String id, @RequestBody Job job) {
+        LOGGER.debug("REST patch update {}", job);
+
+        //validate parameters for job and id
+        Optional.ofNullable(job)
```
radtriste: This should not be needed on a patch.
the job structure contains only the fields which need to be updated
so id is not present in it

tiagodolphine: updated

jobs-service/src/main/java/org/kie/kogito/jobs/service/resource/JobResource.java
```diff
@@ -64,6 +66,18 @@
                 .thenApply(j -> j.orElseThrow(() -> new RuntimeException("Failed to schedule job " + job)));
     }
 
+    @PATCH
+    @Path("/{id}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public CompletionStage<ScheduledJob> patch(@PathParam("id") String id, @RequestBody Job job) {
+        LOGGER.debug("REST patch update {}", job);
+        return jobRepository.merge(id, ScheduledJob.builder().job(job).build())
```
radtriste: `ScheduledJob.builder().job(job).build()`
could we have something like `ScheduledJob.from(job)` instead that is hiding this implementation ?

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/impl/BaseReactiveJobRepository.java
```diff
@@ -66,4 +69,23 @@ public BaseReactiveJobRepository(Vertx vertx, JobStreams jobStreams) {
...
+
+    @Override
+    public CompletionStage<ScheduledJob> merge(String id, ScheduledJob job) {
+        return Optional.ofNullable(id)
+                .filter(StringUtils::isNotBlank)
+                .filter(s -> StringUtils.isBlank(job.getId()) || s.equals(job.getId()))
+                .map(s -> get(s)
+                        .thenApply(Optional::ofNullable)
+                        .thenApply(current -> current
+                                .map(j -> ScheduledJob
```
radtriste: could we have that on one line ?
make it in fact hard to read like that whereas it is a simple thing...

tiagodolphine: I reformatted the method extracting the merge operation into a new method, take a look it is now more clear IMO.

cristianonicolai:
@radtriste good to merge?

tiagodolphine:
@radtriste can you check the last update, are you ok to merge this? 
@cristianonicolai let's just wait for Tristran approval.

radtriste:
one comment remaining if possible ;)
https://github.com/kiegroup/kogito-apps/pull/188#discussion_r420629005

jobs-service/src/main/java/org/kie/kogito/jobs/service/resource/JobResource.java
tiagodolphine: I didn't get, basically, it is building a new instance of `ScheduledJob` based on a `Job`, that's a superclass, the approach of using the builder is to keep immutability. The `jobRepository` is based on `ScheduledJob` that's why it was necessary.

radtriste: yeah. I was just thinking about creating a direct method `ScheduledJob.from(job)` instead of the whole `ScheduledJob.builder().job(job).build()`.
It would avoid the long line. it is just esthetic I would say

radtriste (approved):

tiagodolphine:
@radtriste I applied the helper method `from`.

radtriste (approved):

PR #196, opened 2020-04-23
[KOGITO-1967] Add word break on long strings

srambach:
KOGITO-1967 https://issues.redhat.com/browse/KOGITO-1967

A long string such as a URL or very long process instance ID might not wrap in certain cases, causing the content to overflow its container. This adds a utility class to be used in these cases. This fix applies it to the ID and Endpoint URL on the Process Instance Details page.


AjayJagan (approved):
:+1:

cristianonicolai (approved):

PR #209, opened 2020-05-05
KOGITO-1602 Enhance bulk selections of items in process instance list

AjayJagan:

- [x] Pull Request contains a link to the JIRA issue
- [x] Pull Request contains a link to any dependent or related Pull Request
- [x] Pull Request contains a description of the issue

JIRA link - https://issues.redhat.com/browse/KOGITO-1602
description: shows '-' symbol(intermediate state of the checkbox) in the checkbox when selecting items in the list partially.

### DIFF
--- a/packages/management-console/server/MockData/graphql.js
+++ b/packages/management-console/server/MockData/graphql.js
@@ -1380,7 +1380,7 @@ module.exports = data = [
     rootProcessInstanceId: null,
     serviceUrl:'http://localhost:4000',
     endpoint: 'http://localhost:4000/',
-    addons: [],
+    addons: ['process-management'],
     error: {
       nodeDefinitionId: 'a1e139d5-4e77-48c9-84ae-3459188e90433n',
       message: 'Something went wrong'
@@ -1498,7 +1498,7 @@ module.exports = data = [
     rootProcessInstanceId: null,
     serviceUrl: 'http://localhost:4000',
     endpoint: 'http://localhost:4000/',
-    addons: [],
+    addons: ['process-management'],
     error: {
       nodeDefinitionId: 'a1e139d5-4e77-48c9-84ae-3459188e90433n',
       message: 'Something went wrong'
@@ -2324,7 +2324,7 @@ module.exports = data = [
     rootProcessInstanceId: null,
     serviceUrl:'http://localhost:4000',
     endpoint: 'http://localhost:4000/',
-    addons: [],
+    addons: ['process-management'],
     error: {
       nodeDefinitionId: 'a1e139d5-4e77-48c9-84ae-3459188e90433n',
       message: 'Something went wrong'
--- a/packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
+++ b/packages/management-console/src/components/Molecules/DataListItemComponent/DataListItemComponent.tsx
@@ -35,7 +35,8 @@ import {
   handleRetry,
   handleAbort,
   isModalOpen,
-  modalToggle
+  modalToggle,
+  lengthChecker
 } from '../../../utils/Utils';
 import EndpointLink from '../EndpointLink/EndpointLink';
 interface IOwnProps {
@@ -121,7 +122,6 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
           ]
         : [...expanded, _id];
     setexpanded(newExpanded);
-
     if (!isLoaded) {
       getChildInstances({
         variables: {
@@ -129,6 +129,7 @@ const DataListItemComponent: React.FC<IOwnProps> = ({
         }
       });
     }
+    lengthChecker(initData, setIsAllChecked);
   };
 
   const onCheckBoxClick = () => {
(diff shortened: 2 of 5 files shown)


nmirasch:
@AjayJagan, works fine but throws this "Warning: A component is changing a controlled input of type checkbox to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component".

cristianonicolai:
@AjayJagan not sure if I'm missing something but when select either _all process_ or _all parent process_, only the first process is selected.
Selection_060

cristianonicolai:
my bad, the other processes are not enabled

cristianonicolai (approved):

nmirasch (approved):
Agree with @cristianonicolai in  merging this code as it is and wait for PF fix to be able to manage null value without any warning.

srambach:
I find it a little odd that when you "select all processes" that you only get the parents unless the children are expanded out to be visible. But if the decision is to select only those that are visible, then I will go with that.
However, then we have an inconsistency. If you select all (say you get 3) and then expand one with a child process that is selectable, the bulk selector still shows an "X" and 3 even though "all" is now 4. So I think that it should either select the child that is now shown, or change the selector to be the "-" to indicate that only a subset of all are selected.
I suspect a similar case exists when you load more instances on the page. However, I can't test this given the current test data because all the instances that are loaded after the first few are not selectable.
@cristianonicolai what do you think?

cristianonicolai:
@srambach I would say to just mark the selector as "-". I'm not sure would make sense to keep selecting new items as they appear.

AjayJagan:
> I find it a little odd that when you "select all processes" that you only get the parents unless the children are expanded out to be visible. But if the decision is to select only those that are visible, then I will go with that.
> However, then we have an inconsistency. If you select all (say you get 3) and then expand one with a child process that is selectable, the bulk selector still shows an "X" and 3 even though "all" is now 4. So I think that it should either select the child that is now shown, or change the selector to be the "-" to indicate that only a subset of all are selected.
> I suspect a similar case exists when you load more instances on the page. However, I can't test this given the current test data because all the instances that are loaded after the first few are not selectable.
> @cristianonicolai what do you think?

Hey Sarah, I have made the changes. Also I have enabled the checkbox for some Active instances to check load more as well. Please have a look

srambach:
Oops, @AjayJagan, now I see multiple bugs. First, when you change the filter, it goes to "-" with no number selected, and nothing is actually checked off. Second, there are multiple ways to mess up the count when you expand and collapse and use the bulk selector. I'll attach a video. Also you can see in the video near the beginning that it said "false selected."
@cristianonicolai it is also strange behavior that you'd not include children in the "select all" except if they are showing...but if you then collapse they still are selected. This whole model might make sense knowing what's going on under the surface, but to a user it's just pretty unpredictable. @lclay2 @kywalker-rh 
BulkSelectorBugs

AjayJagan:
> Oops, @AjayJagan, now I see multiple bugs. First, when you change the filter, it goes to "-" with no number selected, and nothing is actually checked off. Second, there are multiple ways to mess up the count when you expand and collapse and use the bulk selector. I'll attach a video. Also you can see in the video near the beginning that it said "false selected."
> @cristianonicolai it is also strange behavior that you'd not include children in the "select all" except if they are showing...but if you then collapse they still are selected. This whole model might make sense knowing what's going on under the surface, but to a user it's just pretty unpredictable. @lclay2 @kywalker-rh
> BulkSelectorBugs

oh sorry!!, I will look at them now

cristianonicolai:
@srambach I see your point, I just not sure which alternative we would have. Loading all sub-process seems a bit too much overhead. @nmirasch @AjayJagan wdyt if we had to load all items including their child process every time in the UI?

AjayJagan:
> @srambach I see your point, I just not sure which alternative we would have. Loading all sub-process seems a bit too much overhead. @nmirasch @AjayJagan wdyt if we had to load all items including their child process every time in the UI?

In my opinion that would be a heavy load to retrieve all the child instances as well. Consider a scenario where we have 1000s of records. For each change of status we do 2 queries 1.to fetch the parent and 2. to fetch its respective children.   But on the other hand, we can use the Table component if we load the child on start (because they are not loaded async and we can create a new array on each fetch.).

cristianonicolai:
@AjayJagan how many items ( parent processes ) are we loading each time? for the pagination/load more, I guess we have a few different options but the biggest is 50 if I remember correctly?

AjayJagan:
> @AjayJagan how many items ( parent processes ) are we loading each time? for the pagination/load more, I guess we have a few different options but the biggest is 50 if I remember correctly?

yes that is true (50 max)

cristianonicolai:
yeah, that doesnt sound a lot but then we dont know how many sub processes we would load :/

nmirasch:
@cristianonicolai @AjayJagan there are a 'Load 100 more' ! but I agree the problem would be load the subprocess, there's no limitation there..

srambach:
As long as we are talking performance - it already takes quite a while to load nearly nothing when running it as dev. Or is there some artificial lag built in there? With more loading, will the time increase proportionally? Or is there just a lot of overhead to make some connection? I think this affects the decision somewhat.

AjayJagan:
> As long as we are talking performance - it already takes quite a while to load nearly nothing when running it as dev. Or is there some artificial lag built in there? With more loading, will the time increase proportionally? Or is there just a lot of overhead to make some connection? I think this affects the decision somewhat.

asnwering 'is there some artificial lag built in there',  there is a 2 second wait , added it the mock server to view the loading component.

cristianonicolai:
@AjayJagan can you please rebase

AjayJagan:
closing because it needs more rework.. will raise a new PR

PR #210, opened 2020-05-06
KOGITO-994: Load more for Domain Explorer

nmirasch:
https://issues.redhat.com/browse/KOGITO-994

- Add load more to domain explorer view
- Encapsulate design and 'loading' state on LoadMoreComponent
- Updated DataListContainer use of LoadMoreComponent


packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.css
```diff
@@ -0,0 +1,4 @@
+.kogito-management-console-load-more{
```
cristianonicolai:
```suggestion
.kogito-management-console-load-more {
```

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.tsx
```diff
@@ -35,18 +44,32 @@ const LoadMoreComponent: React.FC<IOwnProps> = ({
 
   return (
     <React.Fragment>
```
cristianonicolai: @nmirasch do we still need to wrap it inside a _React.Fragment_ ?

packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx
```diff
@@ -35,6 +36,7 @@ const DomainExplorerDashboard = props => {
     (props.location.state && props.location.state.parameters) || [];
   const rememberedSelections =
     (props.location.state && props.location.state.selected) || [];
+  const pSize = 10;
```
cristianonicolai: I might be wrong but I guess the idea is to use pageSize useState instead?

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.tsx
```diff
@@ -1,18 +1,27 @@
...
+  DataListItem,
+  DataListCell,
+  Spinner
+} from '@patternfly/react-core';
+import './LoadMoreComponent.css';
 
 interface IOwnProps {
   offset: number;
   setOffset: (offset: number) => void;
   getProcessInstances: (initval: number, pageSize: number) => void;
```
cristianonicolai: maybe we should call this like callback?

packages/management-console/src/components/Organisms/DomainExplorerColumnPicker/DomainExplorerColumnPicker.tsx
```diff
@@ -355,9 +382,18 @@ const DomainExplorerColumnPicker: React.FC<IOwnProps> = ({
     return unique;
   }
 
+  const onResetQuery = _parameters => {
+    setOffsetVal(0);
+    offsetVal = 0;
```
cristianonicolai, line 387: similar here, I think we need to use the useState hooks only, no?

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.css
nmirasch: done!

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.tsx
nmirasch: removed, yes it wasn't necessary.  :-)

packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx
nmirasch: I have updated this renaming to defaultPageSize and adding useState

cristianonicolai:
@nmirasch maybe add some mocked data to see it in action? :)

srambach:
I'm seeing "Load 10 more" not actually add 10 more, but instead it reloads just 10 without the loading buttons. 

The "Load 20 more" button does add more but I can't tell if it would add 20 more for a total of 30 since there aren't that many. 

Also, there is quite a delay in loading. Can we add the spinner to the button while loading, then redraw the screen? The -xs size spinner should fit into the button. https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/data-loading#main-content Or, we could use a spinner below the list, indicating more are being fetched, but from a UX standpoint, it's probably easier to show on the button.

nmirasch:
@cristianonicolai  I'm trying to add more mock data to see this in action but for the moment the testing have to be against the remote data index with real data.. .
@srambach  
- About the message 'Load X more' are the same as was in Process instances explorer, in fact it's using the same component. But you mean we have to change the 'Load X more" -> "Add X more"?
- the 'load more' buttons bar would appear if could be more that the showed ones.. The point is not to calculate how much items are there. Those buttons meaning would be something like 'Try to load 20 more".
- I don't understand the last thing about the spinner. It have to be shown when the query to load more items is performed, let me re-check that..

srambach:
Sorry :)  what I'm trying to say is:
- When the "Load X more" button is pressed, I'd expect to get X _more_ than I see now, unless that is more than there are in total.
- When I click the button, first the buttons go away, then after a pause, I see more items in the list. What would be better is immediately after the button is clicked, a spinner is added to the button to show work in progress, and then the new items are loaded. If there are no more items, then the buttons can go away when the new items are added to the list. Does that make sense? Is it possible?

nmirasch:
@srambach thanks for your comments, makes absolutely sense to me . We can go throw this at today meeting, to see this in action and see what have to be different, maybe there are some bug that I'm missing right now.

cristianonicolai (approved):

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.css
```diff
@@ -0,0 +1,4 @@
+.kogito-management-console-load-more {
```
srambach: We haven't been 100% consistent with class names, but let's try to stick with `kogito-management-console__` plus the component (and any elements within that are needed) for now, so something like `kogito-management-console__load-more` here. (and where it's used)

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.css
```diff
@@ -0,0 +1,4 @@
...
+  margin-bottom: 24px;
```
srambach: We can use our spacer variables here
`margin-bottom: var(--pf-global--spacer--lg);`

AjayJagan (approved):

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.css
nmirasch: done

packages/management-console/src/components/Atoms/LoadMoreComponent/LoadMoreComponent.css
nmirasch: updated!

nmirasch:
@srambach  Thanks, css updated! @AjayJagan I also added some LoadMoreComponet tests too.

srambach (approved):

Sara4994:
@nmirasch looks good to me :+1:

Sara4994 (approved):

PR #214, opened 2020-05-12
WIP KOGITO-1309 OUIA example

jstastny-cz:
Please see an example of Open UI Automation principles applied to our project.

Main things to note:
* attributes needed to be set on our components
  * data-ouia-component-type
  * data-ouia-component-id
* cypress tests
  * helper commands 'byOuiaId', 'byOuiaType'

To run the tests it's best to go into packages/management-console and use script 'cypress:open-with-dev-server'. The list of 'specs' will appear, run the 'process-instances-page-filters-spec.js'.

### DIFF
--- a/packages/management-console/cypress/support/commands.js
+++ b/packages/management-console/cypress/support/commands.js
@@ -23,3 +23,20 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add("byOuiaType", { prevSubject: 'optional' }, (subject, type, criteria = '', options = {}) => {
+    if (subject) {
+        cy.wrap(subject).get('*[data-ouia-component-type="' + type + '"]' + ((criteria) ? ':' + criteria : ''), options)
+    } else {
+        cy.get('*[data-ouia-component-type="' + type + '"]' + ((criteria) ? ':' + criteria : ''), options)
+    }
+})
+
+
+Cypress.Commands.add("byOuiaId", { prevSubject: 'optional' }, (subject, id, criteria = '', options = {}) => {
+    if (subject) {
+        cy.wrap(subject).get('*[data-ouia-component-id="' + id + '"]' + ((criteria) ? ':' + criteria : ''), options)
+    } else {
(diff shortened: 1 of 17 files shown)


packages/management-console/cypress/support/commands.js
```diff
@@ -23,3 +23,20 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add("byOuiaType", { prevSubject: 'optional' }, (subject, type, criteria = '', options = {}) => {
```
jomarko, line 27: actual invocation seems to use at most three parameters, do we need `options`?

packages/management-console/cypress/support/commands.js
```diff
@@ -23,3 +23,20 @@
...
+    if (subject) {
+        cy.wrap(subject).get('*[data-ouia-component-type="' + type + '"]' + ((criteria) ? ':' + criteria : ''), options)
+    } else {
+        cy.get('*[data-ouia-component-type="' + type + '"]' + ((criteria) ? ':' + criteria : ''), options)
+    }
+})
+
+
+Cypress.Commands.add("byOuiaId", { prevSubject: 'optional' }, (subject, id, criteria = '', options = {}) => {
```
jomarko, line 36: actual invocation seems to use at most one parameter, do we need `id`, `criteria`, `options`?

packages/management-console/cypress/support/commands.js
jstastny-cz: hm the intention here was to offer option to pass options like in:
```
cy.byOuiaType('empty-state', 'contains("Loading process instances")', { timeout: 5000 })
```

The options are simply passed to cy.get().
Though I need still to check if the options do really work as I expect them to :-)

packages/management-console/cypress/support/commands.js
jstastny-cz: it's true that in case of Id, which should be unique the criteria do not make maybe much sense -> in such case simple id match + assertion should work just fine (cause we're probably not gonna filter more than a single element).

And passing the options to cy.get() due to tweaking options like timeout as mentioned above.

PR #215, opened 2020-05-12
KOGITO-1166 Unable to execute a job with expiration time in past

tiagodolphine:
Force a job with an expiration time in the past to be executed based on a configuration property.

### DIFF
--- a/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
+++ b/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
@@ -56,7 +56,7 @@ public abstract class BaseTimerJobScheduler implements ReactiveJobScheduler<Sche
     long maxIntervalLimitToRetryMillis;
 
     /**
-     * Flag to allow anf force a job with expirationTime in the past to be executed immediately. If false and
+     * Flag to allow and force a job with expirationTime in the past to be executed immediately. If false an
      * exception will be thrown.
      */
     @ConfigProperty(name = "kogito.jobs-service.forceExecuteExpiredJobs")
--- a/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
+++ b/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
@@ -56,6 +56,13 @@ public abstract class BaseTimerJobScheduler implements ReactiveJobScheduler<Sche
     long maxIntervalLimitToRetryMillis;
 
     /**
+     * Flag to allow anf force a job with expirationTime in the past to be executed immediately. If false and
+     * exception will be thrown.
+     */
+    @ConfigProperty(name = "kogito.jobs-service.forceExecuteExpiredJobs")
+    Optional<Boolean> forceExecuteExpiredJobs;
+
+    /**
      * The current chunk size  in minutes the scheduler handles, it is used to keep a limit number of jobs scheduled
      * in the in-memory scheduler.
      */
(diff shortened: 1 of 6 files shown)


cristianonicolai (approved):

jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
```diff
@@ -55,6 +55,13 @@
     @ConfigProperty(name = "kogito.jobs-service.maxIntervalLimitToRetryMillis")
     long maxIntervalLimitToRetryMillis;
 
+    /**
+     * Flag to allow anf force a job with expirationTime in the past to be executed immediately. If false and
+     * exception will be thrown.
```
radtriste:
```suggestion
     * Flag to allow and force a job with expirationTime in the past to be executed immediately. If false an
     * exception will be thrown.
```

jobs-service/src/main/resources/application.properties
```diff
@@ -60,6 +60,7 @@ kogito.jobs-service.backoffRetryMillis=1000
 kogito.service.url=http://localhost:8080
 kogito.jobs-service.schedulerChunkInMinutes=10
 kogito.jobs-service.loadJobIntervalInMinutes=10
+kogito.jobs-service.forceExecuteExpiredJobs=true
```
radtriste, line 64: should we enable it by default ?

jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
```diff
@@ -99,7 +99,8 @@ void loadScheduledJobs() {
     }
 
     private PublisherBuilder<ScheduledJob> loadJobsInCurrentChunk() {
-        return repository.findByStatusBetweenDatesOrderByPriority(DateUtil.now(), DateUtil.now().plusMinutes(schedulerChunkInMinutes),
+        return repository.findByStatusBetweenDatesOrderByPriority(DateUtil.now().minusDays(1),
```
radtriste: only minus one day ? this is quite arbitrary ... shouldn't that be also configurable ?

tiagodolphine: yep, I agree, in fact I added this during the tests to allow when loading persisted jobs, to schedule "expired" ones when the force configuration set, but it should be a configuration parameter indeed. Any suggestion for the default value/time for instance minutes, hours, days, etc.

jobs-service/src/main/resources/application.properties
tiagodolphine: we need to decide it, to keep the jbpm behavior it should be true, but we can change with configuration, so up to us to define the default value, wdyt?

jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
```diff
@@ -55,6 +55,13 @@
     @ConfigProperty(name = "kogito.jobs-service.maxIntervalLimitToRetryMillis")
     long maxIntervalLimitToRetryMillis;
 
+    /**
+     * Flag to allow and force a job with expirationTime in the past to be executed immediately. If false an
+     * exception will be thrown.
+     */
+    @ConfigProperty(name = "kogito.jobs-service.forceExecuteExpiredJobs")
+    Optional<Boolean> forceExecuteExpiredJobs;
```
sutaakar, line 63: Is there any specific reason to use Optional here?
The property is specified in application.properties anyway.

tiagodolphine: The property is used on the `calculateInterval` method where I prefer to use a functional approach there, otherwise, I would need to use Optional.of()... anyway any specific reason to not use Optional?

sutaakar: I was just curious what is the rationale behind.
Ok to be kept like that.

sutaakar (approved):

jobs-service/src/main/resources/application.properties
radtriste: let's keep jbpm behavior so we don't confuse the users

radtriste (approved):

PR #217, opened 2020-05-14
KOGITO-2190 Make pages OUIA:Page compliant

jstastny-cz:
Hello, please review. This is an attempt to make management-console OUIA:Page compliant. The rules are straight-forward in this case:

There are following rules for OUIA:Page compliance (1):
* <body> element to have attributes:
  * data-ouia-page-type -> based on the page we're on (process-instances vs. domain-modeler)
  * data-ouia-page-object-id (OPTIONAL) -> in process details or for specific domain-name
  * data-ouia-page-action -> for pages that represent a certain action (optional when there's a single one) - we probably don't have such actions like 'edit', or similar
* page fragments identified by:
  * data-ouia-header="true" - For the header
  * data-ouia-footer="true" - For the footer
  * data-ouia-main="true" - For the main content pane
  * data-ouia-navigation="true" - For the main navigation pane
  * Header, Footer and Main MUST NOT ever be nested inside each other.
  * Navigation MAY live inside any of the other content areas but must appear only once.

(1) https://ouia.readthedocs.io/en/latest/README.html#specification-parts

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -10,3 +10,4 @@ export {
 } from './src/components/Molecules/ServerUnavailable/ServerUnavailable';
 export { default as NoData } from './src/components/Molecules/NoData/NoData';
 export { default as ServerErrors } from './src/components/Molecules/ServerErrors/ServerErrors';
+export * from './src/utils/OuiaUtils'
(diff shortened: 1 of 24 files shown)


cristianonicolai (approved):

Sara4994 (approved):

packages/management-console/src/utils/Utils.tsx
```diff
@@ -202,3 +202,15 @@ export const modalToggle = (
     return handleRetryModalToggle;
   }
 };
+
+export const ouiaPageTypeAndObjectId = (
```
nmirasch: @jstastny-cz  this code would be a good candidate to go to the common package to be shared by management-console and task-console.

nmirasch (approved):
Even the ouiaPageTypeAndObjectId location would be better in the common package to be shared with the task-console, that is something we can do it in next steps. Then Approving, Thanks!

AjayJagan (approved):

jstastny-cz:
Now that we have the common package merged, I'll move the function. I have to resolve conflicts anyway.

jstastny-cz:
Merge after #228

PR #224, opened 2020-05-20
KOGITO-2230 - add trusty project

r00ta:
Link to jira: https://issues.redhat.com/browse/KOGITO-2230

This is the first PR to include the trusty project. Some PRs will follow for the "real" implementation. I will try to keep the PRs <= 500 lines, so that it's easier for everybody to review them.


r00ta:
@danielezonca @jiripetrlik ping

pom.xml
```diff
@@ -62,6 +62,7 @@
     <module>jobs-service</module>
     <module>data-index</module>
     <module>management-console</module>
+    <module>trusty</module>
```
danielezonca: Can you please rename it as `trusty-service`?

trusty/execution-service/src/main/java/org/kie/kogito/trusty/execution/ExecutionService.java
```diff
@@ -0,0 +1,6 @@
+package org.kie.kogito.trusty.execution;
```
danielezonca: License

trusty/execution-service/src/main/java/org/kie/kogito/trusty/execution/IExecutionService.java
```diff
@@ -0,0 +1,5 @@
+package org.kie.kogito.trusty.execution;
```
danielezonca: License

trusty/execution-service/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.swagger-ui.always-include=true
```
danielezonca: I think this should not be hardcoded in the service

trusty/execution-service/src/test/java/org/kie/kogito/trusty/execution/ExecutionServiceTest.java
```diff
@@ -0,0 +1,5 @@
+package org.kie.kogito.trusty.execution;
```
danielezonca: License

trusty/execution-service/src/main/java/org/kie/kogito/trusty/execution/api/ExecutionApi.java
```diff
@@ -0,0 +1,10 @@
+package org.kie.kogito.trusty.execution.api;
```
danielezonca: License

trusty/explanation-service/src/main/java/org/kie/kogito/trusty/explanation/ExplanationService.java
```diff
@@ -0,0 +1,5 @@
+package org.kie.kogito.trusty.explanation;
```
danielezonca: License

trusty/explanation-service/src/main/java/org/kie/kogito/trusty/explanation/IExplanationService.java
```diff
@@ -0,0 +1,5 @@
+package org.kie.kogito.trusty.explanation;
```
danielezonca: License

trusty/explanation-service/src/main/java/org/kie/kogito/trusty/explanation/api/ExplanationApi.java
```diff
@@ -0,0 +1,7 @@
+package org.kie.kogito.trusty.explanation.api;
```
danielezonca: License

trusty/explanation-service/src/test/java/org/kie/kogito/trusty/explanation/ExplanationServiceTest.java
```diff
@@ -0,0 +1,5 @@
+package org.kie.kogito.trusty.explanation;
```
danielezonca: License

trusty/explanation-service/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.swagger-ui.always-include=true
```
danielezonca: I think this should not be hardcoded in the service

danielezonca (commented):
Many minor comments plus some general considerations:
- Please create two different top level module: `trusty-service` and `explainable-service`
- Please add a README both for `trusty-service` and `explainable-service` with minimal introduction
- Please structure the code in multiple modules as we briefly discussed:
    - `core` with proper service logic
    - `api` with REST endpoints
    - `persistence-infinispan` only if needed
    - `dist` with Quarkus final project

danielezonca:
@cristianonicolai 
We are starting to implement some additional services part of TrustyAI initiative based on some code from [here](https://github.com/kiegroup/trusty-ai-sandbox)
As you can see we are just creating modules and folder for now and I would like to apply to both the `experimental` flag we discuss during the yesterday call.
How do you want to proceed? I think this is a good candidate :)

pom.xml
r00ta: I created a top level module called `trusty` and the two submodules `explanation-service` and `execution-service`. Do we want keep all the services as top level @danielezonca ?

r00ta: Are we 100% sure to keep all the services as top level? It might be not very clear imo

r00ta: Atm the data index is structured as following: 
data-index
|- data-index-service
|- data-index-storage
My idea of the trusty initiative was to create a top level module called `trusty` and then create submodules depending on our needs. Submodules might be services, libraries (like the explanability library) and so on. Since this repo contains "kogito apps", I think we should keep as top level the apps and all the services of all the apps. This is going to be messy if all the apps will do the same..

cristianonicolai: I think it makes sense to follow the same as data index, a higher-level grouping, but then just not sure if we want to add _trusty-_ as prefix for the services as well.

cristianonicolai (commented):
@danielezonca I think this is fine, just check how we work with the UI side of things if you want to reuse react components. I would suggest, if we want to mark this as experimental, to create a profile that adds the _trusty_ module only when `-Pexperimental` is used. wdyt?

jiripetrlik (approved):
Hello @r00ta ! Thank you for your PR. I've seen the report of Sonar cloud that there is zero test coverage at the moment. And it is probably fine because this PR looks like a skeleton for future development of service. However, we will need to define some testing strategy for Trusty AI services.

r00ta:
> Hello @r00ta ! Thank you for your PR. I've seen the report of Sonar cloud that there is zero test coverage at the moment. And it is probably fine because this PR looks like a skeleton for future development of service. However, we will need to define some testing strategy for Trusty AI services.

Hi @jiripetrlik , Yes, indeed. This is just a small PR for the repo structure. Regarding the testing of Trusty AI services we can setup a quick call just to be aligned.

danielezonca:
> @danielezonca I think this is fine, just check how we work with the UI side of things if you want to reuse react components. I would suggest, if we want to mark this as experimental, to create a profile that adds the _trusty_ module only when `-Pexperimental` is used. wdyt?

These two services are just for the backend part, soon we will add also the Audit UI module and we are already looking at the common module and similar :)

@r00ta can you please update root pom to implement `experimental` profile?
It should be enough to remove `trusty` and `explainability-service` from [here](https://github.com/kiegroup/kogito-apps/blob/e5c617d24d77abcce0e848f9b42f0d410021624e/pom.xml#L65-L66) and add a section like this
```
  <profiles>
    <profile>
      <id>experimental</id>
      <modules>
       <module>explainability-service</module>
       <module>trusty</module>
      </modules>
    </profile>
  </profiles>
```
After that it should be possible to perform a normal maven build like `mvn clean package` without experimental features and `mvn clean package -Pexperimental` to enable them

@cristianonicolai  what about CI? I would like to have the possibility to build experimental features too

cristianonicolai:
@danielezonca yep, that still open to discussion I guess. We could have at least one daily job that includes _experimental_ profile.

r00ta:
@danielezonca @cristianonicolai Is it feasible to setup a jenkins job that runs only if the branch of the PR matches a prefix or a pattern? Otherwise for the time being we can still avoid using the `experimental` feature and use the standard build. 
My next PR already has some tests and it would be nice to have the CI running with also the code coverage and so on..

r00ta:
@danielezonca @cristianonicolai can we merge?

cristianonicolai (approved):

cristianonicolai:
@r00ta it should be possible to set it up, need to investigate.

danielezonca:
@cristianonicolai 
I think we can clone the just you recently created for sonarcloud ([link](https://github.com/kiegroup/kogito-apps/pull/226)). Wdyt?

In the meanwhile I propose to merge this PR (as soon as we have a green build) and proceed

cristianonicolai:
@danielezonca you can use https://github.com/kiegroup/kogito-runtimes/blob/master/Jenkinsfile.drools as a reference, because we need to build all repos with the _experimental_ profile, and also add it to _kogito-runtimes_ instead.

PR #227, opened 2020-05-21
KOGITO-2104: Configured storybook for common package

Sara4994:
Hello! please review, this PR intended to configure storybook for react components that are present in common package, just to visually display how those components are look like.

Storybook allows you to run and view your React components in an isolated manner and it used for development purpose.

Storybook runs its own web server, separate from your React app, and provides a catalog view for the components that you choose to use it with.

JIRA issue : https://issues.redhat.com/browse/KOGITO-2104


srambach:
I'd rather to see the components in isolation than within a page structure. This has the advantage of showing exactly what the component is (for re-use). In addition, it would make it more clear whether differences such as the background color on No Data is due to the surrounding structure or in the component itself.
Also, the "Server unavailable" appears to have a page within a page.

packages/common/src/components/Atoms/BrandLogo/BrandLogo.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    </div>
+  ))
+  .add('Task Console', () => (
+    <div style={{ backgroundColor: 'black' }}>
+      <BrandLogo
+        src={taskConsoleLogo}
+        alt="Task Console"
+        brandClick={() => null}
+      />
+    </div>
```
srambach, line 24: On the logos, is it possible to give storybook an option to change the preview background to black, rather than adding a div around them? It might be less misleading about what is and isn't part of the component.

srambach:
Does it make sense to put some instructions in the readme about how to run storybook?

Sara4994:
Created a new PR for the same https://github.com/kiegroup/kogito-apps/pull/286. hence closing this one

PR #230, opened 2020-05-21
Kogito 2247 - Trusty service executions endpoint resource

r00ta:
Link to jira: https://issues.redhat.com/browse/KOGITO-2247

The aim of this PR is to implement the `execution` resource endpoint that exposes the list of execution headers. It's possible to query the endpoint by time range (and optionally by prefix on the executionId) and it's possible to set some pagination parameters. 
The endpoint is then unit tested.

In case the datetime and/or the pagination parameters are not valid, then a 400 `BAD_REQUEST` is returned.  

The contract looks like the following: 
Screenshot from 2020-05-21 16-13-10


trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/responses/ExecutionHeaderResponse.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Date;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.kie.kogito.trusty.service.models.Execution;
+
+/**
+ * An execution header.
+ */
+public class ExecutionHeaderResponse {
```
danielezonca, line 31: Why do you want to split "internal" representation of the execution and the external one?
Considering that this service is designed to serve these kind of data I do not expect this to be needed. I would start simple with only one version and then it is always possible to decouple it.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/responses/ExecutionTypeEnumResponse.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package org.kie.kogito.trusty.service.responses;
+
+import org.kie.kogito.trusty.service.models.ExecutionTypeEnum;
+
+/**
+ * The execution enum type.
+ */
+public enum ExecutionTypeEnumResponse {
```
danielezonca: Same comment as `ExecutionHeaderResponse` class

danielezonca (commented):
Two small considerations and a general comment: please update all `Date` usages to leverage on Java8 version that is better designed and thread safe

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/responses/ExecutionHeaderResponse.java
r00ta: It's already needed actually. All the executions we will manipulate will be `Decision` (and in the future `Process`) objects.  The `Decision` object extends the `Execution` class, and the `ExecutionHeaderResponse` ensures that only the `Execution` information is exported.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/responses/ExecutionTypeEnumResponse.java
r00ta: For this class we can remove the copy, that's true. In general I prefer to split the internal model and the reponses, but for this enum I can live with it. I will remove

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1.java
```diff
@@ -0,0 +1,135 @@
...
+     * @param from   The start datetime.
+     * @param to     The end datetime.
+     * @param limit  The maximum (non-negative) number of items to be returned.
+     * @param offset The non-negative pagination offset.
+     * @param prefix The executionId prefix to be matched in the search.
+     * @return The execution headers that satisfy the time range, pagination and prefix conditions.
+     */
+    @GET
+    @APIResponses(value = {
+            @APIResponse(description = "Returns the execution headers.", responseCode = "200", content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(type = SchemaType.OBJECT, implementation = ExecutionsResponse.class))),
```
jiripetrlik: Please split these long lines to make the code more readable.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/models/Execution.java
```diff
@@ -0,0 +1,111 @@
...
+    @JsonProperty("executedModelName")
+    private String executedModelName;
+
+    @JsonProperty("executionType")
+    private ExecutionTypeEnum executionType;
+
+    public Execution() {
+    }
+
+    public Execution(String executionId, Long executionTimestamp, boolean hasSucceeded, String executorName, String executedModelName, ExecutionTypeEnum executionType) {
```
jiripetrlik: Please split.

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1Test.java
```diff
@@ -0,0 +1,94 @@
...
+@QuarkusTest
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class ExecutionsApiV1Test {
+
+    @InjectMock
+    ITrustyService executionService;
+
+    @Test
+    void GivenRequestWithoutLimitAndOffsetParameters_WhenExecutionEndpointIsCalled_ThenTheDefaultValuesAreCorrect() {
+        Mockito.when(executionService.getExecutionHeaders(any(OffsetDateTime.class), any(OffsetDateTime.class), any(Integer.class), any(Integer.class), any(String.class))).thenReturn(new ArrayList<>());
```
jiripetrlik, line 51: Please split this line.

pom.xml
```diff
@@ -62,6 +62,8 @@
     <module>jobs-service</module>
     <module>data-index</module>
     <module>management-console</module>
+    <module>explainability-service</module>
+    <module>trusty</module>
```
jiripetrlik, line 66: Wouldn't it be better to be more specific with the name of this model. Can we for example use something like trusty-ai?

jiripetrlik (approved):
Thank you for your PR. I have only a few minor comments.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1.java
r00ta: Done

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/models/Execution.java
r00ta: Done

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1Test.java
r00ta: Done

pom.xml
r00ta: Fine for me, @danielezonca wdyt?

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1.java
```diff
@@ -0,0 +1,138 @@
...
+                    description = "Start datetime for the lookup. Date in the format \"yyyy-MM-dd'T'HH:mm:ssZ\"",
+                    required = true,
+                    schema = @Schema(implementation = String.class)
+            ) @QueryParam("from") String from,
+            @Parameter(
+                    name = "to",
+                    description = "End datetime for the lookup. Date in the format \"yyyy-MM-dd'T'HH:mm:ssZ\"",
+                    required = true,
+                    schema = @Schema(implementation = String.class)
+            ) @QueryParam("to") String to,
```
danielezonca: I just noticed that there are two mandatory query parameters. In general I don't like GET methods that fails with no parameters.
Is it possible to default to `from` yesterday `to` now?

pom.xml
danielezonca: Personally I prefer `trusty-service` if we want to better identify it. But then (as @r00ta showed me in the previous PR) we will have `trusty-service/trusty-service-service` :smile: 
I don't think we have a defined naming convention for these services: we have `jobs-service` and `explainable-service` but then we have `data-index` without any suffix so I would stay with `trusty` for now.
@jiripetrlik Wdyt?

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1.java
r00ta: What about `from` = `yesterday` and `to` = `now` by default?

r00ta: Done

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExecutionsApiV1.java
```diff
@@ -0,0 +1,135 @@
...
+
+        if (limit < 0 || offset < 0) {
+            return Response.status(Response.Status.BAD_REQUEST.getStatusCode(), "Pagination parameters can not have negative values.").build();
+        }
+
+        OffsetDateTime fromDate;
+        OffsetDateTime toDate;
+        try {
+            fromDate = from.equals("yesterday") ? OffsetDateTime.now(ZoneOffset.UTC).minusDays(1) :  ZonedDateTime.parse(from, DateTimeFormatter.ISO_OFFSET_DATE_TIME).toOffsetDateTime();
+            toDate = to.equals("now") ? OffsetDateTime.now(ZoneOffset.UTC) : ZonedDateTime.parse(to, DateTimeFormatter.ISO_OFFSET_DATE_TIME).toOffsetDateTime();
```
danielezonca: Can you please create a single `parteDate` method that takes a `String` and returns a `OffsetDateTime`?

danielezonca (approved):
Thanks for the changes, approving :+1:

PR #232, opened 2020-05-25
Kogito 2275 - Trusty storage abstraction + mongo 

r00ta:
Jira link: https://issues.redhat.com/browse/KOGITO-2275

This PR includes: 
1) implementation of the endpoint `/executions/decisions/{key}` to retrieve a decision header. 
2) Storage layer abstraction structure: we have a module called `trusty-storage-api` that contains: the definition of the storage abstraction (`search`, `create`, `update`, `delete`) and the abstraction to make queries (`TrustyQuery`).
3) The implementation of the Mongo layer + tests.


gitgabrio:
Hi @r00ta 
IIUC the abstraction should decouple a common API from any specific implementation, so that actual implementation could be choosenn at later time.
But it seem to me that in this current implementation the API is strongly coupled with mongo implementation:
looking at trusty-storage/pom.xml
` <modules>
    <module>trusty-storage-api</module>
    <module>trusty-storage-mongo</module>
  </modules>`
That means whenever trusty-storage-api is built, also trusty-storage-mongo it is.
If for whatever reason trusty-storage-mongo build fails, the whole trusty-storage fails - in a way it seems that the API **depends** on the implementation (which I think in general should not happen).
If, on the other side, this is by design (i.e. trusty-storage must depend on trusty-storage-mongo) putting the API in a separated module would just confuse user - because behind an happarently composition design there is an actual monolithic block.
@danielezonca WDYT ?

r00ta:
> Hi @r00ta
> IIUC the abstraction should decouple a common API from any specific implementation, so that actual implementation could be choosenn at later time.
> But it seem to me that in this current implementation the API is strongly coupled with mongo implementation:
> looking at trusty-storage/pom.xml
> ` <modules> <module>trusty-storage-api</module> <module>trusty-storage-mongo</module> </modules>`
> That means whenever trusty-storage-api is built, also trusty-storage-mongo it is.
> If for whatever reason trusty-storage-mongo build fails, the whole trusty-storage fails - in a way it seems that the API **depends** on the implementation (which I think in general should not happen).
> If, on the other side, this is by design (i.e. trusty-storage must depend on trusty-storage-mongo) putting the API in a separated module would just confuse user - because behind an happarently composition design there is an actual monolithic block.
> @danielezonca WDYT ?

Hi @gitgabrio , sorry, I don't get it. I don't see any dependency in the `trusty-storage-api` module, how can it depend on the implementation?

gitgabrio:
Hi @r00ta , let me exaplain. 
IMO there are two kinds of "dependency": explicit and implicit.
Explicit ones are the one you declare directly (e.g. a "dependency" tag inside a pom, or an "import" inside a java class).
Implicit ones are the "dependencies" that are needed for the full functionality of one component, but are not declared directly.
In that situation, `trusty-storage-api` is declared inside the `trusty-storage-module`. But this module also declare `trusty-storage-mongo`.
If `trusty-storage-mongo` fails to build (for whatever reason) all the `trusty-storage-module` build would fail.
So, whatever is needed by `trusty-storage-mongo` to build is needed by its parent module - but also siblings - modules.
Said differently
1) `quarkus-mongodb-client` is needed to build `trusty-storage-mongo`
2) `trusty-storage-mongo` is needed to have a successful built of `trusty-storage`
3) a successful build of `trusty-storage` is needed to have `trusty-storage-api` 
4) `quarkus-mongo-dbclient` is needed to have `trusty-storage-api`
5) `trusty-storage-api` actually (implicitly) depends on `trusty-storage-mongo` (and all its dependencies).

Hope to have been clear. If I've been not, or you do not agree, I would be happy to talk of that: it is very possible that I'm wrong and I'm misreading something :)

r00ta:
> Hi @r00ta , let me exaplain.
> IMO there are two kinds of "dependency": explicit and implicit.
> Explicit ones are the one you declare directly (e.g. a "dependency" tag inside a pom, or an "import" inside a java class).
> Implicit ones are the "dependencies" that are needed for the full functionality of one component, but are not declared directly.
> In that situation, `trusty-storage-api` is declared inside the `trusty-storage-module`. But this module also declare `trusty-storage-mongo`.
> If `trusty-storage-mongo` fails to build (for whatever reason) all the `trusty-storage-module` build would fail.
> So, whatever is needed by `trusty-storage-mongo` to build is needed by its parent module - but also siblings - modules.
> Said differently
> 
> 1. `quarkus-mongodb-client` is needed to build `trusty-storage-mongo`
> 2. `trusty-storage-mongo` is needed to have a successful built of `trusty-storage`
> 3. a successful build of `trusty-storage` is needed to have `trusty-storage-api`
> 4. `quarkus-mongo-dbclient` is needed to have `trusty-storage-api`
> 5. `trusty-storage-api` actually (implicitly) depends on `trusty-storage-mongo` (and all its dependencies).
> 
> Hope to have been clear. If I've been not, or you do not agree, I would be happy to talk of that: it is very possible that I'm wrong and I'm misreading something :)

Thanks for the clarification @gitgabrio , I'm happy to receive some feedbacks ;) IMHO this is acceptable, with this concept of implicit dependency we might even argue that the trusty services depends on the other services (like the `data-index` and the `management-console`) because the top pom build all the apps.  We might generalize your concerns to all the multi-modules projects actually. 
Personally, I don't think this is a big problem at the moment. But if most of us believe this is an issue, I think we should move the discussion to a bigger audience because this concern impacts basically all the kogito repositories

gitgabrio:
Hi @r00ta, I'm not sure if it is exactly the same situation.
If the logic behind submodules would just be "build'em all", then we could  collapse all sub-modules in one single project :).
If there is a reason to split a repository in submodules (and I personally think most of the time there is), then it make sense to me to keep things clearly separated.
Anyway, a possible approach for this case could be to put all concrete implementations inside the `trust-storage-module`, and that would just give a different meaning (compared to the one I was thinking about) to the module itself.

r00ta:
> Hi @r00ta, I'm not sure if it is exactly the same situation.
> If the logic behind submodules would just be "build'em all", then we could collapse all sub-modules in one single project :).
> If there is a reason to split a repository in submodules (and I personally think most of the time there is), then it make sense to me to keep things clearly separated.
> Anyway, a possible approach for this case could be to put all concrete implementations inside the `trust-storage-module`, and that would just give a different meaning (compared to the one I was thinking about) to the module itself.

Right now the main reason to have different submodules for each implementation is that the storage to be used is selected by CDI. Since this is going to be a service that is provided by the operator (and customers/developers will not probably interact with it), it would be nice to find another way to select the right storage to be used. 
AFAIK there is a bigger discussion with also the kogito team that has the same challenge with the data index service

PR #235, opened 2020-05-26
KOGITO-2279 - Add trusty CI pipeline

r00ta:
Jira link: https://issues.redhat.com/browse/KOGITO-2279

Hi, this PR adds the CI for the trusty services (flagged as experimental). The github action is run if and only if a pull request changes anything in the directories `trusty` and `explanability-service`. 

In my fork you can find some tests that I've made https://github.com/r00ta/kogito-apps/pulls :
1) the `trustyCI.yml` is already in my master branch
2) in the first pull request nothing has changed in the two modules mentioned above -> the CI is not run. 
3) in the second PR a file in the `trusty` module has changed -> the CI is run
4) in the third PR a file in the `explainability-service` has changed -> the CI is run. 

We propose to use this github action instead of a daily job running on jenkins. Since the trusty services are flagged as experimental, we just need a CI to run our tests in the PRs. This CI is going to be removed as soon as the `experimental` flag is removed from the trusty services. 

Looking forward your feedbacks :)

### DIFF
--- a/.github/workflows/trustyCI.yml
+++ b/.github/workflows/trustyCI.yml
@@ -1,0 +1,23 @@
+name: Trusty experimental CI
+
+on:
+  pull_request:
+    branches: [ master ]
+    paths:
+    - 'trusty/**'
+    - 'explainability-service/**'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.11
+    - name: Build kogito-runtimes
+      run: git clone https://github.com/kiegroup/kogito-runtimes.git /tmp/kogito-runtimes && mvn clean install -DskipTests -f /tmp/kogito-runtimes/pom.xml     
+    - name: Build trusty service
+      run: mvn -B package --file trusty/pom.xml
+    - name: Build explanability service
+      run: mvn -B package --file explainability-service/pom.xml 
--- a/.github/workflows/trustyCI.yml
+++ b/.github/workflows/trustyCI.yml
@@ -17,7 +17,5 @@ jobs:
         java-version: 1.11
     - name: Build kogito-runtimes
       run: git clone https://github.com/kiegroup/kogito-runtimes.git /tmp/kogito-runtimes && mvn clean install -DskipTests -f /tmp/kogito-runtimes/pom.xml     
-    - name: Build trusty service
-      run: mvn -B package --file trusty/pom.xml
-    - name: Build explanability service
-      run: mvn -B package --file explainability-service/pom.xml 
+    - name: Build kogito-apps with experimental flag
+      run: mvn -B package --file pom.xml -Pexperimental


.github/workflows/trustyCI.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: Set up JDK 1.11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.11
+    - name: Build kogito-runtimes
+      run: git clone https://github.com/kiegroup/kogito-runtimes.git /tmp/kogito-runtimes && mvn clean install -DskipTests -f /tmp/kogito-runtimes/pom.xml     
+    - name: Build trusty service
+      run: mvn -B package --file trusty/pom.xml
+    - name: Build explanability service
+      run: mvn -B package --file explainability-service/pom.xml
```
danielezonca: Can you please just build the whole project with `-Pexperimental` instead of build single modules?

.github/workflows/trustyCI.yml
```diff
@@ -0,0 +1,23 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.11
+    - name: Build kogito-runtimes
+      run: git clone https://github.com/kiegroup/kogito-runtimes.git /tmp/kogito-runtimes && mvn clean install -DskipTests -f /tmp/kogito-runtimes/pom.xml
```
danielezonca, line 19: It should be nice to build the same branch of the PR in `kogito-runtime` too

.github/workflows/trustyCI.yml
```diff
@@ -0,0 +1,23 @@
+name: Trusty experimental CI
+
+on:
+  pull_request:
+    branches: [ master ]
+    paths:
+    - 'trusty/**'
+    - 'explainability-service/**'
```
danielezonca, line 8: It should be nice to remove these "hardcoded" paths and use something to extract all `experimental` modules but I don't know if it is possible

danielezonca (commented):
Just some minor comments

danielezonca:
I have added some comments to the code but in general we should decide how to proceed.
Trusty services will become probably soon no more experimental so this could be considered a reasonable way to have PR builds in the meanwhile but if we want to preserve the `experimental` profile for the future we should properly setup CI both for master and PRs.
I know we are having discussions around how to structure our jobs so we should decide how we want to proceed. 
@cristianonicolai 
Wdyt? Do you like the github action idea for this?

.github/workflows/trustyCI.yml
r00ta: I don't think this is possible atm with just this yml file. 
What we might do is to create another github action that whenever a PR is merged, it checks what are the experimental modules and eventually update/commit the update of this `trustyCI.yml` pipeline :D

.github/workflows/trustyCI.yml
r00ta: Uhm sorry @danielezonca , what do you mean?

.github/workflows/trustyCI.yml
r00ta: Done

cristianonicolai:
@danielezonca there is a push to try to keep as much as possible our tooling centered around jenkins file, lets have a quick sync today to review how can we incorporate trusty as part of the regular Jenkins build, I think that would be a prefered solution.

r00ta:
As we agreeded, I will close this PR and remove the experimental flag from the trusty modules so that they will be built by the standard CI

PR #236, opened 2020-05-26
KOGITO-2163 Make custom components OUIA:Component compliant

jstastny-cz:
Goals:
* To provide mechanism for specifying `data-ouia` prefixed attributes to comply with [1].
* To provide initial coverage of management-console and common packages.
* Each component to have top-level html attribute `component-type` set.
* Each component to have contextually unique `component-id` (within enclosing component) when there are more occurrences of the same type.
* Each component to have attribute `safe`

Drawbacks:
* Using Higher-order components (HOC) results in polluted snapshots -> typically usage of shallow is not providing satisfying results.
  * Mitigation:
    * using `mount` instead of `shallow`
    * provided `getWrapper` and `getWrapperAsync` functions in `@kogito-apps/common` that take care of creating a wrapper and navigate to the actual component (stripping off the levels added by HOCs)
      * but still you'll notice some fragments being OUIA-specific.
    * using Mocks wherever possible - to limit depth of the snapshot tree
* Only a subset of patternfly components adhere to the rules - should improve in future releases, room for contribution.

Resulting state:
* Tools for OUIA:Component compliance provided.
* Not all components adhere to the rules
  * especially those using `React.Fragment` as top-level component.
  * code duplication of certain parts prevent higher-level components to comply - e.g. breadcrumbs are not implemented as a single component, but the code is duplicated several times

[1] https://ouia.readthedocs.io/en/latest/README.html


packages/management-console/src/components/Templates/DomainExplorerDashboard/DomainExplorerDashboard.tsx
```diff
@@ -296,7 +300,7 @@ const DomainExplorerDashboard: React.FC<IOwnProps & RouteComponentProps<MatchPro
           ) : (
             <Card>
               <Bullseye>
-                <SpinnerComponent spinnerText="Loading domain data..." />
+                <SpinnerComponent spinnerText="Loading domain data..." ouiaId="loading-domain-data" />
```
jstastny-cz: I should fix this to use `attributeOuiaId` function as elsewhere

packages/management-console/src/components/Templates/DataListContainer/DataListContainer.tsx
```diff
@@ -338,6 +342,7 @@ const DataListContainer: React.FC<InjectedOuiaProps> = ({
                     setIsAllChecked={setIsAllChecked}
                     selectedNumber={selectedNumber}
                     setSelectedNumber={setSelectedNumber}
+                    ouiaId="toolbar-process-instances"
```
jstastny-cz: I should use `attributeOuiaId` function instead

packages/management-console/src/components/Organisms/DomainExplorerTable/DomainExplorerTable.tsx
```diff
@@ -305,6 +317,7 @@ const DomainExplorerTable = ({
           aria-label="Domain Explorer Table"
           className="kogito-management-console--domain-explorer__table"
           onCollapse={onCollapse}
+          ouiaId="domain-explorer-table"
```
jstastny-cz: again, should use `attributeOuiaId`

cristianonicolai:
@jstastny-cz so it seems that OUIA is something we would need to include in all components right? could you maybe give a short description to the team about how is that important so it can be took on in the next components that are created?  
Also, it needs rebase now.

cristianonicolai:
tagging all developers so they are aware.

jstastny-cz:
@cristianonicolai yes I thought I would do a brief introduction when the initial 'coverage' is done. As right now there are many moving parts, it would be probably, like you propose, better to share brief info and reasoning behind this already.
And thanks for tagging the rest of the team.

packages/management-console/src/components/Atoms/BrandComponent/tests/__snapshots__/BrandComponent.test.tsx.snap
```diff
@@ -1,7 +1,61 @@
...
+      "path": "/",
+      "url": "/",
+    }
+  }
+  ouiaContext={
+    Object {
+      "isOuia": false,
+      "ouiaId": null,
+    }
+  }
```
jomarko, line 46: @jstastny-cz could you please explain where these `match`, `location` and `history` came from? Going trough the PR and changed React components constructors I expected just `ouia...` will appear in snapshots.

jomarko (commented):
@jstastny-cz I went trough 1/4 of the PR as changes seems to me quite similar. However it is not clear to me:
- why the snapshots were 'polluted' also with some parts, which don't seems to be ouia related, see the inline comment below.
- when I read the ouia standard, my feeling was that to create compliant React component I will need to implement also some mechanism for unique id check, or `safe` readiness. But didn't spot such thing. Is that done by the imported functions from react-core or I just overlooked it?

packages/management-console/src/components/Atoms/BrandComponent/tests/__snapshots__/BrandComponent.test.tsx.snap
jstastny-cz: That's being passed by the Router surrounding component. Those props are needed for the component to be able to navigate to others (at least that's my understanding).
And why it appears only now? I've added the Router, cause otherwise the component can't be mounted using `mount`, `shallow` is much more simplistic thus it didn't complain about missing Router wrapper.

jomarko: @jstastny-cz ok, thank you, I now see you added `Router` component.

May I ask you one more time how it is with the compliance and unique ids and safe readiness?

jstastny-cz: In ideal world:
* Each our component should have (will omit `data-ouia-` prefix for brevity) `component-type` attribute that sort of groups the instances of the component **globally** and all the instances can be approached the same way (performing same set of actions, querying for same characteristics, etc).
* If there are multiple occurences of such component **locally** (within a single container - this can be page, div, but also table-row). Then these need to be distinguished using the `component-id`. So the id is not expected to be unique page-wide: let's suppose following scheme:
```
<rows>
  <row ouia-type='PersonListItem' ouia-id='person-123456'>
    <cell ouia-id='person-id'>123456</cell>
    <cell ouia-id='person-name'>Foo</cell>
    <cell ouia-id='person-age'>12</cell>
  </row>
  <row ouia-type='PersonListItem' ouia-id='person-654321'>
    <cell ouia-id='person-id'>654321</cell>
    <cell ouia-id='person-name'>Bar</cell>
    <cell ouia-id='person-age'>13</cell>
  </row>
</rows>
```
This should be compliant according to the specification.

Now the `data-ouia-safe` attribute serves for dynamic components, let's suppose a table has `refresh` button that starts async action and meanwhile shows spinning wheel and only when has response from server, then renders the results. So in such case upon the `refresh` action call the component need to state itself as `data-ouia-safe=false` until the results are processed and shown - then it can move back to `data-ouia-safe=true`. For static components the safe attribute will be true all the time. And one can then use this safe state check in selectors - to start interaction only when component is safe and to wait for an action to complete before continuing in tests.

jomarko: @jstastny-cz yes, that is also my understanding after reading doc, what is unclear to me, that I didn't see a place in this PR where we would have some id increaser to achieve unique ids, or some if then else that would decide if component is `safe`.


Sorry that I can not ask clearly. My question is that this PR is named make componenet compliant. Does it mean just some parts of standards are enough to implement to say component is ouia compliant ?

jstastny-cz: The ids are not supposed to be generated - see the sample cells above ... the id's are more like categories in that context. And the row elements, I tried to illustrate that it can be also context-aware, but again the numbers are not generated, but it's part of the data the row represents.

And for the latter - if you see any concerns about our compliance/coverage I'll be happy to fix those. Just for what it's worth, not all our components can really be 100% compliant I am afraid, as for example react.fragments are not represented in resulting html by any element, thus for such components that have react.fragment as top-level node, the solution is for all the contained components to be compliant. Here the problem appears - if the component within the react.fragment is a third-party component then we can't fulfill the rule.

```
<> <!-- this is react.fragment -->
  <OurComponent />
  <ThirdPartyComponent />
</>
```
So let's suppose `OurComponent` is OUIA compliant, but ThirdPartyComponent is not ... we would either need to wrap the thirdparty component to our component that could be compliant, or for the initial effort we can relax our rules.

So to answer your concern - yes, the Jira needs probably renaming, cause we've hit the real world and the compliance itself would take more effort and time.

jomarko (approved):
@jstastny-cz thank you for conversation and explanation.

cristianonicolai:
@jstastny-cz needs rebase

jstastny-cz:
@cristianonicolai waiting for the fixes in snapshots to land in master so that I don't spend hours fixing tests that are actually faulty.

jstastny-cz:
Closing this big PR and will create component-related PRs so that it's an incremental process.

PR #237, opened 2020-05-27
KOGITO-2299: Refactored Domain Explorer components

Sara4994:
Hello! please review, this PR contains changes on react components that are related to Domain Explorer. Components are refactored and moved into common package to be reusable on management and task consoles.

JIRA issue: https://issues.redhat.com/browse/KOGITO-2299


packages/common/src/components/Atoms/LoadMore/LoadMore.css
```diff
@@ -0,0 +1,4 @@
+.kogito-management-console__load-more {
```
cristianonicolai, line 1: I guess we would need a more neutral naming as it would be resued across different services. Perhaps just drop _management-console_. @srambach wdyt?

packages/management-console/src/components/Templates/DomainExplorerLandingPage/DomainExplorerLandingPage.tsx
```diff
@@ -55,36 +28,7 @@ const DomainExplorerLandingPage: React.FC<InjectedOuiaProps> = ({
...
-                </>
-              ) : (
-                <TextContent>
-                  <Text component={TextVariants.h2}>No domains available</Text>
-                </TextContent>
-              )}
-            </EmptyState>
-          </CardBody>
-        </Card>
+        <DomainExplorer />
```
cristianonicolai, line 31: I guess the DomainExplorer component should be used inside the CardBody so we avoid it knowing about the page layout?

packages/management-console/src/components/Templates/DomainExplorerPage/DomainExplorerPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+                  <Link to={'/DomainExplorer'}>
+                    {item.replace(/([A-Z])/g, ' $1').trim()}
+                  </Link>
+                </BreadcrumbItem>
+              );
+            }
+          })}
+        </Breadcrumb>
+      </PageSection>
+      <DomainExplorerContainer
```
cristianonicolai, line 107: I guess same here, DomainExplorer would be inside of a CardBody?

packages/common/src/components/Atoms/KogitoSpinner/KogitoSpinner.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
cristianonicolai, line 1: at least in the log here it show some files as deleted and new ones created, would be better to keep the git history and then rename the files. For example from Spinner to KogitoSpinner.

packages/common/src/components/Atoms/LoadMore/LoadMore.css
Sara4994: @cristianonicolai @srambach renaming to "kogito-common__load-more". wdyt?

srambach: I agree we need something different. I could go either way on including `-common` or not.

Sara4994:
Closing this PR and opening minor PRs for the same, just to break larger changes

PR #239, opened 2020-05-28
KOGITO-2172 Implemented display data functionality for the Generic Task List Comp…

yzhao583:
[https://issues.redhat.com/browse/KOGITO-2172](https://issues.redhat.com/browse/KOGITO-2172)

Implemented display data functionality using PF4 Table component for Generic Task List Component. Here are the main functionalities for the DataTableComponent:

- Display data (an array of json object) passed in props.
- Users can decide the columns they want to show in the table by defining the "columns" (an array of string) prop; If there is no "columns" props defined, all properties of the "data" json object will be displayed.
-  Users can define custom "Loading" component and pass it to the table component in props, if there is a custom "Loading" component defined, it will be displayed when the table is loading status; Otherwise, the default "Loading" component is displayed.
- Users can define custom "Error" component and pass it to the table component in props, if there is a custom "Error" component defined, it will be displayed when the table is loading status; Otherwise, the default "Loading" component is displayed.

### DIFF
--- a/packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
+++ b/packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
@@ -1,13 +1,3 @@
-.spinner-center {
-  text-align: center;
-  margin-top: 5%;
-}
-
-.error-text{
-  text-align: center;
-  margin-top: 5%;
-}
-
 .data-table{
   overflow: auto;
 }
--- a/packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
+++ b/packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
@@ -1,0 +1,13 @@
+.spinner-center {
+  text-align: center;
+  margin-top: 5%;
+}
+
+.error-text{
+  text-align: center;
+  margin-top: 5%;
+}
+
+.data-table{
+  overflow: auto;
+}
--- a/packages/task-console/src/components/Templates/DataTableContainer/DataTableContainer.tsx
+++ b/packages/task-console/src/components/Templates/DataTableContainer/DataTableContainer.tsx
@@ -16,12 +16,11 @@ import PageTitleComponent from '../../Molecules/PageTitleComponent/PageTitleComp
 import './DataTable.css';
 import DataTable from '../../Organisms/DataTable/DataTable';
 import { useGetUserTasksByStatesQuery } from '../../../graphql/types';
-import { ouiaPageTypeAndObjectId } from '@kogito-apps/common';
-import SpinnerComponent from '../../Atoms/SpinnerComponent/SpinnerComponent';
+import { ouiaPageTypeAndObjectId, KogitoSpinner } from '@kogito-apps/common';
 
 const UserTaskLoadingComponent = (
   <Bullseye>
-    <SpinnerComponent spinnerText="Loading user tasks..." />
+    <KogitoSpinner spinnerText="Loading user tasks..." />
   </Bullseye>
 );
 
--- a/packages/task-console/src/components/Templates/DataTableContainer/DataTableContainer.tsx
+++ b/packages/task-console/src/components/Templates/DataTableContainer/DataTableContainer.tsx
@@ -1,0 +1,87 @@
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  Card,
+  Grid,
+  GridItem,
+  PageSection,
+  InjectedOuiaProps,
+  withOuiaContext,
+  Bullseye
+} from '@patternfly/react-core';
+import _ from 'lodash';
+import React, { useEffect } from 'react';
+import { Link } from 'react-router-dom';
+import PageTitleComponent from '../../Molecules/PageTitleComponent/PageTitleComponent';
+import './DataTable.css';
+import DataTable from '../../Organisms/DataTable/DataTable';
+import { useGetUserTasksByStatesQuery } from '../../../graphql/types';
+import { ouiaPageTypeAndObjectId } from '@kogito-apps/common';
+import SpinnerComponent from '../../Atoms/SpinnerComponent/SpinnerComponent';
+
+const UserTaskLoadingComponent = (
+  <Bullseye>
+    <SpinnerComponent spinnerText="Loading user tasks..." />
+  </Bullseye>
+);
+
+const DataTableContainer: React.FC<InjectedOuiaProps> = ({ ouiaContext }) => {
+  const {
+    loading,
+    error,
+    data,
+    refetch,
+    networkStatus
+  } = useGetUserTasksByStatesQuery({
+    variables: {
+      state: ['Ready'] // FIXME: state should not be hard-coded.
+    },
+    fetchPolicy: 'network-only',
+    notifyOnNetworkStatusChange: true
+  });
+
+  const columns = [
+    'ProcessId',
+    'Name',
+    'Priority',
+    'ProcessInstanceId',
+    'State'
+  ];
+
+  useEffect(() => {
+    return ouiaPageTypeAndObjectId(ouiaContext, 'user-tasks');
+  });
+
+  return (
+    <React.Fragment>
+      <PageSection variant="light">
+        <PageTitleComponent title="User Tasks" />
+        <Breadcrumb>
+          <BreadcrumbItem>
+            <Link to={'/'}>Home</Link>
+          </BreadcrumbItem>
+          <BreadcrumbItem isActive>User Tasks</BreadcrumbItem>
+        </Breadcrumb>
+      </PageSection>
+      <PageSection>
+        <Grid gutter="md">
+          <GridItem span={12}>
+            <Card className="data-table">
+              <DataTable
+                data={data ? data.UserTaskInstances : undefined}
+                isLoading={loading}
+                columns={columns}
+                networkStatus={networkStatus}
+                error={error}
+                refetch={refetch}
+                LoadingComponent={UserTaskLoadingComponent}
+              />
+            </Card>
+          </GridItem>
+        </Grid>
+      </PageSection>
+    </React.Fragment>
+  );
+};
+
+export default withOuiaContext(DataTableContainer);
--- a/packages/task-console/src/components/Templates/DataTableContainer/__mocks__/DataTableContainer.tsx
+++ b/packages/task-console/src/components/Templates/DataTableContainer/__mocks__/DataTableContainer.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+const MockDataTableContainer = (): React.ReactElement => {
+  return <></>;
+};
+
+export default MockDataTableContainer;
--- a/packages/task-console/src/components/Templates/DataTableContainer/tests/DataTableContainer.test.tsx
+++ b/packages/task-console/src/components/Templates/DataTableContainer/tests/DataTableContainer.test.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import DataTableContainer from '../DataTableContainer';
+import { getWrapper } from '@kogito-apps/common';
+import { MemoryRouter as Router } from 'react-router-dom';
+import { MockedProvider } from '@apollo/react-testing';
+
+jest.mock('../../../Organisms/DataTable/DataTable');
+
+describe('DataTableContainer component tests', () => {
+  it('Snapshot tests', () => {
+    const wrapper = getWrapper(
+      <Router>
+        <MockedProvider>
+          <DataTableContainer />
+        </MockedProvider>
+      </Router>,
+      'DataTableContainer'
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+});
--- a/packages/task-console/src/components/Templates/DataTableContainer/tests/__snapshots__/DataTableContainer.test.tsx.snap
+++ b/packages/task-console/src/components/Templates/DataTableContainer/tests/__snapshots__/DataTableContainer.test.tsx.snap
@@ -1,0 +1,194 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`DataTableContainer component tests Snapshot tests 1`] = `
+<DataTableContainer
+  ouiaContext={
+    Object {
+      "isOuia": false,
+      "ouiaId": null,
+    }
+  }
+>
+  <PageSection
+    variant="light"
+  >
+    <section
+      className="pf-c-page__main-section pf-m-light"
+    >
+      <PageTitleComponent
+        title="User Tasks"
+      >
+        <Title
+          headingLevel="h1"
+          size="4xl"
+        >
+          <h1
+            className="pf-c-title pf-m-4xl"
+          >
+            User Tasks
+          </h1>
+        </Title>
+      </PageTitleComponent>
+      <Component>
+        <ComponentWithOuia
+          component={[Function]}
+          componentProps={
+            Object {
+              "children": Array [
+                <BreadcrumbItem>
+                  <Link
+                    to="/"
+                  >
+                    Home
+                  </Link>
+                </BreadcrumbItem>,
+                <BreadcrumbItem
+                  isActive={true}
+                >
+                  User Tasks
+                </BreadcrumbItem>,
+              ],
+            }
+          }
+          consumerContext={
+            Object {
+              "isOuia": false,
+              "ouiaId": null,
+            }
+          }
+        >
+          <Breadcrumb
+            ouiaContext={
+              Object {
+                "isOuia": false,
+                "ouiaId": null,
+              }
+            }
+          >
+            <nav
+              aria-label="Breadcrumb"
+              className="pf-c-breadcrumb"
+            >
+              <ol
+                className="pf-c-breadcrumb__list"
+              >
+                <BreadcrumbItem>
+                  <li
+                    className="pf-c-breadcrumb__item"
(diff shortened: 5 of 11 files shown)


packages/task-console/src/components/Templates/DataTableContainer/__mocks__/DataTableContainer.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const DataTableContainer = (): React.ReactElement => {
```
jstastny-cz: I would argue to keep the fact that it is a mock visible, what I am doing in other PR is to name the component as mock, in this case it would be MockedDataTableContainer wdyt? It's then apparent in snapshot that we actually don't expect the component to be rendered properly.

packages/task-console/src/components/Templates/DataTableContainer/tests/DataTableContainer.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import DataTableContainer from '../DataTableContainer';
+import { getWrapper } from '@kogito-apps/common';
+import { MemoryRouter as Router } from 'react-router-dom';
+import { MockedProvider } from '@apollo/react-testing';
+
+describe('DataTableContainer component tests', () => {
+  it('Snapshot tests', () => {
+    const wrapper = getWrapper(
+      <Router>
+        <MockedProvider>
```
jstastny-cz, line 13: For any mocks using MockedProvider we need to do additional steps to have nice snapshots (otherwise we get just empty-state component render even if we provide mocked data).
Here there are no values passed to the component, so I guess this test is for rendering the DataTableContainer with a DataTableComponent than is in 'empty-state'. So I see two options now:
* If we don't care about the DataTableComponent in the tests here, we should mock it IMO.
* If we care about DataTableComponent rendering we should also add test where the component would actually show some data.

But I assume option 1 is sufficient for testing this container, and given that you've already created mock for DataTableComponent, I would just mock it for this test.

packages/task-console/src/components/Templates/DataTableContainer/tests/__snapshots__/DataTableContainer.test.tsx.snap
```diff
@@ -0,0 +1,237 @@
...
+            <div
+              className="pf-l-grid__item pf-m-12-col"
+            >
+              <Card
+                className="data-table"
+              >
+                <article
+                  className="pf-c-card data-table"
+                >
+                  <DataTableComponent
```
jstastny-cz: In relation with comment on the test itself -> here would be the place in snapshot to see the mocked component.

packages/task-console/src/components/Organisms/DataTableComponent/tests/DataTableComponent.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+  networkStatus: 1,
+  error: {},
+  refetch: jest.fn(),
+  LoadingComponent: undefined,
+  ErrorComponent: undefined
+};
+
+describe('DataTable component tests', () => {
+  it('Snapshot tests 1', () => {
+    const wrapper = shallow(
```
jstastny-cz: IMO shallow does not actually run the graphql query like `mount` would. That's why the following snapshot is also just showing only the component with properties being passed, no children - which I would expect based on the mocked data passed in.

packages/task-console/src/components/Organisms/DataTableComponent/__mocks__/DataTableComponent.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const DataListComponent = (): React.ReactElement => {
```
jstastny-cz: I would call the component MockedDataList -> for the snapshots to apparently show that the component is actually mocked.

packages/task-console/src/components/Organisms/DataTableComponent/tests/DataTableComponent.test.tsx
jstastny-cz: If you decide to use mount instead mind that the test needs to be async, I have a solution in my PR, where I just introduce another wrapper for such cases: `getWrapperAsync`, then the test would look like:
```
it('', async ()=> {
     const wrapper = await getWrapperAsync(....)
     expect(wrapper).toMatchSnapshot();
})
```
Reach out to me if needed.

jstastny-cz (changes requested):

packages/task-console/src/components/Organisms/DataTableComponent/__mocks__/DataTableComponent.tsx
yzhao583: Thanks! Will update.

packages/task-console/src/components/Templates/DataTableContainer/tests/DataTableContainer.test.tsx
yzhao583: @jstastny-cz Updated, Thanks.

jstastny-cz (approved):
LGTM, thanks.

packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
```diff
@@ -0,0 +1,13 @@
...
+}
+
+.error-text{
+  text-align: center;
+  margin-top: 5%;
+}
+
+.data-table{
+  overflow: auto;
+}
```
srambach: Two comments here. First, I think this is the fourth time we have .spinner-center and .error-text defined between the packages. We definitely need to think about how to share common CSS here.

Second, we need to establish some good naming practices for CSS classes. I have been using something close to `.kogito-packagename—block__element—modifier`. In the domain explorer, we’ve used this to solve the same problem of a too-wide table.
```
.kogito-management-console--domain-explorer__table-OverFlow {
overflow-y: auto;
}
```

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
```diff
@@ -0,0 +1,140 @@
...
+          []
+        ),
+        rowKey: JSON.stringify(rowData).substr(0, 20) // This is a walkaround to bypass the "id" cannot be included in "columns" issue
+      };
+    });
+  }
+  return rowList;
+};
+
+const DataTableComponent: React.FC<IOwnProps> = ({
```
AjayJagan: Hello @yzhao583 , So there was a discussion on naming the components and  everyone feels that it would be better to remove the 'component' at the end of every components(for instance in this case just the DataTable). wdyt @cristianonicolai @Sara4994  @srambach . Other than that the PR looks awesome :)

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
```diff
@@ -0,0 +1,140 @@
+import React, { useEffect, useState } from 'react';
+import { Bullseye } from '@patternfly/react-core';
+import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import SpinnerComponent from '../../Atoms/SpinnerComponent/SpinnerComponent';
```
cristianonicolai: @yzhao583, @Sara4994 is working on moving and renaming these components to the common module, maybe is worth if you guys sync for one PR after the other.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
```diff
@@ -0,0 +1,140 @@
...
+                } else {
+                  result.push(value);
+                }
+              }
+            });
+            return result;
+          },
+          []
+        ),
+        rowKey: JSON.stringify(rowData).substr(0, 20) // This is a walkaround to bypass the "id" cannot be included in "columns" issue
```
cristianonicolai: @yzhao583 I think on management console we end-up using a random uuid, maybe double check with @AjayJagan @Sara4994.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
cristianonicolai: +1 for dropping _Compoment_ from the naming. Also, @yzhao583 is this a Task specific table, do we want to maybe name for specifically or even keep as generic and move to the common module?

packages/task-console/src/components/Templates/DataTableContainer/DataTableContainer.tsx
```diff
@@ -0,0 +1,87 @@
...
+    networkStatus
+  } = useGetUserTasksByStatesQuery({
+    variables: {
+      state: ['Ready'] // FIXME: state should not be hard-coded.
+    },
+    fetchPolicy: 'network-only',
+    notifyOnNetworkStatusChange: true
+  });
+
+  // const columns = [
```
pefernan: I think this could be removed

pefernan (approved):
Hey  @yzhao583 Your code works and looks great. I'd remove some unneeded comments but since this is an ongoing dev we can merge and remove later.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
Sara4994: @yzhao583 you can use generate random key something like this "Math.random().toString()"

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
Sara4994: @yzhao583 Now you can import "KogitoSpinner" component from @kogito-apps/common

packages/task-console/src/components/Organisms/DataTableComponent/tests/DataTableComponent.test.tsx
yzhao583: @jstastny-cz Updated, thanks

packages/task-console/src/components/Organisms/DataTableComponent/__mocks__/DataTableComponent.tsx
yzhao583: @jstastny-cz Updated, thanks

packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
yzhao583: @srambach Thanks for the feedback. Really appreciate that! 

> Two comments here. First, I think this is the fourth time we have .spinner-center and .error-text defined between the packages. We definitely need to think about how to share common CSS here.
Yes, I totally agree we should define some common CSS class here so it can be shared. Could you please suggest how and where should we do that? Thanks.





> Second, we need to establish some good naming practices for CSS classes. I have been using something close to `.kogito-packagename—block__element—modifier`. In the domain explorer, we’ve used this to solve the same problem of a too-wide table.

This component it not only for Task Console, and maybe it is not only for KOGITO as well. So I assume using some general name may make more sense here? wdyt

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
yzhao583: @AjayJagan Thanks for the feedback, really appreciate that! 


> So there was a discussion on naming the components and everyone feels that it would be better to remove the 'component' at the end of every components(for instance in this case just the DataTable)

Yes, I feel like removing 'component' in the name make more sense.

@cristianonicolai Thanks for the feedback, really appreciate that! 


> is this a Task specific table, do we want to maybe name for specifically or even keep as generic and move to the common module?

This is not only for Task Console, the goal here is to make a general Table Component. The reason I did not put it into the Common module is that I found there is also a similar Table Component in Domain Explorer module, and it has more functionalities, I am not sure which one should we put into the Common module. Any suggestion? Thanks

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
yzhao583: @cristianonicolai @Sara4994 Thanks! I will update to use the component from the Common module.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
yzhao583: @cristianonicolai @Sara4994 I personally prefer the "uuid" solution, WDYT? And if we use "uuid" here, it may generate some headache when we do snapshot test, any suggestions? Appreciated!

packages/task-console/src/components/Templates/DataTableContainer/DataTableContainer.tsx
yzhao583: @pefernan Sorry, I just commented it out by mistake, we need that here, but thanks for the feedbacks! Appreciated!

nmirasch (approved):
Looks great, Thanks!

AjayJagan (approved):
:tada:

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
cristianonicolai: task id and process id, are actually uuid, to me that should be enough but again @Sara4994 @AjayJagan might know better the inner works.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
cristianonicolai: @yzhao583, @Sara4994 is working on moving that component into the common module, perhaps we should wait to get that in first?

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
Sara4994: @cristianonicolai for keys i would prefer Math.random(), but if we look for some other alternative there is something called nanoid() which is shorter in length than uuid

packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
srambach: @yzhao583  
Since the overflow is on the data table in task console, then how about `.kogito-task-console--data-table--Overflow`? If at some point the data tables are combined into the common package, then they can share a class in common.
Do we need to set `overflow` here, or is `overflow-y` sufficient?

As for `.error-text` and `.spinner-center` - I don't even see them being used anywhere even though they are defined several times. Can they be removed?

yzhao583: @srambach Yes, you are right, the Data Table is in task console, but it maybe reused for by applications. And the reason to add overflow here is handle the case that there is too many columns.

I will remove the unused classes, thanks.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
pefernan: @yzhao583 @cristianonicolai @Sara4994 I have one question, is both components overlapping? The goal of this PR is creating a _Table Component_, to be used in Task Console but ideally reusable for other purposes (maybe jobs or whatever we may need). If both of them have the same goal, should we modify the existing DE table @Sara4994's working on, to make it generic so we can adopt it in both modules? If not, to me +1 to merge is (maybe renaming it to avoid collisions) so we can continue working. WDYT?

pefernan: @yzhao583 @Sara4994 do you think it would be feasible to do it, that could save us a lot of time :-P

cristianonicolai: @pefernan I know its not ideal but we're focusing on moving domain explorer to common right now, it might be a few prs away see https://github.com/kiegroup/kogito-apps/pull/245 and https://github.com/kiegroup/kogito-apps/pull/242,  I think would be best if we concentrate the effort in the long term approach.

pefernan: @cristianonicolai so we should rename this so we can merge?

packages/task-console/src/components/Templates/DataTableContainer/DataTable.css
srambach: @yzhao583 Great!  What other applications would this class be used in? Do you mean something outside of Kogito?

yzhao583: @srambach Yes, it could be. Although the main goal is build a Table Component for Task Console, there is also another goal that to make is reusable for other applications as well.

srambach: OK, I see. I still feel that it would be best to namespace any classes we are adding. That way it will avoid potential conflicts and identify the source of the class for debugging.

packages/task-console/src/components/Organisms/DataTableComponent/DataTableComponent.tsx
pefernan: @cristianonicolai @yzhao583 @nmirasch why not merge this into the task-console, so @yzhao583 can continue working, and we may move it to commons when needed. In future developmens we'll have to be more synchronized.

PR #241, opened 2020-06-01
KOGITO-536: Data-index-service Keycloak integration: GraphiQL (UI)

nmirasch:
jira related .. https://issues.redhat.com/browse/KOGITO-536
- Adds oidc keycloak configuration to allow data-index work as an service (working with bearer token) and as a webapp to serve the graphiql  redirecting to keycloak login page for authentication.


data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -66,27 +70,38 @@ public void testNoTokenProvided() {
...
-
         given().auth().oauth2(getAccessToken("jdoe"))
                 .contentType(ContentType.JSON)
                 .body("{ \"query\" : \"{ProcessInstances{ id } }\" }")
                 .when().post("/graphql")
                 .then().log().ifValidationFails().statusCode(200);
     }
 
+    @Test
+    public void graphiqlLoginOnKeycloackServedTest(){
```
jstastny-cz: Keycloack -> Keycloak

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -32,18 +33,25 @@
...
-    private static final String KEYCLOAK_SERVER_URL = System.getProperty("keycloak.url", "http://localhost:8281/auth");
-    private static final String KEYCLOAK_REALM = "kogito";
-    private static final String KEYCLOAK_CLIENT_ID = "kogito-data-index-service";
+    private static final String KEYCLOAK_SERVER_URL = System.getProperty("quarkus.oidc.auth-server-url", "http://localhost:8281/auth/realms/kogito");
+    private static final String KEYCLOAK_CLIENT_ID = System.getProperty("quarkus.oidc.client-id", "kogito-data-index-service");
+    private static final String KEYCLOAK_CLIENT_SECRET = System.getProperty("quarkus.oidc.credentials.secret", "secret");
+
+    @BeforeAll
+    public static void setup() {
+        System.setProperty("quarkus.http.auth.policy.role-policy1.roles-allowed", "confidential");
```
jstastny-cz: why moving this from application.properties to here?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -66,27 +70,38 @@ public void testNoTokenProvided() {
...
                 .when().post("/graphql")
                 .then().log().ifValidationFails().statusCode(200);
     }
 
+    @Test
+    public void graphiqlLoginOnKeycloackServedTest(){
+        // Returning keycloak login page
+        given().when().get("/graphiql/")
+                .then()
+                .statusCode(200);
```
jstastny-cz, line 73: Is the status code under our control? Some sites use 302 code to signal that the page is there, but something still needs to be done to access it.

data-index/data-index-service/src/main/resources/application.properties
```diff
@@ -74,6 +74,13 @@ quarkus.oidc.auth-server-url=none
...
-%keycloak.quarkus.http.auth.permission.roles1.policy=role-policy1
+%keycloak.quarkus.oidc.application-type=service
+
+%keycloak.quarkus.oidc.web-app-tenant.auth-server-url=http://localhost:8280/auth/realms/kogito
+%keycloak.quarkus.oidc.web-app-tenant.client-id=kogito-data-index-service
+%keycloak.quarkus.oidc.web-app-tenant.credentials.secret=secret
+%keycloak.quarkus.oidc.web-app-tenant.application-type=web-app
+
+# HTTP Security Configuration
+%keycloak.quarkus.http.auth.permission.authenticated.paths=/*
```
jstastny-cz, line 86: wondering why these two are not in the test properties?

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.quarkus.oidc.TenantResolver;
+import io.vertx.ext.web.RoutingContext;
+
+@ApplicationScoped
+public class MultiTenantResolver implements TenantResolver {
+
+    @Override
+    public String resolve(RoutingContext context) {
+        if (context.request().path().startsWith("/graphiql")) {
```
jstastny-cz: so this decides which tenant is accessing the app based on the context, right. graphiql is a UI, thus web-app-tenant should be used, otherwise it will resort to the default config, is that correct?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
nmirasch: ups! thanks!

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
nmirasch: @jstastny-cz yeah, I agree,  I have removed this. The purpose of this properties setting were regarding isolate  testing at my local, no need to be there.

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
nmirasch: @jstastny-cz Yeah, that value is the one returned by the quarkus oidc extension when it's configured as web-app. It returns request code for 'OK' and redirect to keycloak login.

data-index/data-index-service/src/main/resources/application.properties
nmirasch: @jstastny-cz yeah! same comment as before here. That authentication configuration is passed as system prop at the pom.xml test declaration.

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
nmirasch: @jstastny-cz Correct! that's the idea, define a different oidc configuration for the ui (graphiql) and the graphql service. Regarding the keycloak authentication workflows are different in both cases.

nmirasch:
@jstastny-cz updated, could you take a look, please? Thanks!!

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
```diff
@@ -0,0 +1,34 @@
...
+
+package org.kie.kogito.index.auth;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import io.quarkus.oidc.TenantResolver;
+import io.vertx.ext.web.RoutingContext;
+
+@ApplicationScoped
+public class MultiTenantResolver implements TenantResolver {
```
Sgitario, line 27: At the moment, I see these hard-coded values: 
- "graphiql": before this PR, this value was only used in VertxRouterSetup, but now it's used also in MultiTenantResolved and KeycloakIntegrationIndexingServiceIT.
- "web-app-tenant": it's used here and in the properties to configure the oidc properties. 

What about to add a property to define the "graphiql" property:

´´´
dataindex.vertx-graphql.ui.path=/graphiql
´´´

| Note that the name property is in sync with the default vertx property called "quarkus.vertx-graphql.ui.path" (more in [here](https://quarkus.io/guides/all-config#quarkus-vertx-graphql_quarkus-vertx-graphql))

And also define the tenants for each path in the properties:

```
dataindex.paths."/graphiql".tenant=web-app-tenant
```

This way we can configure the tenants and configuration entirely driven by properties:

```
%keycloak.quarkus.oidc.web-app-tenant.auth-server-url=http://localhost:8280/auth/realms/kogito
%keycloak.quarkus.oidc.web-app-tenant.client-id=kogito-data-index-service
%keycloak.quarkus.oidc.web-app-tenant.credentials.secret=secret
%keycloak.quarkus.oidc.web-app-tenant.application-type=web-app
```

And we can even define more custom tenants for different paths:

```
dataindex.paths."/custom".tenant=custom-app-tenant
%keycloak.quarkus.oidc.custom-app-tenant.auth-server-url=http://localhost:8280/auth/realms/kogitocustom
```

This is specially interesting to define different roles by paths.
What do you think?

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -43,14 +48,37 @@
...
+                return multiMap.add(HttpHeaders.AUTHORIZATION, "Bearer " + token);
+            });
         }
+        router.route("/graphiql/*").handler(graphiQLHandler);
+        router.route().handler(LoggerHandler.create());
+        router.route().handler(StaticHandler.create());
+        router.route().handler(FaviconHandler.create());
+        router.route("/").handler(ctx -> ctx.response().putHeader("location", "/graphiql/").setStatusCode(302).end());
         router.route("/graphql").handler(ApolloWSHandler.create(graphQL));
         router.route("/graphql").handler(GraphQLHandler.create(graphQL, new GraphQLHandlerOptions()));
```
Sgitario, line 68: I know this was not changed as part of this PR, but what is the difference between the graphiql and graphql? As far I can see in [here](https://quarkus.io/guides/all-config#quarkus-vertx-graphql_quarkus-vertx-graphql), the graphql is only available when running the data index in dev mode? 
Can we remove the "/graphql" path and use the property "quarkus.vertx-graphql.ui.path" instead? So if in the future, we change this property, we don't need to change this class.

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -43,14 +48,37 @@
...
     }
+
+    private String getCurrentAccessToken(RoutingContext routingContext) {
+        if ((routingContext.user() != null) &&
+                ((QuarkusHttpUser) routingContext.user()).getSecurityIdentity() != null &&
+                ((QuarkusHttpUser) routingContext.user()).getSecurityIdentity().getCredential(AccessTokenCredential.class) != null) {
+            return ((QuarkusHttpUser) routingContext.user()).getSecurityIdentity().getCredential(AccessTokenCredential.class).getToken();
+        }
+        return "";
+    }
```
Sgitario, line 82: Suggestion, what about using stream and optionals here? 

```
private String getCurrentAccessToken(RoutingContext routingContext) {
        return Optional.ofNullable(routingContext.user())
                .map(user -> ((QuarkusHttpUser) user).getSecurityIdentity())
                .map(identity -> identity.getCredential(AccessTokenCredential.class))
                .map(AccessTokenCredential::getToken)
                .orElse(StringUtils.EMPTY);
    }
```

I read this code better like this, what do you think?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -32,18 +32,18 @@
 @QuarkusTestResource(InfinispanServerTestResource.class)
```
Sgitario: Does it make sense to set the default image properties for infinispan and keycloak?

I'm using this one for Infinispan:

```
private static final String INFINISPAN_IMAGE = System.getProperty("container.image.infinispan", "jboss/infinispan-server:10.1.8.Final");
```

And this one for keycloak: 

```
private static final String KEYCLOAK_IMAGE = System.getProperty("container.image.keycloak", "jboss/keycloak:10.0.2");
```

So we can run the tests directly from the IDE with further configuration.

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -66,27 +62,38 @@ public void testNoTokenProvided() {
...
+        // Returning keycloak login page
+        given().when().get("/graphiql/")
+                .then()
+                .statusCode(200);
+        // Returning keycloak login page
+        given().auth().oauth2(getAccessToken("jdoe"))
+                .when().get("/graphiql/")
+                .then()
+                .statusCode(200);
+         given().when().get("/other/")
```
Sgitario: There is an extra space here at the beginning.

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -66,27 +62,38 @@ public void testNoTokenProvided() {
...
-
         given().auth().oauth2(getAccessToken("jdoe"))
                 .contentType(ContentType.JSON)
                 .body("{ \"query\" : \"{ProcessInstances{ id } }\" }")
                 .when().post("/graphql")
                 .then().log().ifValidationFails().statusCode(200);
     }
 
+    @Test
+    public void graphiqlLoginOnKeycloakServedTest(){
```
Sgitario: I added this new assertion:

```
given().auth().oauth2("any-wrong-token")
        .when().get("/graphiql/")
        .then().statusCode(401);
```

And it's failing because it retuns 200 OK code. Is this expected? Maybe it's redirecting to a Keycloak page when this is not valid?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -32,18 +32,18 @@
...
         //alice only have role User, resource is forbidden
-        given().auth().oauth2(getAccessToken("alice"))
-                .when().get("/graphiql/")
+        given().contentType(ContentType.JSON).body("{ \"query\" : \"{ProcessInstances{ id } }\" }")
+                .auth().oauth2(getAccessToken("alice"))
+                .when().get("/graphql")
                 .then()
-                .statusCode(404);
-
+                .statusCode(403);
```
Sgitario: I'm confused about the tests that cover the "/graphql" path. 
I ran the tests using "mvn clean verify" and I'm aware of these tests are working fine, but as far I can see in the VertxRouterSetup, we are not doing anything new about this path, right? 
Also, this path does not allow the GET method, but POST method... so no idea why is working. What am I missing?

Sgitario (changes requested):
Can you provide information about how to run the data index component with security enabled? Maybe in the README.md or https://github.com/kiegroup/kogito-runtimes/wiki/Data-Index-Service?

Also, I thought this PR will address the issue about how to enable security at runtime, but I tried:

```
java -jar -Dquarkus.oidc.enabled=true
```

And didn't make any difference.

Do you know if this will be handled as part of another ticket or if there is a strategy to provide this feature (for example, via different data index images)?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
jstastny-cz: IMO the important fact is here that the user is not authorized to access the web context, details like methods allowed would be probably checked later on.

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
jstastny-cz: See the comments below, that's the keycloak login page which returns 200.

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
jstastny-cz: Not sure what you mean here. Do you mean like in `data-index/data-index-service/src/test/java/org/kie/kogito/index/InfinispanServerTestResource.java` ? Cause it's already in place.

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
jstastny-cz: That's a different endpoint. graphiql is web app, graphql is the service endpoint.

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
Sgitario: I mean to add a default image in case of the property is not set.

jstastny-cz: I would argue against. Having configuration of the images both in pom and in code is not manageable easily, it could hide the fact that you've forgotten to set the property, the "somewhere defined default" could become outdated, etc. Anyway that should not be taken care of as part of this PR.

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
Sgitario: The web application is also exposed via "graphql-ui" endpoint (this can be configured via the property "quarkus.vertx-graphql.ui.path"). But this is only working when running the app in dev mode (because the property "quarkus.vertx-graphql.ui.always-include" defaults to "false"). 

Therefore, as long as we only have one web app at runtime, we're good. Thanks!

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
Sgitario: Ok with taking care of this as part of another ticket. 
My against of having properties in the POM file is that is not IDE compatible (we can't run suites or single tests directly from the IDE) and for me, this is really useful for debugging.

jstastny-cz: Well you can, you just need to set the property to your test runner. It's maybe just a question of taste, who likes what. Feel free to open a jira if you like.

Sgitario: The problem is that a different test execution is used if we run the whole test suite, or one test or another (at least in Eclipse, not sure about Intellij or VS Code). So, I would need to add the properties in every test execution. Maybe, there could have another approaches.

Having these properties in a test.properties (or similar) and injecting this file into Quarkus would work when running the tests from Maven or any IDE. 

However, this is out of scope here. I wrote it here as I saw we're using default values for some other properties.

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
nmirasch: Hi @Sgitario, I think it's a good idea of making the graphiql path and oidc-tenant  given as configuration properties.Let me give it a try Thanks!

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: Thanks!

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
nmirasch: @Sgitario I have to say that I have that default value for that props in my local environment to execute test from Intellij.. :-)
But maybe we can add there req in another jira just to see it it's convenient .

nmirasch:
@Sgitario I updated this code to include the kogito.dataindex.vertx-graphql.ui.path and kogito.dataindex.vertx-graphql.ui.tenant  properties to be able to configure that relationship by properties.
Could you take a look, please? Thanks!

Sgitario:
jenkins please retest this

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,44 @@
     @ConfigProperty(name = "quarkus.oidc.enabled", defaultValue = "false")
     Boolean authEnabled;
 
+    @Inject
+    @ConfigProperty(name = "kogito.dataindex.vertx-graphql.ui.path", defaultValue = "/graphiql")
```
Sgitario: I see there is an existing property for data index component called "kogito.data-index.storage.type=infinispan". Maybe we should continue this name convention and rename this property to "kogito.data-index.vertx-graphql.ui.path".
This also applies to "kogito.data-index.vertx-graphql.ui.tenant"

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Before
+    public void setup(){
+        lenient().when(routingContextMock.request()).thenReturn(requestMock);
+    }
+
+    @Test
+    void resolveGraphiqlTenantTest() {
+        lenient().when(requestMock.path()).thenReturn(GRAPH_UI_PATH);
+        assertThat(GRAPH_UI_TENANT.equals(multiTenantResolver.resolve(routingContextMock)));
```
Sgitario: This assertion should be:

```java
assertThat(multiTenantResolver.resolve(routingContextMock)).isEqualTo(GRAPH_UI_TENANT);
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @Mock
+    RoutingContext routingContextMock;
+
+    @Mock
+    HttpServerRequest requestMock;
+
+    @InjectMocks
+    MultiTenantResolver multiTenantResolver;
+
+    @Before
```
Sgitario: This is wrong. It should use "BeforeEach" annotation as this is running on JUnit 5.

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void setup(){
+        lenient().when(routingContextMock.request()).thenReturn(requestMock);
+    }
+
+    @Test
+    void resolveGraphiqlTenantTest() {
```
Sgitario, line 54: I think this test relies on some properties being injected from a maven profile.
In order to make it isolated, what do you think of creating a givenTenantConfigured where:

```
private void givenTentantIsConfigured() {
        multiTenantResolver.graphUITenantId = GRAPH_UI_TENANT;
        multiTenantResolver.graphUIPath = GRAPH_UI_PATH;
    }
```

Sgitario (changes requested):
Some minor things and an issue in one unit test.

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,44 @@
...
 
     void setupRouter(@Observes Router router) {
-        if (Boolean.FALSE.equals(authEnabled)) {
-            router.route("/graphiql/*").handler(GraphiQLHandler.create(new GraphiQLHandlerOptions().setEnabled(true)));
-            router.route().handler(LoggerHandler.create());
-            router.route().handler(StaticHandler.create());
-            router.route().handler(FaviconHandler.create());
-            router.route("/").handler(ctx -> ctx.response().putHeader("location", "/graphiql/").setStatusCode(302).end());
+        GraphiQLHandler graphiQLHandler = GraphiQLHandler.create(new GraphiQLHandlerOptions().setEnabled(true));
+        if (Boolean.TRUE.equals(authEnabled)) {
```
Sgitario, line 59: Should we cover these changes in VertxRouterSetup in an unit test?

Sgitario (changes requested):

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: done

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
nmirasch: updated

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
nmirasch: updated

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
nmirasch: added

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: added

nmirasch:
@Sgitario  I added unit test for VertxRouterSetup and added some of your suggestions. Could you take another look? Thanks!

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @InjectMocks
+    MultiTenantResolver multiTenantResolver;
+
+    @BeforeEach
+    public void setup(){
+        lenient().when(routingContextMock.request()).thenReturn(requestMock);
+        givenTenantIsConfigured();
+    }
+
+    private void givenTenantIsConfigured() {
```
Sgitario: I think private methods should go at the bottom after public and protected methods.

Sgitario (approved):
lgtm, just a minor thing.

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
nmirasch: @Sgitario Moved, Thanks!

nmirasch:
@cristianonicolai @jstastny-cz  could you take a look? thanks!

jstastny-cz (approved):

nmirasch:
jenkins please retest this

nmirasch:
jenkins please retest this

cristianonicolai:
jenkins please retest this

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
cristianonicolai: wouldnt be best to not set `.orElse(StringUtils.EMPTY);` then simply do `token.isPresent()` as an `Optional` ?

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,47 @@
...
+            addGraphiqlRequestHeader(graphiQLHandler);
         }
+        router.route(graphUIPath +"*").handler(graphiQLHandler);
```
cristianonicolai:
```suggestion
        router.route(graphUIPath + "*").handler(graphiQLHandler);
```

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,47 @@
...
+            MultiMap multiMap = MultiMap.caseInsensitiveMultiMap();
+            String token = getCurrentAccessToken(rc);
+            if (token.isEmpty()) {
+                token = rc.request().getHeader(HttpHeaders.AUTHORIZATION);
+            }
+            multiMap.add(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, "*");
+            multiMap.add(HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
+            multiMap.add(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS, "Authorization, Content-Type, Accept");
+            multiMap.add("Bearer-token", "Bearer " + token);
+            rc.request().headers().addAll(multiMap);
```
cristianonicolai: @nmirasch I think you don't need to set the headers manually, simply return the multimap that vertx will add it automatically?

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,47 @@
...
+        graphiQLHandler.graphiQLRequestHeaders(rc -> {
+            MultiMap multiMap = MultiMap.caseInsensitiveMultiMap();
+            String token = getCurrentAccessToken(rc);
+            if (token.isEmpty()) {
+                token = rc.request().getHeader(HttpHeaders.AUTHORIZATION);
+            }
+            multiMap.add(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, "*");
+            multiMap.add(HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
+            multiMap.add(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS, "Authorization, Content-Type, Accept");
+            multiMap.add("Bearer-token", "Bearer " + token);
```
cristianonicolai: Do we also need this one, isnt just the `AUTHORIZATION` header the one used to authenticate?

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,47 @@
...
         router.route("/graphql").handler(ApolloWSHandler.create(graphQL));
         router.route("/graphql").handler(GraphQLHandler.create(graphQL, new GraphQLHandlerOptions()));
     }
+
+    protected void addGraphiqlRequestHeader(GraphiQLHandler graphiQLHandler ){
+        graphiQLHandler.graphiQLRequestHeaders(rc -> {
+            MultiMap multiMap = MultiMap.caseInsensitiveMultiMap();
+            String token = getCurrentAccessToken(rc);
+            if (token.isEmpty()) {
+                token = rc.request().getHeader(HttpHeaders.AUTHORIZATION);
```
cristianonicolai: @nmirasch when you get token like this wouldnt it contain the starting string "Bearer " ?

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Mock
+    HttpServerRequest requestMock;
+
+    @InjectMocks
+    MultiTenantResolver multiTenantResolver;
+
+    @BeforeEach
+    public void setup(){
+        lenient().when(routingContextMock.request()).thenReturn(requestMock);
```
cristianonicolai: @nmirasch just curious why do we need to use lenient in these tests?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -31,18 +31,18 @@
...
+    private static final String KEYCLOAK_SERVER_URL = System.getProperty("quarkus.oidc.auth-server-url", "http://localhost:8281/auth/realms/kogito");
+    private static final String KEYCLOAK_CLIENT_ID = System.getProperty("quarkus.oidc.client-id", "kogito-data-index-service");
+    private static final String KEYCLOAK_CLIENT_SECRET = System.getProperty("quarkus.oidc.credentials.secret", "secret");
 
     @Test
     public void testUnauthorizedUserAccess() {
         //alice only have role User, resource is forbidden
-        given().auth().oauth2(getAccessToken("alice"))
-                .when().get("/graphiql/")
+        given().contentType(ContentType.JSON).body("{ \"query\" : \"{ProcessInstances{ id } }\" }")
```
cristianonicolai, line 43: @nmirasch I got confused here, are we just running the same get request twice?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
cristianonicolai: @nmirasch can we perhaps validate something in the body of the response that it actually redirected to Keycloack login page?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -65,27 +61,38 @@ public void testNoTokenProvided() {
...
+        // Returning keycloak login page
+        given().when().get("/graphiql/")
+                .then()
+                .statusCode(200);
+        // Returning keycloak login page
+        given().auth().oauth2(getAccessToken("jdoe"))
+                .when().get("/graphiql/")
+                .then()
+                .statusCode(200);
+        given().when().get("/other/")
```
cristianonicolai: not sure I get the point to check this endpoint as it will always return 404, with or without security :)

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @BeforeEach
+    public void setup(){
+        lenient().when(routerMock.route()).thenReturn(routeMock);
+        lenient().when(routerMock.route(anyString())).thenReturn(routeMock);
+    }
+
+    @Test
+    public void testAuthEnabledTrue() {
+        vertxRouterSetup.authEnabled = true;
+        vertxRouterSetup.graphUIPath="/graphiql";
```
cristianonicolai:
```suggestion
        vertxRouterSetup.graphUIPath = "/graphiql";
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+        vertxRouterSetup.setupRouter(routerMock);
+
+        verify(vertxRouterSetup).addGraphiqlRequestHeader(any());
+    }
+
+    @Test
+    public void testAuthEnabledFalse() {
+        vertxRouterSetup.authEnabled = false;
+        vertxRouterSetup.graphUIPath="/graphiql";
```
cristianonicolai:
```suggestion
        vertxRouterSetup.graphUIPath = "/graphiql";
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+        vertxRouterSetup.setupRouter(routerMock);
+
+        verify(vertxRouterSetup,never()).addGraphiqlRequestHeader(any());
```
cristianonicolai:
```suggestion
        verify(vertxRouterSetup, never()).addGraphiqlRequestHeader(any());
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+        when(requestMock.getHeader(HttpHeaders.AUTHORIZATION)).thenReturn(token);
+        when(requestMock.headers()).thenReturn(multiMapMock);
+
+        vertxRouterSetup.addGraphiqlRequestHeader(graphiQLHandlerMock);
+
+        ArgumentCaptor<Function<RoutingContext, MultiMap>> functionCaptor =ArgumentCaptor.forClass(Function.class);
+        verify(graphiQLHandlerMock).graphiQLRequestHeaders(functionCaptor.capture());
+        functionCaptor.getValue().apply(routingContextMock);
+
+        ArgumentCaptor<MultiMap> multimapCaptor =ArgumentCaptor.forClass(MultiMap.class);
```
cristianonicolai:
```suggestion
        ArgumentCaptor<MultiMap> multimapCaptor = ArgumentCaptor.forClass(MultiMap.class);
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+        HttpServerRequest requestMock = mock(HttpServerRequest.class);
+        MultiMap multiMapMock = mock(MultiMap.class);
+        String token = "TEST_TOKEN";
+        when(routingContextMock.request()).thenReturn(requestMock);
+        when(requestMock.getHeader(HttpHeaders.AUTHORIZATION)).thenReturn(token);
+        when(requestMock.headers()).thenReturn(multiMapMock);
+
+        vertxRouterSetup.addGraphiqlRequestHeader(graphiQLHandlerMock);
+
+        ArgumentCaptor<Function<RoutingContext, MultiMap>> functionCaptor =ArgumentCaptor.forClass(Function.class);
```
cristianonicolai:
```suggestion
        ArgumentCaptor<Function<RoutingContext, MultiMap>> functionCaptor = ArgumentCaptor.forClass(Function.class);
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+    public void testAuthEnabledFalse() {
+        vertxRouterSetup.authEnabled = false;
+        vertxRouterSetup.graphUIPath="/graphiql";
+        vertxRouterSetup.setupRouter(routerMock);
+
+        verify(vertxRouterSetup,never()).addGraphiqlRequestHeader(any());
+    }
+
+    @Test
+    public void testAddGraphiqlRequestHeader() {
```
cristianonicolai, line 98: @nmirasch I think it makes sense to also test when the token is provided by the request headers. perhaps a new test method?

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
```diff
@@ -0,0 +1,116 @@
...
+        vertxRouterSetup.addGraphiqlRequestHeader(graphiQLHandlerMock);
+
+        ArgumentCaptor<Function<RoutingContext, MultiMap>> functionCaptor =ArgumentCaptor.forClass(Function.class);
+        verify(graphiQLHandlerMock).graphiQLRequestHeaders(functionCaptor.capture());
+        functionCaptor.getValue().apply(routingContextMock);
+
+        ArgumentCaptor<MultiMap> multimapCaptor =ArgumentCaptor.forClass(MultiMap.class);
+        verify(multiMapMock).addAll(multimapCaptor.capture());
+
+        assertThat(multimapCaptor.getValue().get(HttpHeaders.AUTHORIZATION).equals("Bearer " + token));
```
cristianonicolai:
```suggestion
        assertThat(multimapCaptor.getValue().get(HttpHeaders.AUTHORIZATION)).isEqualTo("Bearer " + token);
```

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
cristianonicolai: do we actually need or is there a situation where the user could get to the GraphiQL UI and not be logged in or have access to it via a token in the authorization header? Perhaps we can simply send the current access token all the time?

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: yep, good catch absolutely right, Thanks!

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: done

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: I have simplified that headers, just passing the Authorization with the current logged user token.

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: agree, simplified :)

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
nmirasch: done

data-index/data-index-service/src/test/java/org/kie/kogito/index/vertx/VertxRouterSetupTest.java
nmirasch: done

nmirasch:
@cristianonicolai Thanks for the review. I have simplified the graphiql request header just to include the authorization one, and removed the code of retrieving the token from request headers regarding the entry point is the web app that is in charge to set that quarkus user with the credentials. And, finally I passed the reformat on files. Please could you take another look here?

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
nmirasch: yep double check removed

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
cristianonicolai: @nmirasch I think just this one needs some improvement. A 200 could also be returned if the auth is disabled right? Anything that checks that it actually reached the login page.

Sgitario:
@nmirasch can you add the next into the [kogito-realm.json](https://github.com/kiegroup/kogito-apps/blob/82a4331f72edc512219a2013656bf821ce9fa6dd/data-index/data-index-service/src/test/resources/kogito-realm.json#L513)?

```
"redirectUris": [
        "*"
      ],
```

This way the redirect from keycloak to the app will work.

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
```diff
@@ -0,0 +1,53 @@
...
+    String graphUIPath;
+
+    @Inject
+    @ConfigProperty(name = "kogito.data-index.vertx-graphql.ui.tenant", defaultValue = "web-app-tenant")
+    String graphUITenantId;
+
+    @Override
+    public String resolve(RoutingContext context) {
+        if (graphUIPath != null
+                && graphUITenantId != null
```
Sgitario: If we provide the default values above, the graphUIPath and graphUITentantId will never be null.

nmirasch:
@Sgitario  @cristianonicolai Renamed the keycloak client to kogito-service with the aim to reuse that to other kogito services and not to have to define a client to every one. I have updated the kogito-realm and added that to kogito-apps/config as a generic auth config  resource. Added some logic to allow "/graphiql" and "/graphiql/" as entry points to graphql UI. 
Could you  take a look and let me know what you think? Thanks!!

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,46 @@
...
+
+    private String getCurrentAccessToken(RoutingContext routingContext) {
+        return Optional.ofNullable(routingContext.user())
+                .map(user -> ((QuarkusHttpUser) user).getSecurityIdentity())
+                .map(identity -> identity.getCredential(AccessTokenCredential.class))
+                .map(AccessTokenCredential::getToken)
+                .orElse(StringUtils.EMPTY);
+    }
+
+    private String getGraphUIAllowedPath() {
```
cristianonicolai: @nmirasch Can this be a reusable utils function instead of duplicating it?

nmirasch: @cristianonicolai  yes, I thought that but I'm not sure where makes sense to have it. I had the temptation to add a property for that.

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
```diff
@@ -0,0 +1,53 @@
...
+
+import io.quarkus.oidc.TenantResolver;
+import io.vertx.ext.web.RoutingContext;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@ApplicationScoped
+public class MultiTenantResolver implements TenantResolver {
+
+    @Inject
+    @ConfigProperty(name = "kogito.data-index.vertx-graphql.ui.path", defaultValue = "/graphiql/")
```
cristianonicolai: should the default be just _/graphiql_ ?

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
```diff
@@ -39,18 +47,46 @@
...
+        GraphiQLHandler graphiQLHandler = GraphiQLHandler.create(new GraphiQLHandlerOptions().setEnabled(true));
+        if (Boolean.TRUE.equals(authEnabled)) {
+            addGraphiqlRequestHeader(graphiQLHandler);
         }
+        router.route(graphUIPath + "*").handler(graphiQLHandler);
+        router.route().handler(LoggerHandler.create());
+        router.route().handler(StaticHandler.create());
+        router.route().handler(FaviconHandler.create());
+        router.route(getGraphUIAllowedPath()).handler(ctx -> ctx.response().putHeader("location", graphUIPath).setStatusCode(302).end());
+        router.route("/").handler(ctx -> ctx.response().putHeader("location", graphUIPath).setStatusCode(302).end());
```
Sgitario: In order to avoid duplicate code here, does it make sense to have something like:

```
router.route(graphUIPath + "*").handler(setRootLocation());
router.route("/").handler(setRootLocation());
// ...

protected Handler<RoutingContext> setRootLocation() {
     return ctx -> ctx.response().putHeader("location", graphUIPath).setStatusCode(302).end();
}
```

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
cristianonicolai: to be honest, I dont see much value to allow customizing this endpoint.

nmirasch: @cristianonicolai  agree, I keep that as it was before

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
nmirasch: updated

data-index/data-index-service/src/main/java/org/kie/kogito/index/auth/MultiTenantResolver.java
nmirasch: @Sgitario yes, removed!

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/VertxRouterSetup.java
nmirasch: @Sgitario Regarding I just restored the previous served routes, This is not needed now, but thanks!

nmirasch:
@Sgitario @cristianonicolai updated! when you have the chance, could you give thiis a thy? thanks!!!

data-index/data-index-service/src/test/java/org/kie/kogito/index/auth/MultiTenantResolverTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class MultiTenantResolverTest {
+
+    private static final String GRAPH_UI_PATH = System.getProperty("kogito.data-index.vertx-graphql.ui.path", "/graphiql/");
```
cristianonicolai:
```suggestion
    private static final String GRAPH_UI_PATH = System.getProperty("kogito.data-index.vertx-graphql.ui.path", "/graphiql");
```

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
cristianonicolai: @nmirasch ^

data-index/data-index-service/src/test/java/org/kie/kogito/index/service/KeycloakIntegrationIndexingServiceIT.java
```diff
@@ -65,27 +56,43 @@ public void testNoTokenProvided() {
...
                 .when().post("/graphql")
                 .then().log().ifValidationFails().statusCode(200);
     }
 
+    @Test
+    public void graphiqlLoginOnKeycloakServedTest() {
+        // Returning keycloak login page
+        given().when().get("/graphiql/")
+                .then()
+                .statusCode(200);
```
cristianonicolai, line 73: @nmirasch I think just this one needs some improvement. A 200 could also be returned if the auth is disabled right? Anything that checks that it actually reached the login page.

nmirasch:
@cristianonicolai  added assertion to check the login page returned

cristianonicolai (approved):

PR #249, opened 2020-06-04
KOGITO-2299 - Add namespace for GraphQL types

cristianonicolai:

### DIFF
--- a/packages/common/codegen.yml
+++ b/packages/common/codegen.yml
@@ -5,6 +5,13 @@ generates:
   ./src/graphql/types.tsx:
     plugins:
       - add: '/* tslint:disable */'
+      - add:
+          content:
+          - ''
+          - 'export namespace GraphQL {'
+      - add:
+          placement: 'append'
+          content: '}'
       - "typescript"
       - "typescript-operations"
       - "typescript-react-apollo"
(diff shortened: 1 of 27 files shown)


cristianonicolai:
@jstastny-cz this is what I could come up for adding some sort of namespace for the graphql types. Will let others comment if they like it or not, or maybe some other alternative.

jstastny-cz (approved):

nmirasch (approved):

Sara4994:
@cristianonicolai i tried running app with these changes and i get this error "TS2305: Module '"../../../../../../node_modules/@kogito-apps/common"' has no exported member 'GraphQL'."on all the files. Am i missing something to do? 

Also a small suggestion, Can we export something like below from common/index.ts itself, which makes more clear that queries and types can be exported as "GraphQL". because right now i see "export * from './src/graphql/types';" on index.ts which doesnt say about namespace(I see that namespace on codegen.yml). wdyt?

import * as GraphQL from './src/graphql/types';
export { GraphQL }

cristianonicolai:
@Sara4994 try running `yarn run build:prod` first

your suggestion is a lot cleaner, the only downside I see is that in the common package the import would be different. Because in there we should have access to the raw types and not the wrapped namespace. I think is a bit weird to have two different ways to import these types, wdyt?

AjayJagan (approved):

Sara4994:
> @Sara4994 try running `yarn run build:prod` first
> 
> your suggestion is a lot cleaner, the only downside I see is that in the common package the import would be different. Because in there we should have access to the raw types and not the wrapped namespace. I think is a bit weird to have two different ways to import these types, wdyt?

@cristianonicolai yep that makes sense :+1:

Sara4994 (approved):

PR #259, opened 2020-06-09
KOGITO-2416 - Review KogitoEmptyState component for improved reusability

cristianonicolai:

### DIFF
--- a/packages/common/src/components/Atoms/KogitoEmptyState/tests/__snapshots__/KogitoEmptyState.test.tsx.snap
+++ b/packages/common/src/components/Atoms/KogitoEmptyState/tests/__snapshots__/KogitoEmptyState.test.tsx.snap
@@ -1,30 +1,299 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Emptystate component tests snapshot testing 1`] = `
-<Bullseye>
-  <EmptyState
-    variant="full"
-  >
-    <EmptyStateIcon
-      color="var(--pf-global--warning-color--100)"
-      icon={[Function]}
-      size="sm"
-    />
-    <Title
-      headingLevel="h5"
-      size="lg"
+exports[`KogitoEmptyState component tests Info test 1`] = `
+<Component
+  body="This process has no related sub processes"
+  title="No child process instances"
+  type={3}
+>
+  <Bullseye>
+    <div
+      className="pf-l-bullseye"
     >
-      No child process instances
-    </Title>
-    <EmptyStateBody>
-      This process has no related sub processes
-    </EmptyStateBody>
-    <Component
-      onClick={[Function]}
-      variant="primary"
+      <EmptyState
+        variant="full"
+      >
+        <div
+          className="pf-c-empty-state"
+        >
+          <EmptyStateIcon
+            color="var(--pf-global--info-color--100)"
+            icon={[Function]}
+            size="sm"
+          >
+            <InfoCircleIcon
+              aria-hidden="true"
+              className="pf-c-empty-state__icon"
+              color="var(--pf-global--info-color--100)"
+              size="sm"
+            >
+              <MockedComponent />
+            </InfoCircleIcon>
+          </EmptyStateIcon>
+          <Title
+            headingLevel="h5"
+            size="lg"
+          >
+            <h5
+              className="pf-c-title pf-m-lg"
+            >
+              No child process instances
+            </h5>
+          </Title>
+          <EmptyStateBody>
+            <div
+              className="pf-c-empty-state__body"
+            >
+              This process has no related sub processes
+            </div>
+          </EmptyStateBody>
+        </div>
+      </EmptyState>
+    </div>
+  </Bullseye>
+</Component>
+`;
+
+exports[`KogitoEmptyState component tests Refresh test 1`] = `
+<Component
+  body="This process has no related sub processes"
+  onClick={[MockFunction]}
+  title="No child process instances"
+  type={1}
+>
+  <Bullseye>
+    <div
+      className="pf-l-bullseye"
     >
-      Refresh
-    </Component>
-  </EmptyState>
-</Bullseye>
+      <EmptyState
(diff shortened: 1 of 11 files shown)


cristianonicolai:
@AjayJagan I tried to decouple the empty state from the functions that are needed to execute once the button is clicked. Also doesn't seem that running _onFIlterClick_ as an async function is necessary, see https://github.com/kiegroup/kogito-apps/pull/259/files#diff-1818f66c4b3f8e8abf534ef55f6981f7R85 or am I missing something?

cristianonicolai:
@jstastny-cz we might need to find another way to wrap ouia, removing the default export makes a lot easier to handle and it seems to make harder for when you need to export multiple types, see https://github.com/kiegroup/kogito-apps/pull/259/files#diff-9395e6e344be30633b1723848e8871aaR32

cristianonicolai:
will leave this PR open until https://issues.redhat.com/browse/KOGITO-2386 is fixed and test is properly executed in the common module.

AjayJagan:
> @AjayJagan I tried to decouple the empty state from the functions that are needed to execute once the button is clicked. Also doesn't seem that running _onFIlterClick_ as an async function is necessary, see https://github.com/kiegroup/kogito-apps/pull/259/files#diff-1818f66c4b3f8e8abf534ef55f6981f7R85 or am I missing something?

@cristianonicolai  - yep async can be removed(my bad .. sorry)

jstastny-cz:
> @jstastny-cz we might need to find another way to wrap ouia, removing the default export makes a lot easier to handle and it seems to make harder for when you need to export multiple types, see https://github.com/kiegroup/kogito-apps/pull/259/files#diff-9395e6e344be30633b1723848e8871aaR32

Ok I can rewrite it after using the 'rename export':
```
const Component:React.FC<> = ({}){
.
.
}
const ComponentWithOuia = withOuiaContext(Component)
export {ComponentWithOuia as Component}
```

This is actually how it's implemented in patternfly version we use. When you look in master, there's no such thing ... The reason for this wrapper was IMO to enable turn on/off of the ouia support, while in latest it's on by default....

cristianonicolai:
@AjayJagan @jstastny-cz ready for review.

packages/common/src/components/Atoms/KogitoEmptyState/tests/__snapshots__/KogitoEmptyState.test.tsx.snap
```diff
@@ -1,6 +1,29 @@
...
 
-exports[`Emptystate component tests snapshot testing 1`] = `
+exports[`KogitoEmptyState component tests Info test 1`] = `
+<Bullseye>
+  <EmptyState
+    variant="full"
+  >
+    <EmptyStateIcon
+      color="var(--pf-global--info-color--100)"
+      icon={[Function]}
```
jstastny-cz: the differentiation based on types is not visible in these shallow snapshots. Is it possible to use mount instead? Maybe we can even mock the EmptyStateIcon, but using mount we should see the actual icon passed to the component.

jstastny-cz (commented):
Just one question about snapshots, looks good otherwise.

packages/common/src/components/Atoms/KogitoEmptyState/tests/__snapshots__/KogitoEmptyState.test.tsx.snap
cristianonicolai: yeah, makes sense. So now I changed to `mount` but I couldn't figure how to mock only the `EmptyStateIcon`. I could only mock the entire `react-core` module. @AjayJagan any tips on how to do that?
This fails with finding the module.
`jest.mock('@patternfly/react-core/EmptyState/EmptyStateIcon');`
And this ends up mocking all components, then it stops at the `Bullseye` component which is not what we want.
`jest.mock('@patternfly/react-core')`

jstastny-cz: You can do this to mock just a single component from whole package:
```
const MockedComponent = (): React.ReactElement => {
  return <></>;
};

jest.mock('@patternfly/react-core', () => ({
  ...jest.requireActual('@patternfly/react-core'),
  EmptyStateIcon: () => {
    return <MockedComponent />;
  }
}));
```

cristianonicolai: mock works but then I get the following in the snapshot:

```
<EmptyStateIcon
            color="var(--pf-global--warning-color--100)"
            icon={[Function]}
            size="sm"
          >
        <EmptyStateIconMock />
</EmptyStateIcon>
```

jstastny-cz: ah, sorry, I didn't realize, so we can either keep it full or just mock the icons passed to EmptyStateIcon. That would mean mocking ''@patternfly/react-icons'' instead:

```
const MockedComponent = (): React.ReactElement => {
  return <></>;
};

jest.mock('@patternfly/react-icons', () => ({
  ...jest.requireActual('@patternfly/react-icons'),
  InfoCircleIcon: () => {
    return <MockedComponent />;
  },
  ExclamationTriangleIcon: () => { return <MockedComponent /> },
.
.
.
}));
```

jstastny-cz: We'd still save ~80 lines out of 380 in the snapshots. And reducing the noise.

cristianonicolai: @jstastny-cz thanks that helped. Have a look now

AjayJagan (approved):
working fine :+1:

jstastny-cz (approved):

PR #261, opened 2020-06-09
KOGITO-2296 check code coverage for management-console

jstastny-cz:
Including Management Console for inclusion in code coverage statistics by invoking the `test:coverage` check from the management-console maven module (not to be mistaken with packages/management-console package).

PR #264, opened 2020-06-10
KOGITO-2417 use single module for ui services

jstastny-cz:
This brings a slightly different structure of the repo moving all UI quarkus services to a single maven module called ui-services (name proposals welcome).
Together with that to use `lerna` during build, the actual build of UI apps (lerna monorepo with `package.json` in repo root + apps in `packages`) is done in a separate sub-module `ui-services/build-packages`. This dependency is to be added as pom dependency to the quarkus projects building the ui services themselves (e.g. `ui-services/management-console`).
Goals:
* UI app MUST be built only once for all the quarkus ui services
* UI app build MUST be done using lerna (finally in sync with how is configured)
* `build-packages` module MUST be able to collect code-coverage metrics of all ui packages
* respective Quarkus UI apps can contain additional classes and MUST be able to report code coverage metrics.
* A Quarkus UI app CAN be added as a module to `ui-services` while adding the `build-packages` dependency to assure `packages` are built before the actual service. The only outstanding action is to copy the correct `dist` folder into the target before build (as `management-console` was doing already)


jstastny-cz:
Outdating #261 if working properly.

cristianonicolai:
@jstastny-cz what about using the root pom to execute the lerna build? that way the build would be done first and the other quarkus app would just need to copy the dist folder. wdyt?

jstastny-cz:
@cristianonicolai I wanted to achieve two things:
* build UI apps when necessary
* build them just once per build

If I would put the UI app build into a parent `pom.xml` of the Quarkus ui service (considering `management-console` and `task-inbox` here) I am afraid that the profiles would be inherited by the modules and thus the app would be built for each module separately (twice in this case). Let me verify this first.
Anyway this separation of build of the `UI packages` into a separate 'sibling' module will help me in isolating test coverage statistics - cause we want to report those also for `management-console` quarkus app, but combining both coverage statistics from quarkus app and `UI packages` does not seem straight forward to me (I don't know how to do that easily).

jstastny-cz:
The current state looks like: 
Screenshot from 2020-06-11 12-29-57

And within packages:
Screenshot from 2020-06-11 12-30-45

And inside:
Screenshot from 2020-06-11 12-31-21

So outstanding task is to cover only `src` subfolders into the metrics

jstastny-cz:
jenkins retest this

jstastny-cz:
Resolved by adding `<sonar.inclusions>*/src/**</sonar.inclusions>` so resulting state in `packages` is 
Screenshot from 2020-06-11 21-05-19

Which is expected and desirable.

Note - the metrics won't be visible in the sonarcheck for this PR, only after merging to master -> cause in the check for PR are statistics just for the changed code and I didn't touch `packages`. It cost me whole afternoon to realize.

ui-services/build-packages/pom.xml
```diff
@@ -4,16 +4,27 @@
...
   
-  <artifactId>management-console</artifactId>
-  <name>Kogito :: Management Console</name>
+  <artifactId>kogito-apps-ui-packages-build</artifactId>
+  <name>Kogito Apps :: UI Packages Build</name>
 
   <properties>
-    <path.to.frontend.app>../packages/management-console</path.to.frontend.app>
+    <path.to.root>${project.basedir}/../../</path.to.root> <!-- package.json in root dir-->
+    <sonar.projectBaseDir>${path.to.root}/packages/</sonar.projectBaseDir>
```
rsynek, line 18: Nitpicking: the `${path.to.root}/packages` is used several times, thus I would make it a new property.

ui-services/build-packages/pom.xml
```diff
@@ -4,16 +4,27 @@
...
+    <sonar.sources>${path.to.root}/packages/</sonar.sources>
+    <sonar.exclusions>**/__mocks__/**</sonar.exclusions>
+    <sonar.inclusions>*/src/**</sonar.inclusions>
+    <sonar.test.inclusions>**/*test.ts,**/*test.tsx</sonar.test.inclusions>
+    <sonar.javascript.lcov.reportPaths>
+      ${path.to.root}packages/common/coverage/lcov.info,
```
rsynek, line 24: I find it slightly more readable if the `path.to.root` does not end with `/` so that the path delimiter can be used here.

rsynek (approved):
Reviewed mainly from the point of view of test coverage measurement and SonarCloud integration. Just a couple of nitpicks, nice work!

nmirasch (approved):

jstastny-cz:
Closing in favor of #275

PR #275, opened 2020-06-17
KOGITO-2417 ui-packages for building the react app

jstastny-cz:
Variant of #264


pom.xml
```diff
@@ -65,5 +71,6 @@
     <module>management-console</module>
     <module>explainability-service</module>
     <module>trusty</module>
+    <module>ui-packages</module>
```
cristianonicolai: technically it shouldnt make difference but can you add it before the management-console module please.

ui-packages/pom.xml
```diff
@@ -0,0 +1,173 @@
...
+    <sonar.inclusions>*/src/**</sonar.inclusions>
+    <sonar.test.inclusions>**/*test.ts,**/*test.tsx</sonar.test.inclusions>
+    <sonar.javascript.lcov.reportPaths>
+      ${path.to.packages}/common/coverage/lcov.info,
+      ${path.to.packages}/management-console/coverage/lcov.info,
+      ${path.to.packages}/task-console/coverage/lcov.info
+    </sonar.javascript.lcov.reportPaths>
+  </properties>
+  
+  <dependencyManagement>
```
cristianonicolai: @jstastny-cz do we need these deps here?

ui-packages/packages/common/index.ts
```diff
@@ -0,0 +1,38 @@
+export {
```
cristianonicolai: this is the only file that is showing as new instead of a move, is that just the github UI?

cristianonicolai (commented):
@jstastny-cz looks better to me

ui-packages/packages/common/index.ts
jstastny-cz: It is some git magic, I've tried to revert and use `git mv`, but the result was the same. I don't understand why this happens especially when the file is 1-1 the same.

pom.xml
jstastny-cz: sure, no problem.

ui-packages/packages/common/index.ts
jstastny-cz: Ah! It got reformatted heavily. I just moved it :-D . That's the crazy stuff to make our code pretty doing this!

ui-packages/pom.xml
jstastny-cz: Nope, correct, bad copy&paste -> we don't use this as parent or use any deps directly. Will remove.

jstastny-cz: On second look the version for eirslett might be set there ... otherwise I'd expect the build to fail - will investigate.

jstastny-cz: Nope, it's managed in kogito-apps/pom.xml removing section.

cristianonicolai (approved):

jstastny-cz:
Still one outstanding issue during build, investigating.
```
[ERROR] Install TypeScript in the project directory or use NODE_PATH env. variable to set TypeScript location, if it's located outside of project directory.
```

pefernan (approved):
+1 to merge... It will cause a lot of pain on my PR but we are here to play!

nmirasch (approved):
Really nice! Thanks Jan

PR #289, opened 2020-06-22
KOGITO-2506 specify multiple sonar.sources instead of using inclusions

jstastny-cz:
sonar.inclusions property doesn't work as desirable.

Instead of using it this PR specifies comma delimited locations of sources within particular packages.


cristianonicolai (approved):

jstastny-cz:
jenkins retest this

PR #293, opened 2020-06-24
[KOGITO-2121] MongoDB for Data Index service

xieshenzh:
Jira: https://issues.redhat.com/browse/KOGITO-2121
        https://issues.redhat.com/browse/KOGITO-2122
        https://issues.redhat.com/browse/KOGITO-2146
 
Description: 
Support mongodb as persistence solution for data index. Support indexing and health check with mongodb. 


xieshenzh:
@cristianonicolai @danielezonca @r00ta @rhtevan This is the pull request for mongodb. Please take a look. Thanks.

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>data-index-storage-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>persistence-commons-mongodb</artifactId>
+      <version>${project.version}</version>
```
danielezonca: Can you please move version definition to [kogito-runtimes bom](https://github.com/kiegroup/kogito-runtimes/blob/master/bom/pom.xml)?

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>persistence-commons-mongodb</artifactId>
+      <version>${project.version}</version>
```
danielezonca: Can you please move version definition to [kogito-runtimes bom](https://github.com/kiegroup/kogito-runtimes/blob/master/bom/pom.xml)?

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemPropertyVariables>
+            <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+            <mongodb.version>${version.org.mongodb.image}</mongodb.version>
+            <quarkus.test.profile>mongodb</quarkus.test.profile>
```
danielezonca: What is this used for?

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/DomainEntity.java
```diff
@@ -0,0 +1,50 @@
...
+public class DomainEntity {
+
+    public static ObjectNode toObjectNode(String id, Document document) {
+        if (document == null) {
+            return null;
+        }
+
+        ObjectNode node = ModelUtils.documentToJsonNode(document, ObjectNode.class);
+        node.remove(MongoOperations.ID);
+        node.put(DomainStorage.ID, id);
```
danielezonca: Can you explain the reason of this? (same question for fromObjectNode)

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/utils/QueryUtils.java
```diff
@@ -0,0 +1,79 @@
...
+            case CONTAINS_ALL:
+                return format("%s: { $all: [ %s ] }", filterAttributeFunction.apply(filter.getAttribute()), ((List) filter.getValue()).stream().map(v -> filterValueFunction.apply(filter.getAttribute(), v)).collect(joining(", ")));
+            case CONTAINS_ANY:
+                return format("$or: [ %s ]", ((List) filter.getValue()).stream().map(v -> filterValueFunction.apply(filter.getAttribute(), v)).map(v -> format("{ %s: { $in: [ %s ] } }", filterAttributeFunction.apply(filter.getAttribute()), v)).collect(joining(", ")));
+            case OR:
+                return format("$or: [ %s ]", ((List<AttributeFilter<?>>) filter.getValue()).stream().map(f -> format("{ %s }", generateSingleQueryString(f, filterAttributeFunction, filterValueFunction))).collect(joining(", ")));
+            case AND:
+                return format("$and: [ %s ]", ((List<AttributeFilter<?>>) filter.getValue()).stream().map(f -> format("{ %s }", generateSingleQueryString(f, filterAttributeFunction, filterValueFunction))).collect(joining(", ")));
+            default:
+                return null;
```
r00ta: I think here you can use some operators defined in the MongoDB client. Something like
```
import static com.mongodb.client.model.Filters.and;
import static com.mongodb.client.model.Filters.eq;
import static com.mongodb.client.model.Filters.gt;
import static com.mongodb.client.model.Filters.gte;
import static com.mongodb.client.model.Filters.in;
import static com.mongodb.client.model.Filters.lt;
import static com.mongodb.client.model.Filters.lte;
import static com.mongodb.client.model.Filters.not;
import static com.mongodb.client.model.Filters.or;
import static com.mongodb.client.model.Filters.regex;



    @Override
    public List<T> execute() {
        List<Bson> conditions = filters.stream().map(x -> createCondition(x)).collect(Collectors.toList());
        Iterator<T> it = collection.find(and(conditions)).iterator();

        List<T> copy = new ArrayList<T>();
        while (it.hasNext()) {
            copy.add(it.next());
        }
        return copy;
    }

    private Bson createCondition(AttributeFilter filter) {
        switch (filter.getCondition()) {
            case CONTAINS:
                return regex(filter.getAttribute(), getValueForQueryString(filter.getValue()));
            case CONTAINS_ALL:
                return and((Iterable<Bson>) ((List) filter.getValue()).stream().map(x -> regex(filter.getAttribute(), getValueForQueryString(x))).iterator());
            case CONTAINS_ANY:
                return or((Iterable<Bson>) ((List) filter.getValue()).stream().map(x -> regex(filter.getAttribute(), getValueForQueryString(x))).iterator());
            case LIKE:
                return regex(filter.getAttribute(), getValueForQueryString(filter.getValue()) + "*");
            case EQUAL:
                return eq(filter.getAttribute(), filter.getValue());
            case IN:
                return in(filter.getAttribute(), ((List) filter.getValue()).stream().iterator());
            case IS_NULL:
                return eq(filter.getAttribute(), null);
            case NOT_NULL:
                return not(eq(filter.getAttribute(), null));
            case BETWEEN:
                List<Object> value = (List<Object>) filter.getValue();
                return and(gte(filter.getAttribute(), value.get(0)), lte(filter.getAttribute(), value.get(1)));
            case GT:
                return gt(filter.getAttribute(), filter.getValue());
            case GTE:
                return gte(filter.getAttribute(), filter.getValue());
            case LT:
                return lt(filter.getAttribute(), filter.getValue());
            case LTE:
                return lte(filter.getAttribute(), filter.getValue());
            case OR:
                return or(createCondition(filter));
            case AND:
                return and(createCondition(filter));
            default:
                return null;
        }
    }
    private static String getValueForQueryString(Object value) {
        return value instanceof String ? "'" + value + "'" : value.toString();
    }
```
If it works.. we would have less code, easier to understand

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/AbstractStorage.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public void addObjectUpdatedListener(Consumer<V> consumer) {
+        this.objectUpdatedListener = consumer;
+    }
+
+    @Override
+    public void addObjectRemovedListener(Consumer<K> consumer) {
+        this.objectRemovedListener = consumer;
+    }
```
r00ta: Just a very small side note that is not really related to this PR. Isnt't better to rename them `setObjectCreatedListener` since there can be just one listener?

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/AbstractStorage.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public V put(K s, V v) {
+        V oldValue = this.get(s);
+        Optional.ofNullable(v).map(n -> mapToEntity(s, n)).ifPresent(
+                e -> getCollection().replaceOne(
+                        new BsonDocument(MongoOperations.ID, new BsonString(s.toString())),
+                        e, new ReplaceOptions().upsert(true)));
+        Optional.ofNullable(oldValue).ifPresentOrElse(o -> Optional.ofNullable(this.objectUpdatedListener).ifPresent(l -> l.accept(v)),
+                                                      () -> Optional.ofNullable(this.objectCreatedListener).ifPresent(l -> l.accept(v)));
+        return oldValue;
```
r00ta: Shouldn't return `v` instead of the `oldvalue`?

r00ta (commented):
I see that the logic for the mongo persistence is a bit different from the one we've implemented with Infinispan. The implementation of the Storage is delegated to the consumer (in this case `data-index-storage-mongo`). As well as the cache serving is actually provided by the consumer with `MongoDBStorageFactoryImpl`. Is there any specific reason for that? Isnt't possible to keep the logic in the common module so that we remove this logic from the consumer? 
My idea was to have something like 
```
public class MongoStorageManager implements StorageService {

    private static final Logger LOGGER = LoggerFactory.getLogger(MongoStorageManager.class);

    private MongoDatabase database;

    @Inject
    MongoClient defaultMongoClient;

    // TODO: INJECT PROPERTY DB NAME
    @PostConstruct
    void setup() {
        database = defaultMongoClient.getDatabase("myMongoDb");
    }

    @Override
    public Storage<String, String> getCache(String name) {
        return null;
    }

    @Override
    public <T> Storage<String, T> getCache(String name, Class<T> type) {
        return new MongoStorageImpl<>(getOrCreateCollection(name, type), type.getName());
    }

    @Override
    public <T> Storage<String, T> getCacheWithDataFormat(String name, Class<T> type, String rootType) {
        return null;
    }

    private MongoCollection getOrCreateCollection(String collection, Class type) {
        return database.getCollection(collection, type);
    }
}
```
And `MongoStorageImpl` is something like 
```

public class MongoStorageImpl<K, V> implements Storage<K, V> {

    private MongoCollection collection;
    private String rootType;

    public MongoStorageImpl(MongoCollection collection, String rootType) {
        this.collection = collection;
        this.rootType = rootType;
    }

    @Override
    public void addObjectCreatedListener(Consumer<V> consumer) {

    }

    @Override
    public void addObjectUpdatedListener(Consumer<V> consumer) {

    }

    @Override
    public void addObjectRemovedListener(Consumer<K> consumer) {

    }

    @Override
    public Query<V> query() {
        return new MongoQuery<>(collection, rootType);
    }

    @Override
    public V get(Object key) {
        KogitoDocument<K, V> document = (KogitoDocument<K, V>) collection.find(eq("entryId", key)).iterator().next();
        return document.getValue();
    }

    @Override
    public V put(K key, V value) {
        collection.insertOne(new KogitoDocument(key, value));
        return value;
    }

    @Override
    public V remove(Object key) {
        KogitoDocument<K, V> document = (KogitoDocument<K, V>) collection.findOneAndDelete(eq("entryId", key));
        return document.getValue();
    }

    @Override
    public boolean containsKey(K key) {
        return collection.countDocuments(eq("entryId", key)) > 0;
    }

    @Override
    public Set<Map.Entry<K, V>> entrySet() {
        Iterator<KogitoDocument> it = collection.find().iterator();

        List<KogitoDocument> copy = new ArrayList<KogitoDocument>();
        while (it.hasNext()) {
            copy.add(it.next());
        }

        return new HashSet<>(copy);
    }

    @Override
    public void clear() {
        collection.drop();
    }

    @Override
    public String getRootType() {
        return rootType;
    }
}
```

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
```diff
@@ -0,0 +1,210 @@
...
+        entity.lastUpdate = zonedDateTimeToInstant(instance.getLastUpdate());
+        entity.businessKey = instance.getBusinessKey();
+        return entity;
+    }
+
+    public static class NodeInstanceEntity {
+
+        public String _id;
+
+        public String id;
```
danielezonca: Why two different ids?

persistence-commons/persistence-commons-protobuf/src/test/resources/test.proto
```diff
@@ -4,45 +4,19 @@ package org.acme.travels.travels;
...
-    optional string bookingNumber = 2;
-    /* @Field(store = Store.YES) */
-    optional string name = 3;
-    /* @Field(store = Store.YES) */
-    optional string phone = 4;
-    /* @Field(store = Store.YES) */
-    optional string room = 5;
+    optional string name = 1;
+
+    optional string room = 2;
```
danielezonca, line 19: What is the reason for the changes of this class? Unused fields?

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/index/ProcessIndexEvent.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.kogito.persistence.mongodb.index;
+
+import java.util.Objects;
+
+import org.kie.kogito.persistence.api.schema.ProcessDescriptor;
+
+public class ProcessIndexEvent {
```
danielezonca, line 23: I think it makes sense to refactor common module `schema` package to abstract over [ProcessDescriptor](https://github.com/kiegroup/kogito-apps/blob/master/persistence-commons/persistence-commons-api/src/main/java/org/kie/kogito/persistence/api/schema/ProcessDescriptor.java) because it should be generic. In this way it should be possible to remove/update also this `ProcessIndexEvent`.

@r00ta @xiezhang7 @rhtevan
Wdyt? Should we create a ticket for that?

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/query/DomainQueryTest.java
```diff
@@ -0,0 +1,109 @@
...
+        storage.clear();
+    }
+
+    @Test
+    void test() {
+        String processInstanceId1 = UUID.randomUUID().toString() + "_process1";
+        String processInstanceId2 = UUID.randomUUID().toString();
+
+        ObjectNode node1 = TestUtils.createDomainData(processInstanceId1, "John", "Doe");
+        QueryTestBase.testSleep();
```
danielezonca: What is the reason of this sleep?

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/MongoDBStorageService.java
```diff
@@ -0,0 +1,49 @@
...
+    MongoDBStorageFactory mongoDBStorageFactory;
+
+    @Override
+    public Storage<String, String> getCache(String name) {
+        return (Storage<String, String>) mongoDBStorageFactory.getStorage(name);
+    }
+
+    @Override
+    public <T> Storage<String, T> getCache(String name, Class<T> type) {
+        return (Storage<String, T>) mongoDBStorageFactory.getStorage(name);
```
danielezonca: The semantic of all these methods is `getOrCreate` so can you please use `mongoDBStorageFactory.getOrCreateStorage` also in these two methods?
Actually I think it could be nice to rename `getCache` methods to better explain the idea but let's do it in another PR because otherwise we will have a lot of code changes and review it will be harder

persistence-commons/persistence-commons-mongodb/src/test/java/org/kie/kogito/persistence/mongodb/MongoDBServerTestResource.java
```diff
@@ -0,0 +1,54 @@
...
+public class MongoDBServerTestResource implements QuarkusTestResourceLifecycleManager {
+
+    private static final String MONGODB_VERSION = System.getProperty("mongodb.version");
+    private static final Logger LOGGER = LoggerFactory.getLogger(MongoDBServerTestResource.class);
+    private GenericContainer mongoDB;
+
+    @Override
+    public Map<String, String> start() {
+        if (MONGODB_VERSION == null) {
+            throw new RuntimeException("Please define a valid MongoDB image version in system property mongodb.version");
```
danielezonca: Please consider [this](https://github.com/kiegroup/kogito-apps/blob/6505bb3886f0184506e6ed2fd18fed09534667b3/persistence-commons/persistence-commons-infinispan/src/test/java/org/kie/kogito/persistence/infinispan/InfinispanServerTestResource.java#L47-L51) and relative quarkus issue

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/storage/MongoDBStorageFactoryImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+import static org.kie.kogito.index.Constants.JOBS_STORAGE;
+import static org.kie.kogito.index.Constants.PROCESS_ID_MODEL_STORAGE;
+import static org.kie.kogito.index.Constants.PROCESS_INSTANCES_STORAGE;
+import static org.kie.kogito.index.Constants.USER_TASK_INSTANCES_STORAGE;
+import static org.kie.kogito.index.mongodb.Constants.DOMAIN_COLLECTON_NAME_AFFIX;
+import static org.kie.kogito.index.mongodb.Constants.getDomainCollectionName;
+
+@ApplicationScoped
+public class MongoDBStorageFactoryImpl implements MongoDBStorageFactory {
```
danielezonca: Is it possible to move this impl (or part of it) to common module?
I think it could be nice if we can limit code user has to write to use this library to the different implementations of `AbstractStorage` and then leverage on CDI to retrieve in a generic way all the components.
Wdyt?

persistence-commons/persistence-commons-protobuf/src/test/java/org/kie/kogito/persistence/protobuf/ProtoIndexParserTest.java
```diff
@@ -0,0 +1,93 @@
...
+        SerializationContext ctx = new SerializationContextImpl(configureBuilder().build());
+        FileDescriptorSource invalidFileDescriptorSource = FileDescriptorSource.fromString("invalid", "invalid");
+        try {
+            ctx.registerProtoFiles(invalidFileDescriptorSource);
+        } catch (DescriptorParserException ex) {
+            // Successfully throw exception
+            return;
+        }
+
+        fail("Failed to process invalid proto file");
```
danielezonca: What about move this `fail` statement just after `ctx.registerProtoFiles(invalidFileDescriptorSource)`? You can remove `return` from`catch` block in this case and I think it is a bit more readable.
Wdyt?

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/index/IndexManager.java
```diff
@@ -0,0 +1,170 @@
...
+import org.bson.Document;
+import org.kie.kogito.persistence.api.schema.EntityIndexDescriptor;
+import org.kie.kogito.persistence.api.schema.IndexDescriptor;
+import org.kie.kogito.persistence.api.schema.SchemaRegisteredEvent;
+import org.kie.kogito.persistence.api.schema.SchemaRegistrationException;
+
+@ApplicationScoped
+public class IndexManager {
+
+    private static long MAX_INDEX_NUMBER = 63L;
```
danielezonca: Is the limit `63` or `64`? Now with this value the condition
```java
        if (MAX_INDEX_NUMBER < indexNameMap.size()) {
```
will fail with `64` fields

danielezonca (commented):
Thanks for your PR, well done :+1: 
Added some comments.
Can you also please update the [README](https://github.com/kiegroup/kogito-apps/blob/master/persistence-commons/README.md) of the common module to explain how to use the mongodb impl? (feel free to create a different readme in mongodb module if you prefer)

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/AbstractStorage.java
danielezonca: I think we want to use same [Java Map](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#put-K-V-) semantic for `put` so it is correct to return previous value

xieshenzh:
Assembly:
#293
https://github.com/kiegroup/kogito-runtimes/pull/591

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/utils/QueryUtils.java
xieshenzh: Changed to user MongoDB client. Thanks.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/AbstractStorage.java
xieshenzh: I think the infinispan implementation supports multiple listeners, that's why it is named  `addObjectCreatedListener`.

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
xieshenzh: Moved to kogito-runtime bom. Created PR https://github.com/kiegroup/kogito-runtimes/pull/591 .

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
xieshenzh: Removed. Thanks.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/DomainEntity.java
xieshenzh: MongoDB converts the `id` attribute to `_id` as the primary key. This converts it back.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
xieshenzh: Added a dummy `_id` here to avoid MongoDB converting the `id` attribute to `_id` to create primary key. I don't think it is necessary to create primary keys for the embedded documents.

persistence-commons/persistence-commons-protobuf/src/test/resources/test.proto
xieshenzh: Too simplify the tests and avoid creating too many indexes.

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/query/DomainQueryTest.java
xieshenzh: This is used to create desired timestamps for the tests. I have removed the `sleep` and explicitly specify the time interval instead.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/MongoDBStorageService.java
xieshenzh: @r00ta @danielezonca I have changed the implementation of `StorageService` for MongoDB as suggested. 
Actually, I don't think the collections are created here. The collections are created when data insertion or index creation takes place.
Please take a look. Thanks.

persistence-commons/persistence-commons-mongodb/src/test/java/org/kie/kogito/persistence/mongodb/MongoDBServerTestResource.java
xieshenzh: I have changed all the tests that use the test resource to failsafe integration tests.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/storage/MongoDBStorageFactoryImpl.java
xieshenzh: Moved all the MongoDB `Storage` implementation to one class [MongoStorage](https://github.com/kiegroup/kogito-apps/pull/293/files#diff-807233895ce75d2e8848ee063e025c95). Thanks.

persistence-commons/persistence-commons-protobuf/src/test/java/org/kie/kogito/persistence/protobuf/ProtoIndexParserTest.java
xieshenzh: Moved. Thanks.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/index/IndexManager.java
xieshenzh: Because there is a default index on the primary key `_id`. There are only 63 left to be created.

xieshenzh:
@danielezonca Added README in the `data-index-storage-mongodb` module: https://github.com/kiegroup/kogito-apps/pull/293/files#diff-b97258a0e20c82bc1fabedd08bcd1d2b

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <artifactId>data-index-storage</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>data-index-storage-mongodb</artifactId>
+  <name>Kogito :: Data Index Storage MongoDB</name>
+
+  <properties>
+    <container.image.mongodb>mongo:4.0</container.image.mongodb>
```
cristianonicolai: could you use https://www.testcontainers.org/modules/databases/mongodb/ instead of a generic container?

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/JobEntity.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        JobEntity jobEntity = (JobEntity) o;
+        return Objects.equals(id, jobEntity.id) &&
```
cristianonicolai: @xiezhang7 I believe id should be enough to identify the same job.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
```diff
@@ -0,0 +1,171 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ProcessInstanceEntity that = (ProcessInstanceEntity) o;
+        return Objects.equals(id, that.id) &&
```
cristianonicolai: same here, id should be enough to identify the same process instance

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
cristianonicolai: @xiezhang7 wouldnt mongodb create the _id anyway and assign a primary key for that? If thats the case then we can simple resue the id attribute that we already provide.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
```diff
@@ -0,0 +1,171 @@
...
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            NodeInstanceEntity that = (NodeInstanceEntity) o;
+            return Objects.equals(_id, that._id) &&
```
cristianonicolai: _id or id should be enough to identify the node instance.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
```diff
@@ -0,0 +1,171 @@
...
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(_id, id, name, nodeId, type, enter, exit, definitionId);
+        }
+    }
+
+    public static class ProcessInstanceErrorEntity {
+
+        public String _id;
```
cristianonicolai: shouldnt this be just an embedded document, do we actually need an id for it?

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/UserTaskInstanceEntity.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        UserTaskInstanceEntity that = (UserTaskInstanceEntity) o;
+        return Objects.equals(id, that.id) &&
```
cristianonicolai: id should be used enough here to identify the same task.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/storage/MongoModelServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+
+@ApplicationScoped
+public class MongoModelServiceImpl implements MongoModelService {
+
+    @Inject
+    Event<IndexCreateOrUpdateEvent> indexCreateOrUpdateEvent;
+
+    Map<String, Supplier<MongoEntityMapper>> entityMapperMap = new ConcurrentHashMap<>();
+
+    {
```
cristianonicolai: could you use the constructor or init method for this?

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/model/JobEntityMapperTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+class JobEntityMapperTest {
+
+    JobEntityMapper jobEntityMapper = new JobEntityMapper();
+
+    Job job;
+
+    JobEntity jobEntity;
+
+    {
```
cristianonicolai: perhpas use a `@before` or `@beforeall` for initialization?

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntityMapperTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+class ProcessInstanceEntityMapperTest {
+
+    ProcessInstanceEntityMapper processInstanceEntityMapper = new ProcessInstanceEntityMapper();
+
+    ProcessInstance processInstance;
+
+    ProcessInstanceEntity processInstanceEntity;
+
+    {
```
cristianonicolai: same here, using a `@before/all`

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/model/UserTaskInstanceEntityMapperTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+class UserTaskInstanceEntityMapperTest {
+
+    UserTaskInstanceEntityMapper userTaskInstanceEntityMapper = new UserTaskInstanceEntityMapper();
+
+    UserTaskInstance userTaskInstance;
+
+    UserTaskInstanceEntity userTaskInstanceEntity;
+
+    {
```
cristianonicolai: same `@before`

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/query/DomainQueryIT.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @AfterEach
+    void tearDown() {
+        storage.clear();
+    }
+
+    @Test
+    void test() {
+        String processInstanceId1 = UUID.randomUUID().toString() + "_process1";
```
cristianonicolai: do we need ` + "_process1"` in here?

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/query/JobQueryIT.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @AfterEach
+    void tearDown() {
+        storage.clear();
+    }
+
+    @Test
+    void test() {
+        String jobId1 = UUID.randomUUID().toString() + "_job1";
```
cristianonicolai: same, do we need `"_job1"` ?

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/query/UserTaskInstanceQueryIT.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @AfterEach
+    void tearDown() {
+        storage.clear();
+    }
+
+    @Test
+    void test() {
+        String taskId1 = UUID.randomUUID().toString() + "_task1";
```
cristianonicolai: same regarding `"_task1"`

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/storage/ProcessInstanceStorageIT.java
```diff
@@ -0,0 +1,62 @@
...
+    void tearDown() {
+        storage.clear();
+    }
+
+    @Test
+    void testCache() {
+        String processInstanceId = UUID.randomUUID().toString();
+        ProcessInstance processInstance1 = TestUtils.createProcessInstance(processInstanceId, RandomStringUtils.randomAlphabetic(5), UUID.randomUUID().toString(), RandomStringUtils.randomAlphabetic(10), ProcessInstanceState.ACTIVE.ordinal(), 0L);
+        ProcessInstance processInstance2 = TestUtils.createProcessInstance(processInstanceId, RandomStringUtils.randomAlphabetic(5), UUID.randomUUID().toString(), RandomStringUtils.randomAlphabetic(10), ProcessInstanceState.COMPLETED.ordinal(), 1000L);
+        StorageTestBase.testStorage(storage, processInstanceId, processInstance1, processInstance2);
```
cristianonicolai: @xiezhang7 perhpas these tests should extend `StorageTestBase` ?

data-index/data-index-storage/pom.xml
```diff
@@ -13,10 +13,27 @@
...
+        <module>data-index-storage-infinispan</module>
+      </modules>
+    </profile>
+    <profile>
+      <id>experimental</id>
+      <modules>
+        <module>data-index-storage-api</module>
+        <module>data-index-storage-protobuf</module>
+        <module>data-index-storage-infinispan</module>
+        <module>data-index-storage-mongodb</module>
```
cristianonicolai: @xiezhang7 I think we should review and simply include this module by default. Otherwise we will have issues to keep it up-to-date if not part of the default build.

persistence-commons/persistence-commons-infinispan/src/main/java/org/kie/kogito/persistence/infinispan/cache/InfinispanCacheManager.java
```diff
@@ -52,22 +53,33 @@
...
+    Provider<RemoteCacheManager> remoteCacheManagerProvider;
+
     RemoteCacheManager manager;
 
     @PostConstruct
     public void init() {
         jsonDataFormat = DataFormat.builder().valueType(MediaType.APPLICATION_JSON).valueMarshaller(marshaller).build();
-        manager.start();
+        // Not initialize RemoteCacheManager if not using infinispan to avoid connection error
+        if (INFINISPAN_STORAGE.equals(storageType)) {
```
cristianonicolai: @xiezhang7 I guess the `InfinispanCacheManager` should not be initialized at all if not in use?

persistence-commons/persistence-commons-mongodb/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+    <groupId>org.kie.kogito</groupId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>persistence-commons-mongodb</artifactId>
+  <name>Kogito :: Persistence Commons MongoDB</name>
+
+  <properties>
+    <container.image.mongodb>mongo:4.0</container.image.mongodb>
```
cristianonicolai: suggestion to use mongodb testcontainer instead.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/query/QueryUtils.java
```diff
@@ -0,0 +1,85 @@
...
+    private QueryUtils() {
+    }
+
+    static Optional<Bson> generateQuery(List<AttributeFilter<?>> filters, Function<String, String> filterAttributeFunction) {
+        return Optional.ofNullable(filters).filter(f -> !f.isEmpty()).map(fs -> and(fs.stream().map(f -> generateSingleQuery(f, filterAttributeFunction)).collect(toList())));
+    }
+
+    static Bson generateSingleQuery(AttributeFilter<?> filter, Function<String, String> filterAttributeFunction) {
+        switch (filter.getCondition()) {
+            case CONTAINS:
```
cristianonicolai, line 50: contains in here means an element in the array, perhaps look into using `elemMatch` ?

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/query/QueryUtils.java
```diff
@@ -0,0 +1,85 @@
...
+public class QueryUtils {
+
+    private QueryUtils() {
+    }
+
+    static Optional<Bson> generateQuery(List<AttributeFilter<?>> filters, Function<String, String> filterAttributeFunction) {
+        return Optional.ofNullable(filters).filter(f -> !f.isEmpty()).map(fs -> and(fs.stream().map(f -> generateSingleQuery(f, filterAttributeFunction)).collect(toList())));
+    }
+
+    static Bson generateSingleQuery(AttributeFilter<?> filter, Function<String, String> filterAttributeFunction) {
```
cristianonicolai:
```suggestion
    static Bson generateSingleQuery(AttributeFilter<?> filter, Function<String, String> filterFunction) {
```

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/MongoStorage.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public MongoStorage(MongoCollection<E> mongoCollection, String rootType, MongoEntityMapper<K, V, E> mongoEntityMapper) {
+        this.mongoCollection = mongoCollection;
+        this.rootType = rootType;
+        this.mongoEntityMapper = mongoEntityMapper;
+    }
+
+    @Override
+    public void addObjectCreatedListener(Consumer<V> consumer) {
+        this.objectCreatedListener = consumer;
```
cristianonicolai: it should be able to add a list of listeners.
Also, take a look at https://mongodb.github.io/mongo-java-driver/4.0/driver/tutorials/change-streams/
We should be able to get changes done even from another runtime instance.

persistence-commons/pom.xml
```diff
@@ -13,12 +13,27 @@
...
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <modules>
+        <module>persistence-commons-api</module>
+        <module>persistence-commons-protobuf</module>
+        <module>persistence-commons-infinispan</module>
+      </modules>
+    </profile>
+    <profile>
+      <id>experimental</id>
```
cristianonicolai: please include in the main profile

cristianonicolai (commented):
@xiezhang7 would be good to include the data index it tests where infinipan is used, using the mongodb storage. So we would have two pass, one for infinispan and another one for mongo. wdyt?

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/storage/MongoModelServiceImpl.java
danielezonca: Actually I would suggest to make `entityMapperMap` as `static` and same this block, it seems a readonly static map

data-index/data-index-storage/data-index-storage-mongodb/pom.xml
xieshenzh: Changed to use the testcontainers MongoDB module.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/JobEntity.java
xieshenzh: Updated

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/storage/MongoModelServiceImpl.java
xieshenzh: Changed to static

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/model/JobEntityMapperTest.java
xieshenzh: Updated

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/query/DomainQueryIT.java
xieshenzh: Removed and updated tests accordingly

data-index/data-index-storage/data-index-storage-mongodb/src/test/java/org/kie/kogito/index/mongodb/storage/ProcessInstanceStorageIT.java
xieshenzh: Updated

data-index/data-index-storage/pom.xml
xieshenzh: Updated

persistence-commons/persistence-commons-infinispan/src/main/java/org/kie/kogito/persistence/infinispan/cache/InfinispanCacheManager.java
xieshenzh: It is not initialized. But `InfinispanCacheManager` observes `ShutdownEvent`. So it will be created when quarkus terminates and fires `ShutdownEvent`.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/query/QueryUtils.java
xieshenzh: Updated

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/MongoStorage.java
xieshenzh: Changed to use change-streams.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
xieshenzh: Removed it and MongoDB won't generate id for it.

data-index/data-index-storage/data-index-storage-mongodb/src/main/java/org/kie/kogito/index/mongodb/model/ProcessInstanceEntity.java
xieshenzh: MongoDB won't create the `_id` if there is a dummy one. 
I have removed the `_id` and `id` will be converted to `_id`.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/query/QueryUtils.java
xieshenzh: According to this mongodb [tutorial](https://docs.mongodb.com/manual/tutorial/query-arrays/#query-an-array-for-an-element), `eq` is used to query if the array field contains at least one element with the specified value.

xieshenzh:
@cristianonicolai I have extended the infinispan IT tests to run with MongoDB. Please take a look. Thanks.

persistence-commons/persistence-commons-infinispan/src/main/java/org/kie/kogito/persistence/infinispan/cache/InfinispanCacheManager.java
xieshenzh: Moved the `observes` method to another [class](https://github.com/kiegroup/kogito-apps/pull/293/files#diff-781891526489ca36f965e35bfb30ef1c), so that `InfinispanCacheManager` won't be initialized at all.

danielezonca (approved):
Thanks for the PR, well done! Approving :+1:

data-index/data-index-service/src/test/java/org/kie/kogito/index/cache/AbstractStorageIT.java
```diff
@@ -78,10 +77,23 @@ public void testObjectRemovedListener() throws Exception {
...
         assertThat(id).isEqualTo(processInstanceId);
     }
+
+    @AfterEach
+    void tearDown() {
+        Storage<String, ProcessInstance> cache = cacheService.getProcessInstancesCache();
+        cache.clear();
+    }
+
+    static void awaitForListenerReady() {
```
cristianonicolai: @xiezhang7 shouldn't this be handled by the `addObjectCreatedListener` method so it only returns once `subscription.request` is called?

data-index/data-index-service/src/test/resources/deals-mongo.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto2";
```
cristianonicolai, line 1: Could we simply reuse the same proto files for both tests?

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private StorageUtils() {
+    }
+
+    private static MongoClient mongoClient;
+
+    static {
+        String mongoConnectionProperty = "quarkus.mongodb.connection-string";
+        String connection = getConfig().getValue(mongoConnectionProperty, String.class);
+        mongoClient = MongoClients.create(new ConnectionString(connection));
```
cristianonicolai: @xiezhang7 shouldn't we reuse the `MongoClient` created by Quarkus instead? You should be able to retrieve that using injection.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public static <E> MongoCollection<E> getCollection(String collection, Class<E> type) {
+        return MongoOperations.mongoDatabase(type).getCollection(collection, type);
```
cristianonicolai: so it seems that we're creating multiple databases per type. Could we simplify that so there is only one database with multiple collections?

persistence-commons/persistence-commons-mongodb/src/test/java/org/kie/kogito/persistence/mongodb/MongoServerTestResource.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final String MONGODB_CONNECTION_PROPERTY = "quarkus.mongodb.connection-string";
+
+    private MongoDBContainer mongoDBContainer;
+
+    @Override
+    public Map<String, String> start() {
+        mongoDBContainer = new MongoDBContainer();
+        mongoDBContainer.start();
+        addProperty(MONGODB_CONNECTION_PROPERTY, mongoDBContainer.getReplicaSetUrl());
```
cristianonicolai: @xiezhang7 should be enough to override this property using `System.setProperty` that way I think you can avoid creating the `MongoConfigSource`

cristianonicolai (commented):
@xiezhang7 I got a doubt regarding the use of Panache. As stated in [here](https://quarkus.io/guides/mongodb-panache#transactions), Quarkus does not support transactions for Panache. Although they are not in use now, that could be an issue in the future. So the question is how much would we really lose if we switch to simple MongoClient with perhaps [pojo codec](https://quarkus.io/guides/mongodb#the-pojo-codec). wdyt?

data-index/data-index-service/src/test/java/org/kie/kogito/index/cache/AbstractStorageIT.java
xieshenzh: Moved to [addObjectCreatedListener](https://github.com/kiegroup/kogito-apps/blob/6ce2eafe26807cbc43c1b74424fcef3c6e0578c5/persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java#L50) .

data-index/data-index-service/src/test/resources/deals-mongo.proto
xieshenzh: The proto files for infinispan tests have more than 64 index annotations. Mongodb cannot create so many indexes.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java
xieshenzh: Changed to use [mongo client](https://github.com/kiegroup/kogito-apps/blob/6ce2eafe26807cbc43c1b74424fcef3c6e0578c5/persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/client/MongoClientManager.java#L67) .

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java
xieshenzh: Changed to explicitly [use one database](https://github.com/kiegroup/kogito-apps/blob/6ce2eafe26807cbc43c1b74424fcef3c6e0578c5/persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/client/MongoClientManager.java#L63) .

persistence-commons/persistence-commons-mongodb/src/test/java/org/kie/kogito/persistence/mongodb/MongoServerTestResource.java
xieshenzh: Done. Removed `MongoConfigSource` .

xieshenzh:
> @xiezhang7 I got a doubt regarding the use of Panache. As stated in [here](https://quarkus.io/guides/mongodb-panache#transactions), Quarkus does not support transactions for Panache. Although they are not in use now, that could be an issue in the future. So the question is how much would we really lose if we switch to simple MongoClient with perhaps [pojo codec](https://quarkus.io/guides/mongodb#the-pojo-codec). wdyt?

@cristianonicolai Panache is built on top of mongodb client, so mongodb client can do everything panache supports.
Removed all the dependencies on panache, [mongodb client](https://github.com/kiegroup/kogito-apps/blob/6ce2eafe26807cbc43c1b74424fcef3c6e0578c5/persistence-commons/persistence-commons-mongodb/pom.xml#L22) is used directly.  
Please take a look. Thanks.

data-index/data-index-storage/data-index-storage-mongodb/src/test/resources/application.properties
```diff
@@ -0,0 +1,15 @@
+# Quarkus
+quarkus.log.console.enable=true
+quarkus.log.console.level=INFO
+quarkus.log.level=INFO
+quarkus.log.category."org.kie.kogito".level=DEBUG
```
cristianonicolai: maybe just comment out this, otherwise it can be quite verbose on Jenkins build.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public static <V, E> void watchCollection(MongoCollection<E> reactiveMongoCollection, Bson operationType,
+                                              BiConsumer<String, V> consumer, MongoEntityMapper<V, E> mongoEntityMapper) {
+        reactiveMongoCollection.watch(singletonList(match(operationType)))
+                .fullDocument(UPDATE_LOOKUP).subscribe(new ObjectListenerSubscriber<>(consumer, mongoEntityMapper));
+
+        // There is no way to check if MongoDB Change Stream is ready https://jira.mongodb.org/browse/NODE-2247
+        // Pause the execution to wait for the Change Stream to be ready
+        try {
+            TimeUnit.MILLISECONDS.sleep(1500L);
```
cristianonicolai, line 53: @xiezhang7 couldnt this be avoided if you add a `CountDownLatch` that called once `onSubscribe` is called?

persistence-commons/persistence-commons-mongodb/src/test/resources/application.properties
```diff
@@ -0,0 +1,15 @@
+# Quarkus
+quarkus.log.console.enable=true
+quarkus.log.console.level=INFO
+quarkus.log.level=INFO
+quarkus.log.category."org.kie.kogito".level=DEBUG
```
cristianonicolai: same here, please comment out so logs arent so verbose in Jenkis.

cristianonicolai:
@xiezhang7 taking a look in the reports from sonar, maybe now that Panache is not in use, we should return the *Entity to use getters and setters instead of public acessors?

data-index/data-index-storage/data-index-storage-mongodb/src/test/resources/application.properties
xieshenzh: Commented out.

persistence-commons/persistence-commons-mongodb/src/test/resources/application.properties
xieshenzh: Commented out.

persistence-commons/persistence-commons-mongodb/src/main/java/org/kie/kogito/persistence/mongodb/storage/StorageUtils.java
xieshenzh: I have tried that. But even after `onSubscribe` is called, the `ChangeStream` is not ready.

xieshenzh:
> @xiezhang7 taking a look in the reports from sonar, maybe now that Panache is not in use, we should return the *Entity to use getters and setters instead of public acessors?

@cristianonicolai replaced with getters and setters. Thanks.

cristianonicolai:
jenkins please retest this

cristianonicolai (approved):
@xiezhang7 great work!
one minor comment would be to try to get Sonar check to pass. Mostly I think is regarding removing public modifier from tests.

xieshenzh:
> @xiezhang7 great work!
> one minor comment would be to try to get Sonar check to pass. Mostly I think is regarding removing public modifier from tests.

@cristianonicolai Thanks. I have fixed most of the sonar reported issues. The remaining test classes with public modifiers are abstract classes. And most of the sonar reported uncovered new code are covered by integration tests.

ricardozanini:
This change broke the operator. Is there any way to make Infinispan default instead?

EDIT: nevermind. See: https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/Data.20index.20support.20for.20MonfoDB

PR #310, opened 2020-07-05
KOGITO-2464 Process list show load more buttons when no results found

nmirasch:
https://issues.redhat.com/browse/KOGITO-2464

Adds setLimit call on first query used to control how many items were loaded last time and to know if load more buttons have to be displayed. also set initial offset value to 0, to make pagination work propertly


AjayJagan (approved):
looks fine to me :)

cristianonicolai (approved):

PR #329, opened 2020-07-20
KOGITO-2811 Skip UI tests when running Maven with -DskipTests

cristianonicolai:
Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [ ] You have read the [contributors guide](https://github.com/kiegroup/kogito-runtimes#contributing-to-kogito)
- [ ] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [ ] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

PR #330, opened 2020-07-20
KOGITO-2775 Add trustyAI audit overview section

kelvah:
Link to jira ticket: [KOGITO-2775](https://issues.redhat.com/browse/KOGITO-2775)

This PR adds the code for the audit overview section of trustyAI.


ui-packages/packages/trusty/api-mock/mocks/inputData.js
```diff
@@ -0,0 +1,282 @@
+const inputData = [
```
jstastny-cz, line 1: indent level is 2 spaces in the rest of js files, please keep it that way also here.

ui-packages/packages/trusty/api-mock/mocks/modelData.js
```diff
@@ -0,0 +1,303 @@
+const executionIdBase = require("./executionIdBase");
+const DMN_1 = `
```
jstastny-cz, line 2: how many of these will be there? would it be worth introducing a 'reader' method forgetting the DMN contents from separate files and keeping those separately?

ui-packages/packages/trusty/package.json
```diff
@@ -4,20 +4,28 @@
...
-    "node-sass": "4.14.1"
+    "@testing-library/react": "^10.4.7",
+    "@testing-library/react-hooks": "^3.3.0",
+    "date-fns": "^2.14.0",
+    "flatpickr": "^4.6.3",
+    "sass": "^1.26.10",
+    "sass-loader": "^9.0.2"
   },
   "scripts": {
+    "build:prod": "yarn run lint && webpack --config webpack.prod.js && yarn run test:coverage",
```
jstastny-cz: please adapt to #329

ui-packages/packages/trusty/package.json
```diff
@@ -4,20 +4,28 @@
...
+    "lint": "tslint -c ./tslint.json --project . './src/**/*.ts{,x}'",
+    "mock-server": "json-server --watch api-mock/db.js --routes api-mock/routes.json --port 1336 --delay 1000 --id executionId --middlewares api-mock/filterSingular.js",
     "start": "webpack-dev-server --hot --color --progress --info=true --config webpack.dev.js",
-    "build:prod": "yarn run lint && webpack --config webpack.prod.js",
-    "lint": "tslint -c ./tslint.json --project . './src/**/*.ts{,x}'"
+    "test": "jest --runInBand --ci --reporters=default --reporters=jest-junit",
+    "test:coverage": "yarn run test --coverage",
+    "test:report": "yarn run test --ci --reporters=jest-junit",
+    "test:watch": "yarn run test --watch"
```
jstastny-cz: what's the plan for UI e2e testing? to keep things in sync I'd argue for cypress - integration can be added in a separate PR.

ui-packages/packages/trusty/src/components/Atoms/ExecutionStatus/ExecutionStatus.tsx
```diff
@@ -0,0 +1,30 @@
+import { CheckCircleIcon, ErrorCircleOIcon } from '@patternfly/react-icons';
+import React from 'react';
+import './ExecutionStatus.scss';
+
+type ExecutionStatusProps = {
+  result: boolean;
+};
+
+const ExecutionStatus = (props: ExecutionStatusProps) => {
```
jstastny-cz, line 9: Please reach to me whenever you have time regarding the UI testing and writing testable components.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { TooltipProps } from '@patternfly/react-core/dist/js/components/Tooltip/Tooltip';
+
+type FormattedDateProps = {
+  date: string;
+  preposition?: boolean;
+  position?: TooltipProps['position'];
+  fullDateAndTime?: boolean;
+};
+
+const FormattedDate = (props: FormattedDateProps) => {
```
jstastny-cz, line 13: this is new to me, I've usually seen
```
const FormattedDate: React.FC<FormattedDateProps> = ({date, preposition, position, fullDateAndTime}) => {

}
```
which I find more organized, is there some advantage of the approach here?

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
```diff
@@ -0,0 +1,48 @@
...
+    preposition = false,
+    position = 'auto',
+    fullDateAndTime = false
+  } = props;
+  const difference = differenceInDays(new Date(date), new Date());
+  const fullFormattedDate = format(new Date(date), 'PPpp');
+  let formattedDate;
+
+  if (fullDateAndTime) {
+    return <span>{fullFormattedDate}</span>;
```
jstastny-cz: please try to refrain from having multiple return statements.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/tests/FormattedDate.test.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import FormattedDate from '../FormattedDate';
+import { format } from 'date-fns';
+
+describe('FormattedDate', () => {
```
jstastny-cz, line 6: no snapshot for the component?

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useEffect, useRef } from 'react';
+import { InputGroup, TextInput } from '@patternfly/react-core';
+import flatpickr from 'flatpickr';
+import { Instance as flatpickrInstance } from 'flatpickr/dist/types/instance';
+import './DatePicker.scss';
+
+flatpickr.l10ns.en.weekdays.shorthand.forEach((day, index, daysArray) => {
```
jstastny-cz: I'd argue for keeping the top part of the component definition for props and component definition, can you move this either below or elsewhere?

ui-packages/packages/trusty/src/components/Molecules/skeletonRows/skeletonRows.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { IRow } from '@patternfly/react-table';
+import SkeletonStripe from './../../Atoms/SkeletonStripe/SkeletonStripe';
+
+/*
+ * Based on a number of rows and columns, this function creates an array specifically intended
+ * for the Patternfly Table component. Feeding the skeletons array to the table rows prop,
+ * it will produce animated stripes to be displayed while loading real data
+ * */
+
+const skeletonRows = (
```
jstastny-cz: can't this be rather a component? otherwise I would expect for the file not to be placed under components (at least not having a separate skeletonRows folder there)

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+interface AuditToolbarBaseProps {
+  page: number;
+  pageSize: number;
+  setPage: (page: number) => void;
+  setPageSize: (pageSize: number) => void;
+  total: number;
+}
+
+interface AuditToolbarCompleteProps extends AuditToolbarBaseProps {
```
jstastny-cz: previously the props were `type`, now `interface`, is there anything preventing us from unifying?

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/tests/AuditToolbarTop.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const onRefresh = jest.fn();
+    const wrapper = renderAuditToolbarTop('mount', { onRefresh });
+
+    wrapper.find('button#executions-refresh').simulate('click');
+
+    expect(onRefresh).toBeCalledTimes(1);
+  });
+});
+
+const renderAuditToolbarTop = (method: 'shallow' | 'mount', props?: object) => {
```
jstastny-cz: you could instead use
```
const wrapper = mount(<AuditToolbarTop {...{...defaultProps, onRefresh}}/>);
```
and get rid of the `renderAuditToolbarTop`

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+              </EmptyState>
+            </Bullseye>
+          )
+        }
+      ]
+    }
+  ];
+};
+
+const ExecutionTable = (props: ExecutionTableProps) => {
```
jstastny-cz: move component definition to the top of the file

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const { data } = props;
+  const columns = ['ID', 'Description', 'Executor', 'Date', 'Execution Status'];
+
+  return (
+    <>
+      {(data.status === 'LOADING' || data.status === 'NOT_ASKED') && (
```
jstastny-cz: I'd argue to have just a single JSX Table element here and just prepare the columns and rows variables that would be passed there.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/tests/ExecutionTable.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+            executionSucceeded: true,
+            executorName: 'testUser',
+            executedModelName: 'LoanEligibility',
+            executionType: 'DECISION'
+          }
+        ]
+      }
+    };
+    const wrapper = shallow(
+      <ExecutionTable data={data as RemoteData<Error, IExecutions>} />
```
jstastny-cz: can we set the type to the `const data` declaration already?

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/tests/__snapshots__/ExecutionTable.test.tsx.snap
```diff
@@ -0,0 +1,834 @@
...
+              />,
+            },
+          ],
+          "decisionKey": "skeleton-9",
+        },
+      ]
+    }
+  >
+    <TableHeader />
+    <TableBody
```
jstastny-cz: it actually does not render the skeletons, so mount might be needed instead of shallow.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/AuditOverview.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <>
+      <PageSection variant={PageSectionVariants.light}>
+        <TextContent>
+          <Title size="3xl" headingLevel="h2">
+            Audit Investigation
+          </Title>
+          <Text component="p">
+            Here you can retrieve all the available information about past cases
```
jstastny-cz: probably should end with period.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/tests/__snapshots__/AuditOverview.test.tsx.snap
```diff
@@ -0,0 +1,68 @@
...
+        component="p"
+      >
+        Here you can retrieve all the available information about past cases
+      </Text>
+    </TextContent>
+  </PageSection>
+  <PageSection
+    isFilled={true}
+  >
+    <Component
```
jstastny-cz, line 25: consider if the snapshot gives sufficient level of detail, if not use mount instead.

ui-packages/packages/trusty/api-mock/mocks/modelData.js
manstis: There will be no more. They are mocked data for when the UI is ran against a mock backend, `api-mock`.

The real service will exist in `trusty-service` and is part of https://issues.redhat.com/browse/FAI-215.

I don't think there's much value added in improving the mock endpoint here.

ui-packages/packages/trusty/api-mock/mocks/inputData.js
kelvah: Done. I forgot to run Prettier on the `api-mock` folder.

ui-packages/packages/trusty/package.json
kelvah: Done.

ui-packages/packages/trusty/api-mock/mocks/modelData.js
kelvah: I agree with @manstis

ui-packages/packages/trusty/package.json
kelvah: I am not sure there are already plans. I suppose it will be cypress too. I agree to add it with a separate PR.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/tests/ExecutionTable.test.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/AuditOverview.tsx
kelvah: Ops. Done.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
kelvah: Fixed to use a single return.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/tests/FormattedDate.test.tsx
kelvah: Snapshots added.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
kelvah: I agree. Moved them to the bottom.

ui-packages/packages/trusty/src/components/Molecules/skeletonRows/skeletonRows.tsx
kelvah: I was unsure about where to place it. Unfortunately it cannot be a component, given how PatternFly Table handles rows content. I moved it to the `utils` folder.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
kelvah: Switched to types for consistency.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/tests/AuditToolbarTop.test.tsx
kelvah: Done. I am used to avoid using nested spread because it could get difficult to read. This example is simple enough I think.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
kelvah: I think there are no obvious advantages from using `React.FC`. It does type checking for `propTypes` and `defaultProps`, which we hardly use with typescript. It has an implicit definition for `children` (with an [issue](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33006)), but it's probably clearer to explicitly declare `children` anyway when a component relies on it.
There's no much difference in the end when using or not using `React.FC`.
Regarding destructuring props inside the signature, it's just a personal preference. I find it more readable to have it inside the body.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/tests/__snapshots__/ExecutionTable.test.tsx.snap
kelvah: I thought it was enough to be sure of setting up the table with the right data configuration.
I avoided `mount` because we would end up rendering the whole PatternFly Table, which could be out of scope for this test. The snapshot would also contain dozens of thousands of lines, that are difficult to review when something changes.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/tests/__snapshots__/AuditOverview.test.tsx.snap
kelvah: I made the same considerations of the `ExecutionTable` test.

ui-packages/packages/trusty/src/components/Atoms/ExecutionStatus/ExecutionStatus.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { CheckCircleIcon, ErrorCircleOIcon } from '@patternfly/react-icons';
+import './ExecutionStatus.scss';
+
+type ExecutionStatusProps = {
+  result: boolean;
```
tiagobento: You can change this type do `"success" | "failure"`, for example. `<ExecutationStatus result={"success"} />` seems more semantic than `<ExecutationStatus result={true} />`

ui-packages/packages/trusty/src/components/Atoms/ExecutionStatus/ExecutionStatus.tsx
```diff
@@ -0,0 +1,32 @@
...
+    icon = <ErrorCircleOIcon className={className} />;
+  }
+
+  return (
+    <>
+      {icon}
+      <span>{statusDescription}</span>
+    </>
+  );
+};
```
tiagobento:
```suggestion
  return (
    <>
      {result === "success" && <>
         <CheckCircleIcon className={'execution-status-badge execution-status-badge--success'} />;
         <span>Completed</span>
      </>}
      {result === "error" && <>
         <ErrorCircleOIcon className={'execution-status-badge execution-status-badge--error'} />;
         <span>Error</span>
      </>}
    </>
  );
};
```

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const difference = differenceInDays(new Date(date), Date.now());
+  const fullFormattedDate = format(new Date(date), 'PPpp');
+  let formattedDate;
+
+  if (difference === 0) {
+    formattedDate = `${formatDistanceToNowStrict(new Date(date))} ago`;
+  } else {
+    const prefix = preposition ? 'on ' : '';
+    formattedDate = `${prefix}${format(new Date(date), 'PP')}`;
+  }
```
tiagobento, line 29: This should be wrapped in a `useMemo`.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const FormattedDate = (props: FormattedDateProps) => {
+  const {
+    date,
+    preposition = false,
+    position = 'auto',
+    fullDateAndTime = false
+  } = props;
+  const difference = differenceInDays(new Date(date), Date.now());
+  const fullFormattedDate = format(new Date(date), 'PPpp');
```
tiagobento, line 21: This should be wrapped in a `useMemo`.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/FormattedDate.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+const FormattedDate = (props: FormattedDateProps) => {
+  const {
+    date,
+    preposition = false,
+    position = 'auto',
+    fullDateAndTime = false
+  } = props;
+  const difference = differenceInDays(new Date(date), Date.now());
```
tiagobento, line 20: This should be wrapped in a `useMemo`.

ui-packages/packages/trusty/src/components/Atoms/SkeletonStripe/SkeletonStripe.tsx
```diff
@@ -0,0 +1,27 @@
...
+  isInline?: boolean;
+  size?: 'sm' | 'md' | 'lg';
+};
+
+const SkeletonStripe = (props: SkeletonStripeProps) => {
+  const { isInline = false, size = 'sm', customStyle } = props;
+  const stripeDefaultStyle = {};
+  const stripeStyle = customStyle
+    ? Object.assign(customStyle, stripeDefaultStyle)
+    : stripeDefaultStyle;
```
tiagobento:
```suggestion
  const stripeStyle = useMemo(() => customStyle ?? {}, [customStyle]);
```

ui-packages/packages/trusty/src/components/Atoms/SkeletonStripe/SkeletonStripe.tsx
```diff
@@ -0,0 +1,27 @@
...
+  let cssClasses = 'skeleton__stripe';
+  if (isInline) {
+    cssClasses += ' skeleton__stripe--inline';
+  }
+  if (size !== 'sm') {
+    cssClasses += ` skeleton__stripe--${size}`;
+  }
```
tiagobento: This should be wrapped in a `useMemo`.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,71 @@
...
+const DatePicker = (props: DatePickerProps) => {
+  const { fromDate, minDate, maxDate, value, onDateUpdate, id, label } = props;
+  const datePicker = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    let calendar: flatpickrInstance;
+    const onChange = (selectedDates: Date[], dateStr: string) => {
+      onDateUpdate(dateStr);
+    };
+    if (datePicker && datePicker.current) {
```
tiagobento: `datePicker` can never be falsy (you can see that from its type), you can only check for `datePicker.current`

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,71 @@
...
+  label?: string;
+  minDate?: string;
+  maxDate?: string;
+  onDateUpdate: (date: string) => void;
+  value?: string;
+};
+
+const DatePicker = (props: DatePickerProps) => {
+  const { fromDate, minDate, maxDate, value, onDateUpdate, id, label } = props;
+  const datePicker = useRef<HTMLInputElement>(null);
```
tiagobento:
```suggestion
  const datePicker = useRef<HTMLInputElement>();
```

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export default DatePicker;
+
+flatpickr.l10ns.en.weekdays.shorthand.forEach((day, index, daysArray) => {
+  if (daysArray[index] === 'Thu' || daysArray[index] === 'Th') {
+    daysArray[index] = 'Th';
+  } else {
+    daysArray[index] = daysArray[index].charAt(0);
+  }
+});
```
tiagobento: This looks like an unexpected side effect. Every time this file is imported, this code will run, which is very hard to control and predict. It this change in configuration is important to your app, you should have an explicit call to a method that does that. On your application entry-point, you can call a function that executes this code. This way, it will be easier to understand why and when things were modified.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const leftArrow = `<svg fill="currentColor" height="1em" width="1em" viewBox="0 0 256 512" aria-hidden="true" role="img" style="vertical-align: -0.125em;"><path d="M31.7 239l136-136c9.4-9.4 24.6-9.4 33.9 0l22.6 22.6c9.4 9.4 9.4 24.6 0 33.9L127.9 256l96.4 96.4c9.4 9.4 9.4 24.6 0 33.9L201.7 409c-9.4 9.4-24.6 9.4-33.9 0l-136-136c-9.5-9.4-9.5-24.6-.1-34z" transform=""/></svg>`;
+const rightArrow = `<svg fill="currentColor" height="1em" width="1em" viewBox="0 0 256 512" aria-hidden="true" role="img" style="vertical-align: -0.125em;"><path d="M224.3 273l-136 136c-9.4 9.4-24.6 9.4-33.9 0l-22.6-22.6c-9.4-9.4-9.4-24.6 0-33.9l96.4-96.4-96.4-96.4c-9.4-9.4-9.4-24.6 0-33.9L54.3 103c9.4-9.4 24.6-9.4 33.9 0l136 136c9.5 9.4 9.5 24.6.1 34z" transform=""/></svg>`;
```
tiagobento: These images should be in their own file. You can use Webpack to load SVGs as strings if you need.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const onSearchSubmit = (): void => {
+    if (searchField && searchField.current) {
+      setSearchString(searchField.current.value);
+    }
+  };
+  const onSearchEnter = (event: React.KeyboardEvent): void => {
+    if (searchField && searchField.current && event.key === 'Enter') {
+      setSearchString(searchField.current.value);
+    }
+  };
```
tiagobento, line 58: This should be wrapped in a `useCallback`.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
```diff
@@ -0,0 +1,152 @@
...
+    setPageSize,
+    onRefresh
+  } = props;
+
+  const searchField = useRef<HTMLInputElement>(null);
+  const onSearchSubmit = (): void => {
+    if (searchField && searchField.current) {
+      setSearchString(searchField.current.value);
+    }
+  };
```
tiagobento, line 53: This should be wrapped in a `useCallback`.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
```diff
@@ -0,0 +1,152 @@
...
+    setToDate,
+    total,
+    pageSize,
+    page,
+    setPage,
+    setPageSize,
+    onRefresh
+  } = props;
+
+  const searchField = useRef<HTMLInputElement>(null);
```
tiagobento:
```suggestion
  const searchField = useRef<HTMLInputElement>();
```

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+import skeletonRows from '../../../utils/skeletonRows/skeletonRows';
+import { IExecution, IExecutions, RemoteData } from '../../../types';
+
+type ExecutionTableProps = {
+  data: RemoteData<Error, IExecutions>;
+};
+
+const ExecutionTable = (props: ExecutionTableProps) => {
+  const { data } = props;
+  const columns = ['ID', 'Description', 'Executor', 'Date', 'Execution Status'];
```
tiagobento, line 23: You can either inline that or wrap it in a `useMemo`.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+      title: <FormattedDate date={item.executionDate} />
+    });
+    cells.push({
+      title: <ExecutionStatus result={item.executionSucceeded} />
+    });
+    row.cells = cells;
+    row.executionKey = 'key-' + item.executionId;
+    rows.push(row);
+  });
+  return rows;
```
tiagobento:
```suggestion
  return rowData.map(item => ({
      executionKey: 'key-' + item.executionId,
      cells: [
        {
          title: (
            <Link to={`/audit/${item.executionType.toLocaleLowerCase()}/${item.executionId}`}>
               {'#' + item.executionId.toUpperCase()}
            </Link>)
        },
        item.executedModelName,
        item.executorName,
        { title: <FormattedDate date={item.executionDate} /> },
        { title: <ExecutionStatus result={item.executionSucceeded} /> }
     ]
    }));
```

ui-packages/packages/trusty/src/components/Organisms/PaginationContainer/PaginationContainer.tsx
```diff
@@ -0,0 +1,42 @@
...
+    page,
+    pageSize,
+    paginationId,
+    onSetPage,
+    onSetPageSize
+  } = props;
+
+  const updatePage = (event: any, pageNumber: number) => {
+    onSetPage(pageNumber);
+  };
```
tiagobento, line 25: This should be wrapped in a `useCallback`.

ui-packages/packages/trusty/src/components/Organisms/PaginationContainer/PaginationContainer.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const updatePageSize = (event: any, pageNumber: number) => {
+    onSetPageSize(pageNumber);
+  };
```
tiagobento, line 28: This should be wrapped in a `useCallback`.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/AuditOverview.tsx
```diff
@@ -0,0 +1,95 @@
...
+    fromDate: string;
+    toDate: string;
+  };
+};
+
+const AuditOverview = (props: AuditOverviewProps) => {
+  const { dateRangePreset } = props;
+  const toPreset = dateRangePreset
+    ? dateRangePreset.toDate
+    : formatISO(new Date(), { representation: 'date' });
```
tiagobento: This should be wrapped in a `useMemo`.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/AuditOverview.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const fromPreset = dateRangePreset
+    ? dateRangePreset.fromDate
+    : formatISO(sub(new Date(), { months: 1 }), { representation: 'date' });
```
tiagobento: This should be wrapped in a `useMemo`.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/useExecutions.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { isCancelledRequest } from '../../../utils/api/httpClient';
+import { RemoteData, IExecutions } from '../../../types';
+
+const useExecutions = (
+  searchString: string,
+  from: string,
+  to: string,
+  limit: number,
+  offset: number
+) => {
```
tiagobento: Since these parameters are all of primitive types, I feel like this custom Hook should have its signature improved by forcing "named parameters". If you create an object with these parameters, users will have to explicitly type the parameter name (which is only the property names of the parameter object). This will increase readability on places using this Hook.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/useExecutions.tsx
```diff
@@ -0,0 +1,43 @@
...
+          setExecutions({ status: 'FAILURE', error });
+        }
+      });
+    return () => {
+      isMounted = false;
+    };
+  }, [searchString, from, to, limit, offset]);
+
+  useEffect(() => {
+    loadExecutions();
```
tiagobento, line 40: You can inline this function call so you don't have to practically repeat the list of dependencies on the `useCallback` statement.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/useExecutions.tsx
```diff
@@ -0,0 +1,43 @@
...
+        }
+      })
+      .catch(error => {
+        if (!isCancelledRequest(error)) {
+          setExecutions({ status: 'FAILURE', error });
+        }
+      });
+    return () => {
+      isMounted = false;
+    };
```
tiagobento, line 36: Nice use of the `isMounted` variable check. If I can push a little bit, maybe you can create a custom Hook for all request types, so that you don't have to repeat this logic for every request.

tiagobento: This is just a suggestion, though. I think it would improve readability of code making requests.

ui-packages/packages/trusty/src/index.css
```diff
@@ -1,18 +1,20 @@
...
+    'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue',
+    sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
 }
 
 code {
-    font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New", monospace;
+  font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
+    monospace;
```
tiagobento, line 12: I see you reformatted the code here without changes. Could it be that you're not using Prettier as the formatter? I see that you have prettier configured for this repository. Please check that.

ui-packages/packages/trusty/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  executedModelName: string;
+  executionType: string;
+  executionSucceeded: boolean;
+  executorName: string;
+}
+
+export interface IExecutions {
+  headers: IExecution[];
+  total: number;
+}
```
tiagobento: Any particular reason for these not be called `Execution` and `Executions`?

ui-packages/packages/trusty/src/utils/api/auditApi.ts
```diff
@@ -0,0 +1,25 @@
...
+    url: EXECUTIONS_PATH,
+    method: 'get',
+    params: {
+      search: searchString,
+      from,
+      to,
+      limit,
+      offset
+    }
+  };
```
tiagobento: You can inline that.

ui-packages/packages/trusty/src/utils/api/httpClient.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (call) {
+    call.cancel('Request superseded');
+  }
+  call = axios.CancelToken.source();
+
+  config.cancelToken = call.token;
+  return httpClient(config);
+};
+
+const isCancelledRequest = axios.isCancel;
```
tiagobento, line 21: I think this function can be inlined. This indirection causes code reading to be more difficult, since its a code of your application, people will expect extra-behaviour. Instead, this is only an "alias" for `axios.isCancel`.

ui-packages/packages/trusty/src/utils/api/httpClient.ts
```diff
@@ -0,0 +1,23 @@
...
+  timeout: 5000,
+  headers: {}
+});
+
+let call: CancelTokenSource;
+
+const callOnce = (config: AxiosRequestConfig) => {
+  if (call) {
+    call.cancel('Request superseded');
+  }
```
tiagobento, line 14:
```suggestion
  call?.cancel('Request superseded');
```

ui-packages/packages/trusty/src/utils/api/httpClient.ts
```diff
@@ -0,0 +1,23 @@
+import axios, { AxiosRequestConfig, CancelTokenSource } from 'axios';
+
+const httpClient = axios.create({
+  baseURL: process.env.KOGITO_TRUSTY_API_HTTP_URL,
+  timeout: 5000,
+  headers: {}
+});
+
+let call: CancelTokenSource;
```
tiagobento, line 9: This is a global variable. We should avoid that at all costs, since they can be dangerous. Can you explain why you need this to be global?

tiagobento (changes requested):
Hi @kelvah! As requested by @porcelli and @ederign, here's my review for this PR.

I noticed some recurring issues regarding the use of hooks inside function components. As a general rule, you want every statement on a function component to be wrapped in a hook (e.g. `useCallback`, `useMemo` etc). This prevents your component from re-creating functions and objects every time its function is called. Using hooks is how React enabled us to memoize values and avoid unnecessary renders. I left some comments inline where I saw those, but please double-check components I didn't comment on to look for places where hooks should be used.

Regarding code formatting, I left a comment inline about Prettier, which I think you're already using to format the code, but just to be sure :) I also noticed that you didn't configure the `printWidth` property on `prettier.config.js`, leaving it as its default 80. Although Prettier discourages changing this setting (see https://prettier.io/docs/en/options.html), specially for JSX, I think we should make an exception. See the comment on this very interesting thread about that: https://github.com/prettier/prettier/issues/4298#issuecomment-383411063. We've been using 120 on kogito-tooling and it's been great. I'm not sure if you guys have been bothered by that or not, but it's something I noticed, so I'm leaving this here as a suggestion :)

Regarding the mocks you created for the API, I'm concerned that using JavaScript sources will make it very hard to maintain in the medium/long term. We're using TypeScript because we believe that statically-typed source code helps us catch bugs during compilation that would only present during runtime otherwise, and I don't see why we should treat testing code differently (actually even more so, since we don't test our test code 😝 ). To help with that, I recommend configuring 'noImplicitAny' and 'noImplicitThis' on your `tsconfig.json` file. (more about that here: https://www.staging-typescript.org/tsconfig#noImplicitAny) and changing the `api-mock` package files into TS files. I know the type system can get in the way sometimes, but I'm convinced that what it does is push us harder to end up with more cohesive and coherent abstractions.

Lastly I noticed that you're sometimes using `enzyme` to shallow-render your components during tests. Although this seems great for "unit testing" components, it encourages practices that are not considered 'the best' during tests, such as directly interacting with props and state. React Testing Library (RTL), on the other hand, does not have a shallow-rendering capability (more here: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering). In kogito-tooling, we decided to never shallow-render components during tests. We've been very happy with it so far, making our tests look much more like UI tests than unit tests (without obviously loosing execution speed, since they remain unit tests).

I hope this has been helpful to you!

kelvah:
Hi @tiagobento, thanks for your review.

You touched many topics. I will try to answer you one by one.

Regarding performance, I tend to disagree with the idea of memoizing everything by general rule. `useMemo` and `useCallback` are abstractions that comes with a cost. Using them on fairly simple operations will probably add no benefits, on the contrary they add some computation and eventually make performance worse. I would consider their usage when facing computational expensive operations or when a little change could cause a re-render on a big components tree.
Don't get me wrong, I stand for a performance driven approach when writing code. But premature optimization could lead to unexpected results. I would try to identify performance bottlenecks first and then apply tools to measurably improve things.

About Prettier. I agree with you on the `printWidth` size (maybe 100 could be fine but that's personal taste). I did not change the default value because Trusty is the last package being added to the repo, and the `ui-packages/.prettierrc` configuration was already there. I did not want to force everyone to reformat their code. I would bring however the topic for discussion because it's a reasonable change.

On the mocked API written with JS. I set that up because when we started there was still no `trusty-service` available. They were meant just to provide a back-end while developing and honestly I did not think it was relevant to rely on TypeScript. Endpoints changed many times in the PoC phase and It was a way to keep up with the pace of the various iterations. They are not used by automated tests, which rely on specific typed mocks. And I suppose they could easily be removed in the long term. By the way I agree on the addition of the `notImplicit` rules. 

Finally testing. I kind of agree with you about `enzyme`. I also have some doubts about the overuse of snapshot testing. Surely they are useful for unit testing of presentational components and fairly small UI pieces. Probably they are not the best when dealing with integration tests. I know about `react-testing-library`, I followed Kent C. Dodds for a while (I think he's one of its contributors). I don't have a clear answer on this topic because it's difficult to have a single general strategy. Often the line between unit and integration tests is blurry on the front-end side, where there are many wrapper and containers components and where encapsulation is the general approach to UI composition. It would be great to have some discussion about it, even maybe produce some agreed general guideline about how tests should be written within our projects, with real world examples etc. It could help putting everyone, including QA, on the same page given that there are many ways (and tools) for dealing with tests and that some of them are obviously opinionated.

tiagobento:
@kelvah Thanks for your reply. I'm glad you're going to bring topics such as `noImplicitThis` and Prettier configuration to your team. I really believe that this will improve the coding experience.

About the tests, yes, I agree with you that there's no general rule on what to make when, but also I think that making every case a special case can have a negative impact. Since different people have different experiences, they tend to do what they're used to and possibly bring more libraries/frameworks to a project where it's not needed. I'm not saying that's the case with your tests, but I believe that RTL is a superset of enzyme regarding component rendering to test, so that's why we opted to only use RTL on kogito-tooling, feel free to talk to your team about it if you think it's worth it. I'd happy to join the discussion!

About "performance", When I talk about `useCallback` and `useMemo` hooks, my first concern is not about performance, but about predictability. In my view, avoiding unnecessary re-renders make a much more stable and future-proof application. Predictability is very important when developing, because once your app grows, many moving parts start interacting with each other, making very hard to track where a referential inequality triggered a new render, for example. If you go back to my PR reviews I'm always talking about premature optimisations, so I agree with you in every way that we should avoid making it. Controversially, I tend to view that *not using* React's memoization mechanisms is what's actually a premature optimisation, since you're giving up predictability in favour of performance. And that's where my "general rule" for memoizing stuff comes from. I'm happy to pay the memoization price where I don't actually need, if it leads my app to have predictable behaviour (i.e. avoiding unnecessary re-renders) at any point in the future. Hope that makes sense to you.

I'll wait for your replies inline :)

kelvah:
Hi @tiagobento,

Sure, let's join efforts on the testing strategy subject. I will surely involve you when we approach discussing it.

Thank you for the explanation on your take on performance. I get what you are aiming to. Let me add some observations though.
Thinking that if you avoid to memoize everything you are doing premature optimization can be viewed as a stretch.
I would not consider re-renders a problem by default. We can agree the logic React uses to trigger re-renders can be obscure sometime and lead to strange application behaviors (your predictability concerns). But that's hardly such a sensitive issue.
Re-renders don't always imply a DOM update, which is the bottleneck to avoid whenever possible. Actually the main selling point of React is its ability to carefully and fastly handle re-renders, updating the DOM only when and where it's strictly necessary. They "invented" the virtual DOM concept just for that and it's the reason why we use React, isn't it?

I suppose you had past experiences that lead you to your conclusions and I agree that things could get difficult to manage with huge applications and deeply nested trees. But I don't see the omnipresent memoization as the cure for that. I consider hooks authors advice to use memoization wisely and only when there's a clear performance improvement as probably the most reasonable approach. Otherwise they'll introduce complexity, less readable code and the other not-so-good things we mentioned.
Well, it seems a long discussion. I surely value exchanging different ideas on such controversial topics though.
Maybe the main advantage of frameworks more opinionated then React is that there's little ambiguity about how things work and a clear imposed take on how to manage UI development. Just a side consideration.

ui-packages/packages/trusty/src/components/Atoms/ExecutionStatus/ExecutionStatus.tsx
kelvah: I agree. Done.

ui-packages/packages/trusty/src/components/Atoms/ExecutionStatus/ExecutionStatus.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Atoms/SkeletonStripe/SkeletonStripe.tsx
kelvah: I am sure there were some default styles before. Forgot to cleanup.
Done without nullish coalescing. The version of TypeScript we use doesn't support it yet. We are gradually updating everything. Now it's the turn of PatternFly. TypeScript could be next. I am considering updating it just for the trusty package to avoid refactoring though.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
kelvah: You have a point. Done.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
kelvah: Moved the configuration in the app initialization. I didn't dwell too much with this datepicker implementation because it was meant to be temporary while waiting for the release of the PatternFly one in June. Unfortunately it was not released like planned.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Organisms/ExecutionTable/ExecutionTable.tsx
kelvah: Done.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/useExecutions.tsx
kelvah: I agree. I was used to some `ESLint` rule that warned you from adding more then 3 parameters to functions. Done.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/useExecutions.tsx
kelvah: I am not sure about it. I am exporting the callback because it's needed for refreshing data. I am afraid I cannot inline it.

ui-packages/packages/trusty/src/components/Templates/AuditOverview/useExecutions.tsx
kelvah: Yes, I've been thinking about it. I am moving ajax calls inside dedicated hooks while migrating the code for the PR's and it could help. I am also thinking about getting rid of the `auditApi.ts` wrappers if every ajax call will be confined in its custom hook. It seems a useless abstraction now.

ui-packages/packages/trusty/src/index.css
kelvah: Probably `index.css` was not formatted on the first PR for some reason. Maybe I did not set Prettier to run on `.css` files from the start. Now it does.

ui-packages/packages/trusty/src/types.ts
kelvah: No particular reason. I was following some common pattern of prefixing interfaces with `I`. Which honestly I don't like. Changed as suggested.

ui-packages/packages/trusty/src/utils/api/auditApi.ts
kelvah: Done.

ui-packages/packages/trusty/src/utils/api/httpClient.ts
kelvah: Cannot do optional chaining with the available version of TypeScript, sadly.

ui-packages/packages/trusty/src/utils/api/httpClient.ts
kelvah: I am afraid I cannot inline it because it depends on the Axios instance. For now it's just an alias. It will eventually be gone with the custom hooks I was referring in the [previous comment](https://github.com/kiegroup/kogito-apps/pull/330#discussion_r463558995).

ui-packages/packages/trusty/src/utils/api/httpClient.ts
kelvah: I agree a module scoped variable is not the best solution. I will rewrite that together with the Axios handling [I mentioned](https://github.com/kiegroup/kogito-apps/pull/330#discussion_r463558995) in the following PR.

ui-packages/packages/trusty/src/utils/api/httpClient.ts
tiagobento: Since you're importing `axios` directly from the library, I think you could simply do that in other places too, but if you already have plans to improve that in the future, I'm ok with that atm :)

ui-packages/packages/trusty/src/utils/api/httpClient.ts
tiagobento: Oh :( Ok.

ui-packages/packages/trusty/src/utils/api/httpClient.ts
tiagobento: Ok, thx for letting me know!

tiagobento:
@kelvah Thanks for the reply! I'll be happy to discuss the testing strategy.

Yes, I think we've both made our points and it's clear that they're both different approaches, but equally valid (at least I think so 😝 ). I appreciate you taking the time to reply me with such detail. Resolved the comments inline and approved. :)

tiagobento (approved):

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/tests/FormattedDate.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  test('displays a relative timestamp if the date is in the last 24h', () => {
+    const fixedDate = '2020-01-01T00:00:00.000Z';
+    jest
+      .spyOn(global.Date, 'now')
+      .mockImplementation(() => new Date(fixedDate).getTime());
+    const wrapper = shallow(<FormattedDate date={fixedDate} />);
+
+    expect(wrapper).toMatchSnapshot();
```
jstastny-cz: these snapshots seem redundant as we're checking the span content below.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,72 @@
...
+      />
+    </InputGroup>
+  );
+};
+
+export default DatePicker;
+
+export const datePickerSetup = () => {
+  flatpickr.l10ns.en.weekdays.shorthand.forEach((day, index, daysArray) => {
+    if (daysArray[index] === 'Thu' || daysArray[index] === 'Th') {
```
jstastny-cz, line 66: Is this a distinction between Tuesday and Thursday? (that's the only possible explanation I could come up with) Shouldn't similar workaround apply to Saturday vs. Sunday?

jstastny-cz (commented):
@kelvah @tiagobento I'd be happy to join your discussion about testing strategy, so that it is in sync across our projects.

ui-packages/packages/trusty/src/components/Atoms/FormattedDate/tests/FormattedDate.test.tsx
kelvah: I added them following [your comment above](https://github.com/kiegroup/kogito-apps/pull/330#discussion_r457985722) :grimacing: 
I agree they are redundant. Should I remove them all?

jstastny-cz: Ah! My bad. I am sorry. Is there any benefit in let's say keeping just a single snapshot check? So that we notice the possible change in how component is rendered, but keep the snapshot file small.

kelvah: Ok, let's keep the first one, just to track eventual PatternFly updates on the tooltip.

ui-packages/packages/trusty/src/components/Molecules/DatePicker/DatePicker.tsx
kelvah: Yes, that's the reason. I don't know how I overlooked the same thing for Sat/Sun. Added disambiguation for that too.

jstastny-cz (approved):

PR #332, opened 2020-07-22
KOGITO-2599: DataTable component API changes

yzhao583:
Jira: https://issues.redhat.com/browse/KOGITO-2599

Create a DataTableColumn interface to represent a column for our DataTable containing:

- path (required): jsonPath to the object property we want to show,  this will be used to get the row cells value.

- label (required): String for the table header

- bodyCellTransformer (optional):  a functions to convert the cell content to a different format for a better visualization. The function should return either a string with the formatted content or either a React Component to be displayed on the Table cell. As parameters it will receive:

     - value: the actual cell value (calculated by the column path attribute)

     - rowDataObj: the full object rendered on the row. 

### DIFF
--- a/ui-packages/yarn.lock
+++ b/ui-packages/yarn.lock
@@ -8247,7 +8247,7 @@ escape-string-regexp@1.0.5, escape-string-regexp@^1.0.2, escape-string-regexp@^1
   resolved "https://registry.yarnpkg.com/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz#1b61c0562190a8dff6ae3bb2cf0200ca130b86d4"
   integrity sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ=
 
-escodegen@^1.11.1:
+escodegen@^1.11.1, escodegen@^1.8.1:
   version "1.14.3"
   resolved "https://registry.yarnpkg.com/escodegen/-/escodegen-1.14.3.tgz#4e7b81fba61581dc97582ed78cab7f0e8d63f503"
   integrity sha512-qFcX0XJkdg+PB3xjZZG/wKSuT1PnQWx57+TVSjIMmILd2yC/6ByYElPwJnslDsuWuSAp4AwJGumarAAmJch5Kw==
@@ -8266,6 +8266,11 @@ eslint-scope@^4.0.3:
   dependencies:
     esrecurse "^4.1.0"
     estraverse "^4.1.1"
+
+esprima@1.2.2:
+  version "1.2.2"
+  resolved "https://registry.yarnpkg.com/esprima/-/esprima-1.2.2.tgz#76a0fd66fcfe154fd292667dc264019750b1657b"
+  integrity sha1-dqD9Zvz+FU/SkmZ9wmQBl1CxZXs=
 
 esprima@^4.0.0, esprima@^4.0.1, esprima@~4.0.0:
   version "4.0.1"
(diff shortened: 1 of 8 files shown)


ui-packages/yarn.lock
```diff
@@ -8277,9 +8277,9 @@ escape-string-regexp@1.0.5, escape-string-regexp@^1.0.2, escape-string-regexp@^1
   resolved "https://registry.yarnpkg.com/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz#1b61c0562190a8dff6ae3bb2cf0200ca130b86d4"
   integrity sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ=
 
-escodegen@^1.11.1:
+escodegen@^1.11.1, escodegen@^1.8.1:
```
pefernan, line 8250: @yzhao583 I think this shouldn't be pushed...

yzhao583: @pefernan Thanks for the review. I just installed the "jsonpath" package, don't know why so many dependencies got updated as well....

Are you suggesting not push local change of "yarn.lock"? Since I did not see it in the ".gitignore", so just want to confirm. Thanks
@cristianonicolai @AjayJagan @Sara4994

jstastny-cz: It is a transitive dependency of jsonpath - this line specifies what versions were required either directly or transitively, wherease the decisive part is the line below - `version "1.14.3"` that's what actually got resolved as matching to all the "requirements" from above. And that version didn't change, so we're still on the same one.

But we should stick to a single registry url in yarn.lock imo - please run
`yarn locktt --registry=https://registry.yarnpkg.com -s`

More concerning is the esprima entry below - there we've added imo a second version of the same dependency.

pefernan: @jstastny-cz yea, I agree with the esprima.

@yzhao583 I know I already asked some changes on the design. But I think that, since the nested properties aren't required for now, we could rollback the that part and remove the jsonpath dep for now. I don't think we waht to have different dep versions.

pefernan (changes requested):
@yzhao583 thank you very much, the code looks well and works perfect. The only issue is the transitive dep. Since that change isn't a blocking requirement I may remove the jsonpath dep, and leave that part as it was before.

ui-packages/yarn.lock
yzhao583: @jstastny-cz @pefernan If the change of "yarn.lock" in this PR is discarded, will the issue be fixed? 
@pefernan I think the most important part of this PR is to support nested properties, if we decide to not have the feature for now, then I do not think it is necessary to have this PR opened.

yzhao583: @pefernan @jstastny-cz I did some research about the issue, looks like we are not the only one to have the pain..and "esprima" is not the only dependency got duplicated. In our case, we have many dependencies got duplicated, such as "esutils", "eventemitter3", "extend-shallow", etc. 

Looks like it is a issue has already been reported to yarn, here are some links to describe the issue:
https://github.com/yarnpkg/yarn/issues/3778
https://medium.com/@scinos/de-duplicating-yarn-lock-ae30be4aa41a

Here are some solutions I found maybe useful:
https://medium.com/@bnaya/yarn-deduplicate-the-hero-we-need-f4497a362128
https://github.com/atlassian/yarn-deduplicate

In conclusion, this is definitely a issue that need our attention, but I would suggest to have a seperated PR to fix this, since it is related to so many dependencies and may need thoroughly test after the fix.

pefernan:
@yzhao583 could you please rebase?

ui-packages/yarn.lock
pefernan: @yzhao583 good point with https://github.com/atlassian/yarn-deduplicate.

I already seen that we are facing this issue of dupplicated deps with different versions

@jstastny-cz @cristianonicolai 
To me I'd say that we should open a JIRA to fix that (. Seems like yarn deduplicate could solve the issue.) and continue with this PR.

pefernan (approved):

ui-packages/yarn.lock
cristianonicolai: +1 to create a JIRA to look into `yarn-deduplicate`. That might actually be related to the issues we see when building on https://github.com/kiegroup/kogito-apps/pull/349. 
I'm really just not sure how there two versions of "esprima" can coexist. @yzhao583 can you maybe that a look if we can safely use both?

kelvah: Hi everyone, I was brought here by @cristianonicolai's PR mention.

I want to reassure you about having multiple version of the same module in the lock file. It's not a big deal, actually that's quite common when the dependency list grows. If a package needs an older version of a module already installed, there's no other way then install it and add it to the lock (the `esprima` example).

It's `yarn` (`npm`) job to install different versions in the right place, nesting `node_modules` inside deps that have specific version requirements, so that they will resolve the correct one.
The downside is an increase in the bundle size. Besides that, there are generally no issues for transitive deps. There are some exceptions though, like with multiple version of `react` for example. But when eventual conflicts arise, they hardly go unnoticed.

Regarding dedupe, its purpose is a bit different. `yarn` already takes care of de-duplication in most of the cases. Some special treatment is required when multiple versions of a specific module are resolved even when only one of them could suffice according to listed versions inside `package.json` . It could happen sometimes, especially for projects with a long lifetime (the order in which deps are added matters) and a high number of deps. Again, I don't think we should worry about it until we face a specific case requiring dedupe.

The only problem I see, already fixed, was the presence of `npm` registry urls inside `yarn.lock`.

Sorry for the intrusion :)

jstastny-cz (approved):

PR #334, opened 2020-07-22
KOGITO-2842 add --skipIntegrity to locktt

jstastny-cz:
Skip removing the integrity fields from yarn.lock.

### DIFF
--- a/ui-packages/pom.xml
+++ b/ui-packages/pom.xml
@@ -113,7 +113,7 @@
                 </goals>
                 <configuration>
                   <npmRegistryURL>${env.NPM_REGISTRY_URL}</npmRegistryURL>
-                  <arguments>run locktt --</arguments>
+                  <arguments>run locktt -- --skipIntegrity</arguments>
                 </configuration>
               </execution>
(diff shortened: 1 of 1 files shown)


jstastny-cz:
#335 alternative approach

ui-packages/pom.xml
```diff
@@ -113,7 +113,7 @@
                 </goals>
                 <configuration>
                   <npmRegistryURL>${env.NPM_REGISTRY_URL}</npmRegistryURL>
-                  <arguments>run locktt --</arguments>
+                  <arguments>run locktt --skipIntegrity --</arguments>
```
kelvah: The correct order for the arguments should be `run locktt -- --skipIntegrity`.

The `--` should [precede](https://docs.npmjs.com/cli/run-script) all the options passed to the script.

kelvah (commented):
I favor this approach instead of #335.
Integrity info should be preserved if possible.

cristianonicolai (approved):

pefernan (approved):

PR #365, opened 2020-08-03
KOGITO-2941: Sonarcloud coverage report is not adding the react part

nmirasch:
Related jira: https://issues.redhat.com/browse/KOGITO-2941
Adds the test coverage when test are executed in ui-packages

### DIFF
--- a/ui-packages/package.json
+++ b/ui-packages/package.json
@@ -11,7 +11,7 @@
     "build": "lerna run build --stream --",
     "build:fast": "lerna run build:fast --stream --",
     "build:prod": "lerna run build:prod --stream --",
-    "test": "lerna run test --stream --",
+    "test": "lerna run test:coverage --stream --",
     "cypress:run": "lerna run cypress:run --",
     "cypress:run-management-console": "cypress run --project ./packages/management-console",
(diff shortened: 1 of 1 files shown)


cristianonicolai:
@nmirasch I though` --coverage` would only print in the console the test coverage, if we actually need this, wouldn't be best to add `--coverage` to the default *test* execution?

nmirasch:
@cristianonicolai I changed the default test script calling each package script including --coverage.
@jstastny-cz, Does this work? do we need that extra script to differentiate between 'test' and 'test with coverage' at ui-package level?

cristianonicolai (approved):

jstastny-cz:
my concerns is that now `yarn test` invoked in top-level will behave differently from `yarn test` invoked in the package (e.g. common). So I would rather override `test` scripts in all packages and from lerna just invoke `test.

jstastny-cz (approved):
Solves our PR coverage reports, approving regardless my nitpick above.

cristianonicolai:
> my concerns is that now `yarn test` invoked in top-level will behave differently from `yarn test` invoked in the package (e.g. common). So I would rather override `test` scripts in all packages and from lerna just invoke `test.

+1 I guess test command should include coverage by default in this case.

PR #408, opened 2020-08-21
KOGITO-3150 Common package OUIA for Organisms

jstastny-cz:
*Goals*
 * Use utils function from OUIAUtils.
 * Introduce OUIA attributes in common package for Organisms components.
 * Fix any Snapshots linked to the change.

*Non goals*
 * OUIA id attributes won't be assigned globally in places of the Component usage.

Note:
* formatting changes forced by the commit-hook, in some cases needed to move also ts-compiler hints to the exact call after a single line was broken into several.


nmirasch (approved):
Looks good to me and Thanks for the format updates!

Sara4994 (approved):

PR #410, opened 2020-08-24
KOGITO-2914: connect Trusty service and Explainability service

kostola:
## General description

This PR includes multiple tickets because the are all correlated to close the end to end loop ([involved component]):
- [KOGITO-3086](https://issues.redhat.com/browse/KOGITO-3086): implement a remote `PredictionProvider` that uses `/predict` API to calculate explainability. To do that, refactor `explainability-core` to support not blocking execution [explainability-addon][explainability-core]
- [KOGITO-2944](https://issues.redhat.com/browse/KOGITO-2944): implement explainability DTOs for input and output and connect them to `explainability-core` [trusty-service][explainbility-service]
- [KOGITO-2945](https://issues.redhat.com/browse/KOGITO-2945): store and read explainability result inside `trusty-service` [trusty-service]
- [KOGITO-2914](https://issues.redhat.com/browse/KOGITO-2914): expose explainability information as REST featureImportance endpoint [trusty-service]

## Description of this PR

This PR involves `trusty-service` and `explainability-service`.
- Implementation of `ExplainabilityRequestDto` and `ExplainabilityResultDto` with all data needed for explainabiltiy
- Big `explainability-core` refactor to make `explain` method async (now `explainAsync`) using `CompletableFuture`. The most important changes are inside `LimeExplainer`. NOTE: test suite is still the same so the behavior has not changed (a part from API changes of course)
- Implementation of `RemotePredictionProvider`
- `ExplainableServiceImpl` is no more mocked and now is properly connected with `explainability-core` library
- `trusty-service` now properly store `ExplainabilityResultDto`
- `trusty-service` now expose explainability values via `ExplainabilityApiV1`

NOTE: SonarCloud test coverage is not considering IT as part of the coverage so most of endpoint classes seem untested but they are

## List of PRs
- https://github.com/kiegroup/kogito-runtimes/pull/720
- https://github.com/kiegroup/kogito-apps/pull/410


explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/ExplainabilityRequestDto.java
```diff
@@ -16,26 +16,50 @@
...
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class ExplainabilityRequestDto {
 
     @JsonProperty("executionId")
     private String executionId;
 
-    public ExplainabilityRequestDto(){
+    @JsonProperty("serviceUrl")
+    private String serviceUrl;
```
r00ta, line 32: I think the URL should not be in the dto (at least for the scenarios we will support in this first release). The URL to be used is going to be injected by the operator with an enviroment variable. wdyt?

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/ExplainabilityResultDto.java
```diff
@@ -16,23 +16,33 @@
...
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class ExplainabilityResultDto {
+
     @JsonProperty("executionId")
     private String executionId;
 
-    // TODO: add properties
+    @JsonProperty("saliency")
+    private Map<String, Map<String, Double>> saliency;
```
r00ta: Create a dedicated object for this to avoid `Map<Map<>>` that is not very informative? imo it's better to encapsulate it in a separated object in general since it will be easier to add properties in the future

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -75,17 +77,23 @@ public static double impactScore(PredictionProvider model, Prediction prediction
         }
 
         PredictionInput predictionInput = new PredictionInput(copy);
-        List<PredictionOutput> predictionOutputs = model.predict(List.of(predictionInput));
-        PredictionOutput predictionOutput = predictionOutputs.get(0);
+        List<PredictionOutput> predictionOutputs;
+        try {
+            predictionOutputs = model.predict(List.of(predictionInput)).get();
+        } catch (InterruptedException | ExecutionException e) {
+            predictionOutputs = Collections.emptyList();
```
r00ta: I'm not sure about this reaction to these exceptions. Let's assume that something bad happens and we catch the exception, we would anyway produce an (kind of empty?) explaination and display it right? If something bad happens, I think we should make it transparent to the user somehow (or even leave the explaination as "not available" or something) and not showing him wrong information

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/LimeStabilityTest.java
```diff
@@ -53,41 +55,53 @@ private void assertStable(PredictionProvider model, List<Feature> featureList) {
...
                 }
-            }
-            assertTrue(topFeature);
+                // check that the topmost important feature is stable
+                List<String> names = new LinkedList<>();
+                saliencies.stream().map(s -> s.getPositiveFeatures(1)).forEach(f -> names.add(f.get(0).getFeature().getName()));
+                Map<String, Long> frequencyMap = names.stream().collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+                boolean topFeature = false;
+                for (Map.Entry<String, Long> entry : frequencyMap.entrySet()) {
+                    if (entry.getValue() >= 0.9) {
```
r00ta: move to `TOP_FEATURE_THRESHOLD` constant or something?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/SampleWeighterTest.java
```diff
@@ -38,41 +38,41 @@ void testSamplingEmptyDataset() {
...
-            trainingSet.add(doubles);
-        }
-        PredictionInput targetInput = new PredictionInput(features);
-        double[] weights = SampleWeighter.getSampleWeights(targetInput, trainingSet);
-        // check that weights decrease with the distance from the 1 vector (the target instance)
-        for (int i = 0; i < weights.length - 1; i++) {
-            assertTrue(weights[i] > weights[i + 1]);
-        }
-    }
+//    @Test
```
r00ta: disable the test instead of commenting?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,162 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ExplanationServiceImpl.class);
+
     @Inject
     ManagedExecutor executor;
 
     @Override
     public CompletionStage<ExplainabilityResultDto> explainAsync(ExplainabilityRequest request) {
-        // TODO: get explainability from expl library https://issues.redhat.com/browse/KOGITO-2920
-        return CompletableFuture.supplyAsync(() -> new ExplainabilityResultDto(request.getExecutionId()), executor);
+        LOG.info("** explainAsync called ***");
```
r00ta: remove this log or make it more informative like `processing explainability for execution id xyz`?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,162 @@
...
+        // TODO: restore limeExplainer when loop works and fix triggered exceptions
+//        RemoteKogitoPredictionProvider provider = new RemoteKogitoPredictionProvider(request, Vertx.vertx(), ThreadContext.builder().build());
+//        LimeExplainer limeExplainer = new LimeExplainer(100, 1);
+        Prediction prediction = getPrediction(request.getInputs(), request.getOutputs());
+        return CompletableFuture
+                // .supplyAsync(() -> limeExplainer.explain(prediction, provider))
+                .supplyAsync(() -> mockedExplainationOf(prediction))
```
r00ta: should't this call the expl lib?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,162 @@
...
+        return new Saliency(
+                output,
+                input.getFeatures().stream()
+                        .map(f -> new FeatureImportance(f, 1.0))
+                        .collect(Collectors.toList())
+        );
+    }
+
+    private static ExplainabilityResultDto createResultDto(CompletableFuture<Map<String, Saliency>> inputFuture, String executionId) {
+        if (!inputFuture.isDone() || inputFuture.isCompletedExceptionally() || inputFuture.isCancelled()) {
```
r00ta: why check `isDone`?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,162 @@
...
+            return new ExplainabilityResultDto(executionId, Collections.emptyMap());
```
r00ta: Not sure this is the expected explaination result in case of an exception

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,162 @@
...
+
+    private static Prediction getPrediction(Map<String, TypedValue> inputs, Map<String, TypedValue> outputs) {
+        PredictionInput input = getPredictionInput(inputs);
+        PredictionOutput output = getPredictionOutput(outputs);
+        return new Prediction(input, output);
+    }
+
+    private static PredictionInput getPredictionInput(Map<String, TypedValue> inputs) {
+        // TODO : convert inputs to a PredictionInput
+        LOG.info("** getPredictionInput called with " + inputs.size() + " inputs ***");
```
r00ta: debug?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/RemoteKogitoPredictionProvider.java
```diff
@@ -0,0 +1,125 @@
...
+                .map(input -> sendPredictRequest(input, namespaceAndName))
+                .reduce(completedFuture(emptyList()),
+                        (cf1, cf2) -> cf1.thenCombine(cf2, this::addElement),
+                        (cf1, cf2) -> cf1.thenCombine(cf2, this::merge));
+    }
+
+    private PredictionOutput toPredictionOutput(JsonObject json) {
+        List<Output> outputs = new LinkedList<>();
+        for (Map.Entry<String, Object> entry : json) {
+            Output output = new Output(entry.getKey(), Type.UNDEFINED, new Value<>(entry.getValue()), 1d);
```
r00ta: Maybe it's not really related to this PR, but why the `score` in the `Output` is set to `1`? what's that?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/RemoteKogitoPredictionProvider.java
```diff
@@ -0,0 +1,125 @@
...
+        Map<String, Object> map = new HashMap<>();
+        for (Feature f : features) {
+            if (Type.COMPOSITE.equals(f.getType())) {
+                List<Feature> compositeFeatures = (List<Feature>) f.getValue().getUnderlyingObject();
+                Map<String, Object> maps = new HashMap<>();
+                for (Feature cf : compositeFeatures) {
+                    Map<String, Object> compositeFeatureMap = toMap(List.of(cf));
+                    maps.putAll(compositeFeatureMap);
+                }
+                map.put(f.getName(), maps);
```
r00ta: Is this going to work if there are more than 2 levels of nesting for structures? a test for that?

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExplainabilityApiV1.java
```diff
@@ -0,0 +1,52 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getStructuredInputs(
+            @Parameter(
+                    name = "executionId",
+                    description = "The execution ID.",
+                    required = true,
+                    schema = @Schema(implementation = String.class)
+            ) @PathParam("executionId") String executionId) {
+        // TODO: implement this
+        return Response.ok(new FeaturesImportanceResponse(List.of(
```
r00ta: Still to be done? Link to ticket?

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/api/ExplainabilityApiV1IT.java
```diff
@@ -0,0 +1,19 @@
...
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
+public class ExplainabilityApiV1IT {
+
+    @Test
+    void testFeatureImportance() {
+        // TODO: implement this
+        given().filter(new ResponseLoggingFilter())
+                .when().get("/executions/decisions/ID/featureImportance")
```
r00ta: still to be done?

r00ta (commented):
Some comments.. The usage of the env variable instead of the property in the dto requires some changes here and there

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/ExplainabilityRequestDto.java
danielezonca: In general it is not correct to have this value injected by the operator because explainable-service (like job-service) needs to know who to call and this is not (or should not be) hard-coded.
I'm fine to have a fallback value provisioned by the operator but we cannot get rid of this parameter.

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/ExplainabilityResultDto.java
danielezonca: Yes @kostola is already working on this :+1: 
Let's keep this comment opened until the change is pushed

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: Refactored

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/ExplainabilityResultDto.java
danielezonca: Introduced `SaliencyDto`

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: This check was wrong, code changed

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: This code was wrong, now has changed

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/api/ExplainabilityApiV1.java
danielezonca: Now it is implemented

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/ExplainabilityRequestDto.java
r00ta: Hi @danielezonca @kostola , following this architecture, who, where and how this `serviceUrl` should be set?

danielezonca: This value contains KogitoApp host, it is injected inside `ConfigBean` ( [template](https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-codegen/src/main/resources/class-templates/config/CdiConfigBeanTemplate.java#L5) ) by the operator  ( [link](https://github.com/kiegroup/kogito-cloud-operator/blob/da94f01ec4e134c5cda407ff85e006ad32ccf829/pkg/controller/kogitoruntime/deployer.go#L99) ).
The flow is:
- Kogito has its own URL
- TraceEvent contains this value
- This value is send to Expl-service
- Expl-service interact with Kogito using it

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/RemoteKogitoPredictionProvider.java
danielezonca: That number is the confidence of output value. We don't have this concept with Kogito so it is always 1 (aka confidence 100%). You can find some usage with OpenNLP tests ( [link](https://github.com/kiegroup/kogito-apps/blob/master/explainability/explainability-integrationtests/explainability-integrationtests-opennlp/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/opennlp/OpenNLPLimeExplainerTest.java#L77) )

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
danielezonca: Done, now the error is propagated

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/SampleWeighterTest.java
danielezonca: Restored

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: It was a WIP, fixed :)

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: It was a WIP, removed

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/RemoteKogitoPredictionProvider.java
danielezonca: Done, added a test for this scenario

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: Fixed :)

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/LimeStabilityTest.java
danielezonca: Done

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/api/ExplainabilityApiV1IT.java
kostola: Done now

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainer.java
```diff
@@ -108,8 +111,15 @@ public PartialDependencePlotExplainer() {
                         predictionInputs.add(input);
                     }
 
+                    List<PredictionOutput> predictionOutputs;
+                    try {
+                        predictionOutputs = model.predict(predictionInputs).get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit());
+                    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                        LOGGER.error("Impossible to obtain prediction {}", e.getMessage());
```
r00ta, line 118: since we throw the same checked exception, isnt't better to log on the consumer side of the method?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -68,24 +75,32 @@ public static double quantifyExplainability(int inputCognitiveChunks, int output
...
 
         PredictionInput predictionInput = new PredictionInput(copy);
-        List<PredictionOutput> predictionOutputs = model.predict(List.of(predictionInput));
-        PredictionOutput predictionOutput = predictionOutputs.get(0);
+        List<PredictionOutput> predictionOutputs;
+        try {
+            predictionOutputs = model.predict(List.of(predictionInput))
+                    .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit());
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            LOGGER.error("Impossible to obtain prediction {}", e.getMessage());
```
r00ta, line 90: same here

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
```diff
@@ -35,35 +40,37 @@
...
-        PredictionProviderMetadata metadata = new PredictionProviderMetadata() {
-            @Override
-            public DataDistribution getDataDistribution() {
-                return DataUtils.generateRandomDataDistribution(10, 100, new Random());
-            }
+    PartialDependencePlotExplainer partialDependencePlotProvider = new PartialDependencePlotExplainer();
+    PredictionProviderMetadata metadata = new PredictionProviderMetadata() {
+        @Override
+        public DataDistribution getDataDistribution() {
+            return DataUtils.generateRandomDataDistribution(10, 100, new Random());
```
r00ta: is this a random unit test?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/LimeExplainerTest.java
```diff
@@ -31,41 +27,60 @@
...
+                    .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit())
+                    .get(0);
             Prediction prediction = new Prediction(input, output);
-            PredictionProvider model = mock(PredictionProvider.class);
-            Assertions.assertThrows(LocalExplanationException.class, () -> limeExplainer.explain(prediction, model));
+            try {
+                limeExplainer.explainAsync(prediction, model)
+                        .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit());
+            } catch (LocalExplanationException e) {
+                // this is expected
```
r00ta: what is the expectation of this test? Should this call raise a `LocalExplanationException` or should it run without exceptions?

explainability/explainability-service-messaging/src/main/java/org/kie/kogito/explainability/messaging/ExplainabilityMessagingHandler.java
```diff
@@ -78,8 +73,7 @@ public ExplainabilityMessagingHandler(ExplanationService explanationService, Exe
             }
 
             CloudEventImpl<ExplainabilityRequestDto> cloudEvent = cloudEventOpt.get();
-            return CompletableFuture
-                    .supplyAsync(() -> handleCloudEvent(cloudEvent), executor)
```
r00ta, line 82: don't we want to use vertx executors?

explainability/explainability-service-messaging/src/main/java/org/kie/kogito/explainability/messaging/ExplainabilityMessagingHandler.java
```diff
@@ -107,15 +101,16 @@ public ExplainabilityMessagingHandler(ExplanationService explanationService, Exe
 
         LOGGER.info("Received CloudEvent with id {} from {}", attributes.getId(), attributes.getSource());
 
-        ExplainabilityRequestDto explainabilityResult = optData.get();
+        ExplainabilityRequest request = ExplainabilityRequest.from(optData.get());
+        PredictionProvider provider = predictionProviderFactory.createPredictionProvider(request);
```
r00ta, line 105: why do we create a new instance at every event?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/PredictionProviderFactory.java
```diff
@@ -16,6 +16,10 @@
 
 package org.kie.kogito.explainability;
 
-public class ExplanationServiceTest {
+import org.kie.kogito.explainability.model.PredictionProvider;
+import org.kie.kogito.explainability.models.ExplainabilityRequest;
 
+public interface PredictionProviderFactory {
```
r00ta, line 22: Why a test class is renamed/moved to the src folder?

r00ta (commented):
Some comments/questions while quickly scrolling the code..

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/LimeExplainerTest.java
danielezonca: `LocalExplainationException` is expected, the previous code was using `assertThrows` but SonarCloud complained that this code
```java
limeExplainer.explainAsync(prediction, model)
      .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit())
```
can throws multiple exceptions so I changed it to explicitly catch the only expected exception and let the test fails otherwise

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainer.java
danielezonca: I prefer to have the log here instead of add a similar code every time this method is invoked. Then the caller can decide to mute or propagate again the exception but the log will be preserved

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
danielezonca: Same as above :)

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/PredictionProviderFactory.java
danielezonca: The original class was empty so probably I moved it and reused, no real reason :)

explainability/explainability-service-messaging/src/main/java/org/kie/kogito/explainability/messaging/ExplainabilityMessagingHandler.java
danielezonca: Prediction provider is tied to request information: for now it is only using `serviceUrl` but in the future I expect we will support multiple `predictionProvider`s and based on the request it will produce the proper one.

explainability/explainability-service-messaging/src/main/java/org/kie/kogito/explainability/messaging/ExplainabilityMessagingHandler.java
danielezonca: I did a quick review of the usage of `CompletableFuture.*Async` in some classes because I think we should use it only when it is really needed (aka long running/blocking tasks): every time we use that code, a new task is scheduled on Vert.x task list for the next loop and this has a cost. We should use it when we start a long task or when we compose an existing one with another long task.

In this case the full execution flow is:
- `explanationService.explainAsync` -> this is the starting point and it already uses (in `RemotePredictionProvider`) Vert.x WebClient to dispatch the REST call as async task
- `sendEvent` -> this should not be blocking because it builds cloud event and then just schedule it with `onNext`
- `message.ack` -> this is a terminal operation that complete the `CompletableFuture` so it is not blocking

Btw we can easily re-introduce it if needed :)

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
jiripetrlik: Please use at least static seed.

danielezonca: This test generates a random set of data because it just tests the cardinality of the result and not the specific value

danielezonca: Btw random removed :)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
```diff
@@ -43,6 +34,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.security.SecureRandom;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
```
r00ta: refactor `import java.util.*;` ;)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
```diff
@@ -97,139 +96,175 @@ public LimeExplainer(int noOfSamples, int noOfPerturbations, int noOfRetries, Ra
...
+                actualOutputs,
+                noOfRetries);
+    }
 
-                        // only fetch the single output to explain in the generated prediction outputs
-                        List<Output> predictedOutputs = new LinkedList<>();
-                        for (PredictionOutput trainingOutput : trainingOutputs) {
-                            Output output = trainingOutput.getOutputs().get(o);
-                            predictedOutputs.add(output);
+    protected CompletableFuture<Map<String, Saliency>> explainRetryCycle(
```
r00ta, line 139: why protected?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
```diff
@@ -97,139 +96,175 @@ public LimeExplainer(int noOfSamples, int noOfPerturbations, int noOfRetries, Ra
...
+            PredictionProvider model,
+            PredictionInput originalInput,
+            PredictionInput targetInput,
+            List<Feature> linearizedTargetInputFeatures,
+            List<Output> actualOutputs,
+            int noOfRetries) {
+
+        List<PredictionInput> perturbedInputs = getPerturbedInputs(originalInput.getFeatures());
+
+        return model.predict(perturbedInputs)
```
r00ta: shouldn't `model.predict` be an async task as well? Use vertx executors?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
```diff
@@ -97,139 +96,175 @@ public LimeExplainer(int noOfSamples, int noOfPerturbations, int noOfRetries, Ra
...
+        double loss = linearModel.fit(trainingSet, sampleWeights);
+        if (!Double.isNaN(loss)) {
+            // create the output saliency
+            int i = 0;
+            for (Feature linearizedFeature : linearizedTargetInputFeatures) {
+                FeatureImportance featureImportance = new FeatureImportance(linearizedFeature, linearModel.getWeights()[i]);
+                featureImportanceList.add(featureImportance);
+                i++;
+            }
+        }
```
r00ta, line 210: if this condition is false we are going to return a kind of empty explaination without any additional information, is this correct from a user perspective?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionProvider.java
```diff
@@ -16,18 +16,19 @@
...
 public interface PredictionProvider {
 
     /**
      * Perform a batch of predictions, given a batch of inputs.
      * @param inputs the input batch
      * @return a batch of prediction outputs
      */
-    List<PredictionOutput> predict(List<PredictionInput> inputs);
-
+    CompletableFuture<List<PredictionOutput>> predict(List<PredictionInput> inputs);
```
r00ta: `predictAsync`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -68,24 +75,32 @@ public static double quantifyExplainability(int inputCognitiveChunks, int output
      * @param topFeatures the list of important features that should be dropped
      * @return the saliency impact
      */
-    public static double impactScore(PredictionProvider model, Prediction prediction, List<FeatureImportance> topFeatures) {
+    public static double impactScore(PredictionProvider model, Prediction prediction, List<FeatureImportance> topFeatures) throws Exception {
```
r00ta: throws `Exception`?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
```diff
@@ -35,35 +33,44 @@
...
-        PredictionProviderMetadata metadata = new PredictionProviderMetadata() {
-            @Override
-            public DataDistribution getDataDistribution() {
-                return DataUtils.generateRandomDataDistribution(10, 100, new Random());
-            }
+    PartialDependencePlotExplainer partialDependencePlotProvider = new PartialDependencePlotExplainer();
+    PredictionProviderMetadata metadata = new PredictionProviderMetadata() {
+        @Override
+        public DataDistribution getDataDistribution() {
+            return DataUtils.generateRandomDataDistribution(10, 100, new FakeRandom());
```
r00ta, line 53: I have mixed feelings about this test in general, but isnt't generating only `1`s? Not sure this is really representative data, if this is actually meaningful for the test itself..

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
```diff
@@ -35,35 +33,44 @@
...
+        public PredictionInput getInputShape() {
+            List<Feature> features = new LinkedList<>();
+            features.add(FeatureFactory.newTextFeature("text", ""));
+            return new PredictionInput(features);
+        }
+
+        @Override
+        public PredictionOutput getOutputShape() {
+            List<Output> outputs = new LinkedList<>();
+            outputs.add(new Output("spam", Type.BOOLEAN, new Value<>(null), 0d));
```
r00ta, line 66: why `score` is set to `0`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -68,24 +75,32 @@ public static double quantifyExplainability(int inputCognitiveChunks, int output
...
+            double size = predictionOutput.getOutputs().size();
+            for (int i = 0; i < size; i++) {
+                Output original = prediction.getOutput().getOutputs().get(i);
+                Output modified = predictionOutput.getOutputs().get(i);
+                impact += (!original.getValue().asString().equals(modified.getValue().asString())
+                        || modified.getScore() < original.getScore() * CONFIDENCE_DROP_RATIO) ? 1d / size : 0d;
+            }
         }
-        return impact / size;
+        return impact;
```
r00ta, line 103: why moving the division inside the sum?

explainability/explainability-service-rest/src/test/java/org/kie/kogito/explainability/rest/PredictionProviderMock.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.kie.kogito.explainability.model.PredictionInput;
+import org.kie.kogito.explainability.model.PredictionOutput;
+import org.kie.kogito.explainability.model.PredictionProvider;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public class PredictionProviderMock implements PredictionProvider {
```
r00ta, line 26: is this used?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,84 @@
...
 
 @ApplicationScoped
 public class ExplanationServiceImpl implements ExplanationService {
 
+    private static final Logger LOG = LoggerFactory.getLogger(ExplanationServiceImpl.class);
+
+    private final LocalExplainer<Map<String, Saliency>> localExplainer;
+
     @Inject
-    ManagedExecutor executor;
```
r00ta, line 33: Don't we use vertx?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,84 @@
...
+        Prediction prediction = getPrediction(request.getInputs(), request.getOutputs());
+        return localExplainer.explainAsync(prediction, predictionProvider)
+                .thenApply(input -> createResultDto(input, request.getExecutionId()))
+                .exceptionally(throwable -> {
+                    LOG.error("Exception thrown during explainAsync", throwable);
+                    return new ExplainabilityResultDto(request.getExecutionId(), Collections.emptyMap());
+                });
+    }
+
+    protected static ExplainabilityResultDto createResultDto(Map<String, Saliency> saliencies, String executionId) {
```
r00ta, line 73: why protected?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -16,25 +16,84 @@
...
+        LOG.debug("Explainability request with executionId {} for model {}:{}",
+                request.getExecutionId(),
+                request.getModelIdentifier().getResourceType(),
+                request.getModelIdentifier().getResourceId());
+        Prediction prediction = getPrediction(request.getInputs(), request.getOutputs());
+        return localExplainer.explainAsync(prediction, predictionProvider)
+                .thenApply(input -> createResultDto(input, request.getExecutionId()))
+                .exceptionally(throwable -> {
+                    LOG.error("Exception thrown during explainAsync", throwable);
+                    return new ExplainabilityResultDto(request.getExecutionId(), Collections.emptyMap());
```
r00ta, line 69: Can we avoid returning an empty map as result? What about adding a property in the DTO with the success information and eventually the reason why there is no result?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/LimeExplainerProducer.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Inject
+    public LimeExplainerProducer(
+            @ConfigProperty(name = "trusty.explainability.numberOfSamples", defaultValue = "100") Integer numberOfSamples,
+            @ConfigProperty(name = "trusty.explainability.numberOfPerturbations", defaultValue = "1") Integer numberOfPerturbations) {
+        this.numberOfSamples = numberOfSamples;
+        this.numberOfPerturbations = numberOfPerturbations;
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(LimeExplainerProducer.class);
```
r00ta: move at top?

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/ConversionUtilsTest.java
```diff
@@ -0,0 +1,160 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.kie.kogito.explainability;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.*;
```
r00ta: remove wildcard?

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/TrustyServiceTest.java
```diff
@@ -61,7 +82,7 @@ void givenADecisionWhenStoreDecisionIsCalledThenNoExceptionsAreThrown() {
     @Test
     @SuppressWarnings("unchecked")
     void givenADecisionWhenADecisionIsStoredAndRetrievedThenTheOriginalObjectIsReturned() {
-        String executionId = "executionId";
+        String executionId = TEST_EXECUTION_ID;
```
r00ta: Remove these assignments ;)

trusty/trusty-storage/trusty-storage-infinispan/src/main/java/org/kie/kogito/trusty/storage/infinispan/ExplainabilityResultItemMarshaller.java
```diff
@@ -0,0 +1,86 @@
...
+        writer.writeObject(ExplainabilityResultItem.SALIENCY_FIELD, input.getSaliency(), Saliency.class);
+    }
+
+    @Override
+    public String getTypeName() {
+        return String.format("%s.%s", ExplainabilityResult.class.getPackageName(), ExplainabilityResultItem.class.getSimpleName());
+    }
+}
+
+class ExplainabilityResultItem {
```
r00ta, line 56: move to another file?

trusty/trusty-storage/trusty-storage-infinispan/src/test/java/org/kie/kogito/trusty/storage/infinispan/testfield/MapToListTestField.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.kie.kogito.trusty.storage.infinispan.testfield;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+public class MapToListTestField<M, K, V, L> extends ListTestField<M, L> {
```
r00ta, line 24: what's that?

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/TestUtils.java
```diff
@@ -0,0 +1,67 @@
...
+public class TestUtils {
+
+    private TestUtils() {
+        // prevent initialization
+    }
+
+    public static final String executionId = "executionId";
+    public static final String serviceUrl = "localhost:8080";
+
+    public static final ModelIdentifier modelIdentifier = new ModelIdentifier("dmn", "name:namespace");
```
r00ta: everything with capital letters?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/LimeExplainerTest.java
r00ta: @danielezonca 
```
Throwable throwable = catchThrowable(() -> limeExplainer.explainAsync(prediction, model)
                        .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit()));
assertThat(throwable).isInstanceOf(LocalExplanationException.class)
```
?

r00ta (commented):
Some comments

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/TestUtils.java
kostola: why? :thinking:

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
danielezonca: No specific reason, it is the "core" method so I prefer to keep it `protected` to allow override if needed

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
danielezonca: This value is not important in this test so it is mocked

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/TestUtils.java
r00ta: because they are constants https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
danielezonca: I agree that this test is not really meaningful so I have created this ticket to improve it
https://issues.redhat.com/browse/FAI-249

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
danielezonca: I prefer to let the implementation of `PredictionProvider` to decide how to provide the result: if it is heavy computation do it on a new thread or just return the value if not. In case of `RemotePredictionProvider` is already async ( [link](https://github.com/kostola/kogito-apps/blob/8a6da048283f548cbcfaff46ecdc80809f2cb8c9/explainability/explainability-service/src/main/java/org/kie/kogito/explainability/RemotePredictionProvider.java#L127) )

explainability/explainability-service-rest/src/test/java/org/kie/kogito/explainability/rest/PredictionProviderMock.java
danielezonca: Yes it is used by [this mock](https://github.com/kostola/kogito-apps/blob/de1d89b35d77e657d62f0fab17d7af67d394ef3d/explainability/explainability-service-rest/src/test/java/org/kie/kogito/explainability/rest/PredictionProviderFactoryMock.java#L31)

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: We are using context propagation abstraction to decouple from the specific executor impl ( [see Quarkus doc](https://quarkus.io/guides/context-propagation) and [MP doc](https://github.com/eclipse/microprofile-context-propagation/blob/master/api/src/main/java/org/eclipse/microprofile/context/ManagedExecutor.java) )

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: Why not? :)
It is a pure static function so it could be even `public` :)
Joking aside, the reason is because it is possible to reuse it if (when?) `ExplainableServiceImpl` will be extended

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
danielezonca: Done

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionProvider.java
danielezonca: Done

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
danielezonca: Fixed

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/LimeExplainerProducer.java
danielezonca: Done

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/ConversionUtilsTest.java
danielezonca: Done

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/TrustyServiceTest.java
danielezonca: Done

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
```diff
@@ -74,4 +81,25 @@ public PredictionOutput getOutputShape() {
...
+        Config.INSTANCE.setAsyncTimeout(1);
+        Config.INSTANCE.setAsyncTimeUnit(TimeUnit.MILLISECONDS);
+
+        PredictionProvider brokenProvider = inputs -> supplyAsync(
+                () -> {
+                    try {
+                        Thread.sleep(1000);
+                        return Collections.emptyList();
+                    } catch (InterruptedException e) {
+                        throw new RuntimeException(e);
```
jiripetrlik: Can we add some description here?

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/TrustyServiceTest.java
```diff
@@ -205,12 +284,68 @@ void givenAModelWhenAModelIsStoredAndRetrievedByIdThenTheOriginalObjectIsReturne
...
+        when(trustyStorageServiceMock.getExplainabilityResultStorage()).thenReturn(storageMock);
+
+        assertThrows(IllegalArgumentException.class, () -> trustyService.getExplainabilityResultById(TEST_EXECUTION_ID));
+    }
+
+    private static JsonNode toJsonNode(String jsonString) {
+        try {
+            return MAPPER.reader().readTree(jsonString);
+        } catch (JsonProcessingException e) {
+            return null;
```
jiripetrlik: Is this correct behaviour? Shouldn't we throw an exception instead?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
danielezonca: Ticket created https://issues.redhat.com/browse/KOGITO-3213

jiripetrlik (approved):
Looks good to me. Few minor comments.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/LimeExplainer.java
danielezonca: Ticket created https://issues.redhat.com/browse/KOGITO-3213

trusty/trusty-storage/trusty-storage-infinispan/src/test/java/org/kie/kogito/trusty/storage/infinispan/testfield/MapToListTestField.java
danielezonca: It is an utility test class to handle map and reuse `MarshallerTestTemplate` mechanism

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
danielezonca: The previous code was considering only one `PredictionOutput` to calculate the impact while now it consider all of them (excluding the top one on purpose). Adding this additional for required to move the division inside the loop.

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
danielezonca: I just wanted to simulated a long call to check the timeout. Improved the message

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/TrustyServiceTest.java
danielezonca: Fixed

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/TestUtils.java
danielezonca: Done

trusty/trusty-storage/trusty-storage-infinispan/src/main/java/org/kie/kogito/trusty/storage/infinispan/ExplainabilityResultItemMarshaller.java
danielezonca: In general I am no a fan of not public inner classes but I think it makes sense in this case: this is an utility class used only to properly serialize and de-serialize a Map with proto. We can create a new file but it is not supposed to be used and it could be even dangerous to have it: someone might see this class and modify it for other reason while this should never happen.
I added a javadoc to clarify

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/lime/LimeExplainerTest.java
danielezonca: My bad, the exception is thrown directly by `explainAsync` without the need of the additional `get`. I have updated the test with the original `assertThrows` and it should be SonarCloud friendly too :)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
r00ta: Understood, but then I have the doubt that now the method is doing something different: in this method  `model.predictAsync` is used with just one request, `predictionOutputs` should a list of just 1 element as before right?

r00ta (approved):
one comment

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
danielezonca: It was a bug not shown by the test: the test is providing 2 features, 1 is removed (top feature) and the other 1 was used for the calculus so the old value was correct (aka equivalent to the new one)

danielezonca (approved):

PR #417, opened 2020-08-26
KOGITO-2777 Add trustyAI input-data and model-lookup sections

kelvah:
Link to [Jira ticket](https://issues.redhat.com/browse/KOGITO-2777).

This PR adds input data and model lookup sections of the execution detail page.


ui-packages/yarn.lock
```diff
@@ -2748,6 +2806,11 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@patternfly/patternfly@2.71.5":
+  version "2.71.5"
+  resolved "https://registry.yarnpkg.com/@patternfly/patternfly/-/patternfly-2.71.5.tgz#db096c07cf3478a3c05f9020e3cd16be9c098be2"
+  integrity sha512-kkPVr+B9UKDeko8F31RzO7g1ecMbp4rVVQhg6X95rrzPd9mqBC8TGuwhyXnnHCbSAyNyqsp+ldGhDMsN5yMPiw==
+
```
kelvah, line 2830: This really old PatternFly version is required by `@kogito-tooling/embedded-editor`. Its usage is confined to the iframe of the embedded editor.

manstis: We can probably move `kogito-apps/ui-packages/packages/trusty` to `@kogito-tooling/embedded-editor@0.6.1` 

This uses `@patternfly/patternfly@4.16.7` see [here](https://github.com/kiegroup/kogito-tooling/blob/master/yarn.lock#L2002).

I'll try it locally first.

manstis (approved):
LGTM, but trusty's `prepare` script did not run following a `yarn install` (even with `node_modules` deleted). I had to run it manually to copy the GWT editor and _envelope_ to the `static` folder.

manstis:
@kelvah FYI, I changed the dependency for `@kogito-tooling/embedded-editor` to `@0.6.0` and everything still works OK.

IDK if you want to move to the newer version of `kogito-tooling` (IMO it'd be advisable).

ui-packages/packages/trusty/package.json
```diff
@@ -27,9 +29,13 @@
     "test:coverage": "yarn test --coverage",
     "test:watch": "jest --watch",
     "lint": "tslint -c ./tslint.json --project . './src/**/*.ts{,x}'",
+    "prepare": "yarn run copy-embedded-editor",
```
jstastny-cz, line 32: Can you walk me through this? When is/should be the prepare script being used?

ui-packages/packages/trusty/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>kie-bc-editors-unpacked</artifactId>
+    <groupId>org.kie.kogito</groupId>
+    <packaging>jar</packaging>
+    <version>0.0.0</version>
```
jstastny-cz, line 24: Hm, is there any other reason for this pom than downloading the editor war file? Should the packaging be still jar? I'd say pom could be enough ...

ui-packages/packages/trusty/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+
+    <repositories>
+        <repository>
+            <id>JBoss Nexus</id>
+            <url>https://repository.jboss.org/nexus/content/groups/public</url>
```
jstastny-cz, line 29: I don't see any particular problem right now, just such configuration can cause problems in future.

ui-packages/packages/trusty/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+            <snapshots>
+                <enabled>true</enabled>
+                <updatePolicy>always</updatePolicy>
+            </snapshots>
+        </repository>
+    </repositories>
+
+    <properties>
+        <version.dmn.webapp>7.39.0-SNAPSHOT</version.dmn.webapp>
```
jstastny-cz, line 38: this version needs to be explicitly updated, no automated update happens when the project moves between versions. Just checking if that's desirable.

ui-packages/packages/trusty/src/components/Organisms/ModelDiagram/pmml/LinearRegressionView.tsx
```diff
@@ -0,0 +1,133 @@
...
+  readonly title: string;
+
+  constructor(m: number, c: number, title: string) {
+    this.m = m;
+    this.c = c;
+    this.title = title;
+  }
+}
+
+// tslint:disable-next-line:max-classes-per-file
```
jstastny-cz: Please comment with any reservations to the current lint settings in KOGITO-3189.

ui-packages/packages/trusty/src/components/Templates/InputData/useInputData.tsx
```diff
@@ -0,0 +1,35 @@
...
+    let isMounted = true;
+    const config: AxiosRequestConfig = {
+      url: `${EXECUTIONS_PATH}/decisions/${executionId}/structuredInputs`,
+      method: 'get'
+    };
+    setInputData({ status: 'LOADING' });
+    httpClient(config)
+      .then(response => {
+        if (isMounted) {
+          setInputData({ status: 'SUCCESS', data: response.data.inputs });
```
jstastny-cz, line 21: the status field is not from axios here? cause there it's defined as number ...

jstastny-cz (commented):
I've noticed a few generated files being committed here. Is that expected? I'd expect we rely on the generation.

kelvah:
> @kelvah FYI, I changed the dependency for `@kogito-tooling/embedded-editor` to `@0.6.0` and everything still works OK.
> 
> IDK if you want to move to the newer version of `kogito-tooling` (IMO it'd be advisable).

@manstis I've updated `@kogito-tooling/embedded-editor` to `0.6.0`. Thank you.

I've also added a `prepare` script to the root `package.json` because yarn workspaces is not running npm lifecycle scripts at package level. Now the embedded editor should be automatically extracted after installation.

ui-packages/packages/trusty/package.json
kelvah: The `prepare` script is one of the predefined [npm life-cycle scripts](https://docs.npmjs.com/misc/scripts). It gets executed automatically after the install script. Which is the right moment to extract the embedded editor.

ui-packages/packages/trusty/src/components/Templates/InputData/useInputData.tsx
kelvah: Almost. Here the status field is a different one used to save the request status inside `setInputData`. The instruction is inside the `then` callback, which is executed when the request succeeds.

manstis:
Jenkins retest this

ui-packages/packages/trusty/pom.xml
manstis: This copies the approach (100%) in `kogito-tooling`. See [here](https://github.com/kiegroup/kogito-tooling/blob/master/packages/kie-bc-editors-unpacked/pom.xml). If there are recommendations to be made that can be applied across all of this type of use-case I think a JIRA reporting the issue and proposing a solution would be great.

ui-packages/packages/trusty/pom.xml
manstis: Correct.. and it's probably not a bad thing; as trusty can tie itself to a stable editor release if needed.

`kogito-tooling` takes an identical approach (updating the version as needed). See [here](https://github.com/kiegroup/kogito-tooling/commits/master/packages/kie-bc-editors-unpacked/pom.xml).

Admittedly we could probably move to `7.42.0-SNAPSHOT` in this PR.

danielezonca (approved):
LGTM 👍

manstis:
Hi @mbiarnes @elguardian @cristianonicolai @r00ta IDK if the failure of this [build](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/blue/organizations/jenkins/KIE%2Fkogito%2Fghprb-webhooks%2Fkogito-apps/detail/kogito-apps/1160/pipeline) is of interest to you `com.google.cloud.tools.jib.http.ResponseException: 429 Too Many Requests. "Too Many Requests. Please see https://docs.docker.com/docker-hub/download-rate-limit/"`. I suspect either the job or a test is downloading repeatedly from Docker Hub (either for this repository in isolation or across `kie` repositories?)

manstis:
Jenkins retest this

kelvah:
Jenkins retest this

kelvah:
@manstis I think we may have an issue with `cxsd`. The last build failed at `yarn install` because:
```
[INFO] make: g++: Command not found
[INFO] make: *** [Release/obj.target/node_expat/node-expat.o] Error 127
```

`cxsd` requires `node-expat`, which relies on `node-gyp` for compilation. So [it needs gcc](https://github.com/nodejs/node-gyp#installation) to work and jenkins seems not to have it installed.

How should we proceed? Is the _temporary_ removal of `cxsd` a possibility? Its generated files are under vcs.

manstis:
Hi @kelvah, please feel free to remove all reference to `cxsd` and keep just the generated sources.

It might be useful to put a `README` pointing to [here](https://github.com/manstis/pmml-codegen/commit/b2d8f8a231044c9c4ec19cbb6084787cc04c9d84) like I have for `jsonix` in the PMML marshaller [too](https://github.com/kiegroup/kogito-tooling/blob/master/packages/pmml-editor-marshaller/src/marshaller/model/README.md).

kelvah:
hi @manstis, I'm afraid it's not enough. [`unmarshaller.ts`](https://github.com/kiegroup/kogito-apps/pull/417/files#diff-17a5a65c6c758015412b346308aac26e) is using `cxml`, a transitive dep of `cxsd`. and it has the same compilation issue.

manstis:
@kelvah oh dear :-(

The use of `cxsd` is replaced by the new PMML marshaller in `kogito-tooling` (see my open PR that is first in line for merge but needs QE approval). Therefore I suggest killing everything relating to `cxsd` in your PR meaning we'll have no Linear Regression Viewer for now (please leave files somewhere I can get them again later when my PR is merged and I can complete the changes).

It's a public holiday in tbe UK today but I'll be checking emails from time to time.

ui-packages/packages/trusty/src/components/Templates/InputData/useInputData.tsx
jstastny-cz: Ah, ok, resolving the comment then.

ui-packages/packages/trusty/pom.xml
jstastny-cz: Thanks for clarifying, I'll leave to you the decision which version to have here.

ui-packages/packages/trusty/package.json
jstastny-cz: Ah, so this will run on each build of the ui-packages maven module, right?
That might lead to excessive time needed for build of unrelated projects (e.g. management-console), when they're being run separately. I am not aware of the details here, regarding the size of downloaded artifact or how the changes affect the build speed - please provide your view.

The current setup is the following:
* ui-packages - maven project to build ALL the UI packages using yarn
* quarkus-apps (e.g. kogito-apps/management-console) that take the bundle from ui-packages/packages/management-console/dist and creating a executable jar.
* the quarkus app has maven dependency on `ui-packages` to build the react apps first when building just a single module instead of whole repo - that's when the build time takes the hit.

If we decide this change is affecting our build time in local dev workflow (not sure if anyone actually works with the project that way) we can deal with this in a separate enhancement.

kelvah:
Jenkins retest this

ui-packages/packages/trusty/package.json
kelvah: Yes, it will run on each build of ui-packages.
I don't see any performance issue though. The `prepare` script is way faster then the install one for example.
It takes ~2s on my local env.

jstastny-cz: that might be due to the fact it is cached (maven command using -nsu).

Just checked the PR check - in clean environment the kie-bc-editors-unpacked maven build takes  17.100 s , that seems reasonable. As I mentioned, it can be tweaked later.

Thanks for the explanation.

jstastny-cz (approved):
LGTM

PR #428, opened 2020-09-01
KOGITO-3228: add outcome ids to saliency model

kostola:
[JIRA task](https://issues.redhat.com/browse/KOGITO-3228)


explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -61,17 +62,17 @@ public ExplanationServiceImpl(
                 request.getExecutionId(),
                 request.getModelIdentifier().getResourceType(),
                 request.getModelIdentifier().getResourceId());
-        Prediction prediction = getPrediction(request.getInputs(), request.getOutputs());
-        return localExplainer.explainAsync(prediction, predictionProvider)
+        return CompletableFuture.supplyAsync(() -> getPrediction(request.getInputs(), request.getOutputs()))
+                .thenCompose(prediction -> localExplainer.explainAsync(prediction, predictionProvider))
```
danielezonca: What is the reason of this change?
`getPrediction` is plain local method and `localExplainer.explainAsync` already create the `CompletableFuture` internally.
I don't see the reason to add this additional wrapper. Btw you should not use any `*Async` methods without providing the executor in a managed environment

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/ExplanationServiceImplTest.java
```diff
@@ -57,27 +58,49 @@ void init() {
...
+        ExplainabilityResultDto resultDto = assertDoesNotThrow(() ->
+                explanationService.explainAsync(REQUEST, predictionProviderMock)
+                        .toCompletableFuture()
+                        .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit())
+        );
+
+        assertNotNull(resultDto);
+        assertEquals(EXECUTION_ID, resultDto.getExecutionId());
+        assertSame(ExplainabilityStatus.FAILED, resultDto.getStatus());
+        assertNotNull(resultDto.getStatusDetails());
```
danielezonca: Please add an assertion on the value too

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumerTest.java
```diff
@@ -43,16 +47,29 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+// TODO: improve this test with decision values
```
danielezonca: Wdym?
Btw can you please create a ticket and link it here if you don't want to include these tests in the PR?

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumerTest.java
```diff
@@ -75,48 +92,48 @@ void testInvalidPayload() {
...
-        Assertions.assertEquals(resultDto.getExecutionId(), explainabilityResult.getExecutionId());
-        Assertions.assertEquals(resultDto.getSaliencies().size(), explainabilityResult.getSaliencies().size());
-        Assertions.assertTrue(resultDto.getSaliencies().containsKey("saliency"));
-        Assertions.assertEquals(resultDto.getSaliencies().get("saliency").getFeatureImportance().size(),
-                explainabilityResult.getSaliencies().get("saliency").getFeatureImportance().size());
+        Assertions.assertEquals(TEST_RESULT_DTO.getExecutionId(), explainabilityResult.getExecutionId());
+        Assertions.assertEquals(TEST_RESULT_DTO.getSaliencies().size(), explainabilityResult.getSaliencies().size());
+        Assertions.assertTrue(TEST_RESULT_DTO.getSaliencies().containsKey("saliency"));
+        // Assertions.assertEquals(TEST_RESULT_DTO.getSaliencies().get("saliency").getFeatureImportance().size(),
+        // explainabilityResult.getSaliencies().get("saliency").getFeatureImportance().size());
```
danielezonca: Can you please update/re-enable these assertions?

trusty/trusty-storage/trusty-storage-api/pom.xml
```diff
@@ -13,6 +13,10 @@
   <name>Kogito :: Trusty Storage API</name>
 
   <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>explainability-api</artifactId>
+    </dependency>
```
danielezonca: From what I can see, the only reason to add this dependency is `ExplainabilityStatus`. Is this really needed? We have `Saliency` and many other classes that are sort of duplicated here so I don't know if it makes sense to add the dependency just for a simple enum. Wdyt?

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
```diff
@@ -82,36 +85,43 @@ protected void handleCloudEvent(CloudEventImpl<ExplainabilityResultDto> cloudEve
         LOGGER.info("Received CloudEvent with id {} from {}", attributes.getId(), attributes.getSource());
 
         ExplainabilityResultDto explainabilityResult = optData.get();
+        String executionId = explainabilityResult.getExecutionId();
+        Decision decision = service.getDecisionById(executionId);
```
danielezonca: Can you please add a log message if there are no `Decision` that match with `executionId`?
Can you also add a test for it?

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
kostola: This is needed imho. The implementation of `LimeExplainer` does not wrap everything in a `CompletableFuture`, but instead does some computation before, that may lead to an exception being thrown (e.g. when the `Prediction` object it receives has no inputs).
With the original implementation, exceptions thrown directly in `explainAsync` wouldn't be catched by the `exceptionally` block and then bubble up to the event loop thread, blocking the whole computation (which means no failure message being sent back to the Trusty Service).
This way, even if the `localExplainer` implementation is not perfect, we are sure to catch every exception it may throw in the `exceptionally` block.

explainability/explainability-service/src/test/java/org/kie/kogito/explainability/ExplanationServiceImplTest.java
kostola: Done :+1:

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
kostola: As @danielezonca suggested in a chat discussion we had, I now handled this issue using a `try catch` block instead of calling `CompletableFuture::supplyAsync` to avoid overloading the event loop too much.

trusty/trusty-storage/trusty-storage-api/pom.xml
kostola: I removed the dependency and created a dedicated `ExplainabilityStatus` enum in `trusty-storage-api`. When we're done with the first release we may discuss a refactoring of this part.

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
```diff
@@ -61,17 +63,17 @@ public ExplanationServiceImpl(
...
-                .thenApply(input -> createResultDto(input, request.getExecutionId()))
-                .exceptionally(throwable -> {
-                    LOG.error("Exception thrown during explainAsync", throwable);
-                    return new ExplainabilityResultDto(request.getExecutionId(), Collections.emptyMap());
-                });
+        try {
+            return localExplainer.explainAsync(getPrediction(request.getInputs(), request.getOutputs()), predictionProvider)
+                    .thenApply(input -> createSucceededResultDto(request.getExecutionId(), input))
+                    .exceptionally(e -> createFailedResultDto(request.getExecutionId(), e));
+        } catch (Exception e) {
```
r00ta, line 70: Is the try catch needed in this case? Is there a test for such scenario?

r00ta (commented):
Just a small comment

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/ExplanationServiceImpl.java
kostola: Yes it is needed and there is a test. Please check [this discussion](https://github.com/kiegroup/kogito-apps/pull/428#discussion_r482002768).

danielezonca: This catch is to prevent `explainAsync` method to throw an unexpected exception instead of return the `CompletableFuture`

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumerTest.java
kostola: Sorry that was a leftover. I implemented that part now.

trusty/trusty-service/src/test/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumerTest.java
kostola: Done. It needed some refactoring but now it's ok.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
kostola: Done :+1:

danielezonca (approved):
LGTM :+1:

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
```diff
@@ -82,36 +86,65 @@ protected void handleCloudEvent(CloudEventImpl<ExplainabilityResultDto> cloudEve
...
+
+        service.storeExplainabilityResult(executionId, explainabilityResultFrom(explainabilityResult, decision));
     }
 
-    protected static ExplainabilityResult explainabilityResultFrom(ExplainabilityResultDto dto) {
+    protected Decision getDecisionById(String executionId) {
+        try {
+            return service.getDecisionById(executionId);
+        } catch (IllegalArgumentException e) {
+            return null;
```
jiripetrlik, line 103: Shouldn't we rather throw or exception or return some error info here?

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
```diff
@@ -82,36 +86,65 @@ protected void handleCloudEvent(CloudEventImpl<ExplainabilityResultDto> cloudEve
...
-                        .collect(Collectors.toMap(Map.Entry::getKey, e -> saliencyFrom(e.getValue())));
-        return new ExplainabilityResult(dto.getExecutionId(), saliencies);
+                        .map(e -> saliencyFrom(outcomeNameToIdMap.get(e.getKey()), e.getKey(), e.getValue()))
+                        .collect(Collectors.toList());
+        return new ExplainabilityResult(dto.getExecutionId(), statusFrom(dto.getStatus()), dto.getStatusDetails(), saliencies);
     }
 
     protected static FeatureImportance featureImportanceFrom(FeatureImportanceDto dto) {
         if (dto == null) {
             return null;
```
jiripetrlik, line 125: Please consider to throw exception.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
```diff
@@ -82,36 +86,65 @@ protected void handleCloudEvent(CloudEventImpl<ExplainabilityResultDto> cloudEve
...
         }
-        return new FeatureImportance(dto.getFeatureId(), dto.getScore());
+        return new FeatureImportance(dto.getFeatureName(), dto.getScore());
     }
 
-    protected static Saliency saliencyFrom(SaliencyDto dto) {
+    protected static Saliency saliencyFrom(String outcomeId, String outcomeName, SaliencyDto dto) {
         if (dto == null) {
             return null;
```
jiripetrlik, line 132: Please consider to throw exception.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
```diff
@@ -82,36 +86,65 @@ protected void handleCloudEvent(CloudEventImpl<ExplainabilityResultDto> cloudEve
...
-}
 
+    protected static ExplainabilityStatus statusFrom(org.kie.kogito.explainability.api.ExplainabilityStatus status) {
+        switch (status) {
+            case SUCCEEDED:
+                return ExplainabilityStatus.SUCCEEDED;
+            case FAILED:
+                return ExplainabilityStatus.FAILED;
+        }
+        return null;
```
jiripetrlik, line 148: Please consider to throw exception.

jiripetrlik (commented):
I have some comments about exception handling.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
kostola: Throwing a `IllegalArgumentException` is the current behavior of the `TrustyServiceImpl` when it doesn't find a decision with the associated ID.
In this case in my implementation I need to return `null` when this happens.
For me it's not an error and in fact it's logged as warning by the caller method because the computation can continue in a "best effort" way even without a `Decision` object.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
kostola: It's not an error in my opinion: if the dto for some reason is `null`, then the model should be `null`.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
kostola: It's not an error in my opinion: if the dto for some reason is `null`, then the model should be `null`.

trusty/trusty-service/src/main/java/org/kie/kogito/trusty/service/messaging/incoming/ExplainabilityResultConsumer.java
kostola: It's not an error in my opinion: if the dto for some reason is `null`, then the model should be `null`.

danielezonca:
To be merged together with
https://github.com/kiegroup/kogito-apps/pull/430
(endpoint update)

jiripetrlik (approved):
Thank you for info. Looks good to me.

r00ta (approved):

PR #430, opened 2020-09-01
KOGITO-3229 Update TrustyUI Explanation section following API changes

kelvah:
[Link to jira ticket](https://issues.redhat.com/browse/KOGITO-3229)

Following changes to the APIs regarding decisions explanation, this PR updates trusty UI to reflect them.

Explanation data is no more provided at execution level. The `/saliencies` endpoint replaces the old `/featureImportance` and it now provides a different set of features scores for each execution outcome.


ui-packages/packages/trusty/api-mock/routes.json
```diff
@@ -4,5 +4,5 @@
   "/executions/:executionType/:executionId/structuredInputs": "/inputs?executionId=:executionId",
   "/executions/:executionType/:executionId/outcomes": "/outcomes?header.executionId=:executionId&singular=1",
   "/executions/:executionType/:executionId/outcomes/:outcomeId": "/outcomeDetail?outcomeId=:outcomeId&singular=1",
-  "/executions/:executionType/:executionId/featureImportance": "/featureImportance?executionId=:executionId"
+  "/executions/:executionType/:executionId/saliencies": "/saliencies"
```
manstis: There's talk on zulip of using `executions/decisions/{exId}/explain/saliencies`.

@r00ta @kostola What was finally agreed?

ui-packages/packages/trusty/src/components/Templates/Explanation/useFeaturesScores.tsx
```diff
@@ -1,57 +1,42 @@
...
+import { FeatureScores, RemoteData, Saliencies } from '../../../types';
+import { orderBy, find } from 'lodash';
 
-const useFeaturesScores = (executionId: string) => {
-  const [featuresScores, setFeaturesScores] = useState<
-    RemoteData<Error, FeatureScores[]>
-  >({
-    status: 'NOT_ASKED'
-  });
+const useFeaturesScores = (
```
manstis, line 5: Is this still the correct name for this hook?

manstis: (And file name)

manstis (commented):
I could not find how to show the `ExplanationError` component in the UI.

Also, unrelated to this PR, I found the popup for the "Full Tornado Chart" could not be closed.

ui-packages/packages/trusty/src/components/Templates/Explanation/useFeaturesScores.tsx
kelvah: `useSaliencies` is now the hook fetching the `saliences` endpoint and retrieving all the explainability info for the current execution, while `useFeaturesScores` gets the scores for the currently selected outcome (plus some manipulation).
it seemed OK to call them like that. Do you have any suggestion about renaming them?

kelvah:
> I could not find how to show the `ExplanationError` component in the UI.
> 
> Also, unrelated to this PR, I found the popup for the "Full Tornado Chart" could not be closed.

Thank you @manstis for spotting the modal bug. Fixed it.

Unfortunately it's not easy to add an example for the explanation error to the mocked APIs. I can't filter mocked responses because I don't have `executionId` inside them. For now the only way is to alter saliencies mock with a `FAILED` status :|

ui-packages/packages/trusty/api-mock/routes.json
kelvah: I've updated the endpoint to `/explanations/saliencies`.

manstis (approved):
Thank-you @kelvah. 

I set the `saliences` to `FAILED` and could see the component.

I also confirm the popup can now be dismissed too.

kelvah:
Jenkins retest this

ui-packages/packages/trusty/src/components/Templates/Explanation/Explanation.tsx
```diff
@@ -162,9 +177,9 @@ const Explanation = ({ outcomes }: ExplanationProps) => {
               </Title>
             </StackItem>
             <StackItem>
-              {(featuresScores.status === 'LOADING' ||
-                (featuresScores.status === 'SUCCESS' &&
-                  featuresScores.data.length > 0)) && (
+              {(saliencies.status === 'LOADING' ||
```
jstastny-cz, line 180: I would consider putting all the status values into an enum and replace the occurrences of string literals, but not introduced in this PR.

ui-packages/packages/trusty/src/components/Templates/Explanation/tests/useFeaturesScores.test.tsx
```diff
@@ -1,21 +1,37 @@
...
+    }
+
+    expect(result.current.featuresScores).toStrictEqual(sortedFeatures);
+  });
+});
+
+const saliencies = {
+  status: 'SUCCESS',
+  data: {
+    status: 'SUCCEEDED',
```
jstastny-cz, line 31: especially here would the enums be useful - I have troubles identifying what the applicable values are in which property.

ui-packages/packages/trusty/src/types.ts
```diff
@@ -70,9 +70,20 @@ export interface Outcome {
...
+  featureId: string;
   featureScore: number;
 }
 
+export interface Saliency {
+  outcomeId: string;
+  featureImportance: FeatureScores[];
+}
+export interface Saliencies {
+  status: 'SUCCEEDED' | 'FAILED';
```
jstastny-cz, line 82: I would suggest rely here on an enum that'd be exported and used globally.

jstastny-cz (commented):
I would suggest moving from string literal values to enums in status fields. See my comments inline.

Otherwise I don't have any reservations.

kelvah:
> I would suggest moving from string literal values to enums in status fields. See my comments inline.
> 
> Otherwise I don't have any reservations.

Thank you @jstastny-cz. I think I agree with you, enums would help with type assertions.

Given the change would impact the whole app (both `Saliencies` and `RemoteData` types), I would open a separate jira and proceed with the refactoring in another PR. What do you think?

jstastny-cz:
> > I would suggest moving from string literal values to enums in status fields. See my comments inline.
> > Otherwise I don't have any reservations.
> 
> Thank you @jstastny-cz. I think I agree with you, enums would help with type assertions.
> 
> Given the change would impact the whole app (both `Saliencies` and `RemoteData` types), I would open a separate jira and proceed with the refactoring in another PR. What do you think?

Sure, that would be enough.

kelvah:
@jstastny-cz Activity tracked [here](https://issues.redhat.com/browse/KOGITO-3268)

jstastny-cz (approved):

danielezonca (approved):

danielezonca:
To be merged together with
https://github.com/kiegroup/kogito-apps/pull/428
(endpoint update)

PR #449, opened 2020-09-08
KOGITO-3299 Index Job endpoint url in Data Index

cristianonicolai:


pom.xml
```diff
@@ -21,6 +21,8 @@
     <version.npm>6.10.3</version.npm>
     <version.yarn>v1.19.1</version.yarn>
     <maven.compiler.release>11</maven.compiler.release>
+    <maven.compiler.source>11</maven.compiler.source>
```
cristianonicolai, line 24: @nmirasch @tiagodolphine I had to change this because Intellij was not picking up the correct compiler level locally. In theory, this should not have any impact as we already use version 11 anyway.

tiagodolphine: I faced the same issue on a different PR, I was keeping this locally, so that's fine to insert this IMO.

tiagodolphine (approved):

AjayJagan (approved):

ricardozanini:
@cristianonicolai this endpoint should be set by the operator, right? Mind opening an issue to me? Thanks!

nmirasch (approved):

Sara4994 (approved):

cristianonicolai:
> @cristianonicolai this endpoint should be set by the operator, right? Mind opening an issue to me? Thanks!

@tiagodolphine can you confirm this is the `kogito.service.url` property that needs to be set?

tiagodolphine:
> > @cristianonicolai this endpoint should be set by the operator, right? Mind opening an issue to me? Thanks!
> 
> @tiagodolphine can you confirm this is the `kogito.service.url` property that needs to be set?

Job Service uses the property `kogito.service.url` when setting the `source` on the cloud event https://github.com/kiegroup/kogito-apps/blob/5222e30a194acab82d32b24b2872f49c75522867/jobs-service/src/main/java/org/kie/kogito/jobs/service/stream/KafkaJobStreams.java#L76
This is not a new property but can't confirm if the operator is already setting it... 
@ricardozanini should I open a JIRA?
Thanks.

ricardozanini:
@tiagodolphine don't mind, if it's using this property we are fine, but we will change that to use the **internal** endpoint, not the **external**, since exposing endpoints is an admin / company policy, not our operator's job. @cristianonicolai we need to figure how the management console will behave. Have we prioritized the NGINX solution?

cristianonicolai:
> @tiagodolphine don't mind, if it's using this property we are fine, but we will change that to use the **internal** endpoint, not the **external**, since exposing endpoints is an admin / company policy, not our operator's job. @cristianonicolai we need to figure how the management console will behave. Have we prioritized the NGINX solution?

Not really, we still need to finish the backend gateway implementation.

PR #466, opened 2020-09-22
KOGITO-3119 - Job Service - Infinispan 11 upgrade

tiagodolphine:

### DIFF
--- a/jobs-service/docker/infinispan/infinispan.xml
+++ b/jobs-service/docker/infinispan/infinispan.xml
@@ -1,0 +1,27 @@
+<infinispan xmlns='urn:infinispan:config:10.0' xsi:schemaLocation='urn:infinispan:config:10.0 http://www.infinispan.org/schemas/infinispan-config-10.0.xsd                             urn:infinispan:server:10.0 http://www.infinispan.org/schemas/infinispan-server-10.0.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>
+  <server xmlns='urn:infinispan:server:10.0'>
+    <interfaces>
+      <interface name='public'>
+        <inet-address value='${infinispan.bind.address:0.0.0.0}'/>
+      </interface>
+    </interfaces>
+    <socket-bindings default-interface='public' port-offset='0'>
+      <socket-binding name='default' port='11222'/>
+    </socket-bindings>
+    <security>
+      <security-realms>
+        <security-realm name='default'>
+          <properties-realm groups-attribute='Roles'>
+            <user-properties path='users.properties' relative-to='infinispan.server.config.path'/>
+            <group-properties path='groups.properties' relative-to='infinispan.server.config.path'/>
+          </properties-realm>
+        </security-realm>
+      </security-realms>
+    </security>
+    <endpoints socket-binding='default' security-realm='default'>
+      <hotrod-connector name="hotrod"/>
+      <rest-connector name="rest"/>
+    </endpoints>
+  </server>
+</infinispan>
+
--- a/jobs-service/docker/infinispan/infinispan.xml
+++ b/jobs-service/docker/infinispan/infinispan.xml
@@ -1,5 +1,6 @@
-<infinispan xmlns='urn:infinispan:config:10.0' xsi:schemaLocation='urn:infinispan:config:10.0 http://www.infinispan.org/schemas/infinispan-config-10.0.xsd                             urn:infinispan:server:10.0 http://www.infinispan.org/schemas/infinispan-server-10.0.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>
-  <server xmlns='urn:infinispan:server:10.0'>
+<infinispan xmlns='urn:infinispan:config:11.0' xsi:schemaLocation='urn:infinispan:config:11.0 http://www.infinispan.org/schemas/infinispan-config-11.0.xsd urn:infinispan:server:11.0 http://www.infinispan.org/schemas/infinispan-server-11.0.xsd'
+            xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>
+  <server xmlns='urn:infinispan:server:11.0'>
     <interfaces>
       <interface name='public'>
         <inet-address value='${infinispan.bind.address:0.0.0.0}'/>
@@ -24,4 +25,3 @@
     </endpoints>
   </server>
 </infinispan>
-
--- a/jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanConfiguration.java
+++ b/jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanConfiguration.java
@@ -77,10 +77,10 @@ public class InfinispanConfiguration {
 
     void initializeCaches(@Observes @Priority(Interceptor.Priority.PLATFORM_BEFORE) StartupEvent startupEvent,
                           @ConfigProperty(name = PERSISTENCE_CONFIG_KEY) Optional<String> persistence,
-                          RemoteCacheManager remoteCacheManager,
+                          Instance<RemoteCacheManager> remoteCacheManager,
                           Event<InfinispanInitialized> initializedEvent) {
         isEnabled(persistence)
-                .map(c -> remoteCacheManager.administration().getOrCreateCache(JOB_DETAILS, getCacheTemplate()))
+                .map(c -> remoteCacheManager.get().administration().getOrCreateCache(JOB_DETAILS, getCacheTemplate()))
                 .ifPresent(c -> {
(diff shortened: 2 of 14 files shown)


jobs-service/src/main/resources/META-INF/library.proto
```diff
@@ -1,16 +1,20 @@
 package job.service;
 
+/* @Indexed */
 message JobDetails {
   optional string id = 1;
   optional string correlationId = 2;
+  /* @Field(store = Store.YES, analyze = Analyze.YES, analyzer = @Analyzer(definition = "keyword")) */
```
cristianonicolai: @tiagodolphine do we need the analyze part? AFAIK that would be to break down the content, but this status field is a single value right?

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanJobRepository.java
```diff
@@ -89,36 +107,37 @@ public InfinispanJobRepository(Vertx vertx,
...
-                                                    .between(DateUtil.zonedDateTimeToInstant(from),
-                                                             DateUtil.zonedDateTimeToInstant(to))
-                                                    .orderBy("priority", SortOrder.DESC)
-                                                    .<JobDetails>build()
-                                                    .list());
+        Query<JobDetails> query = queryFactory.create("from job.service.JobDetails j " +
+                                                              "where " +
+                                                              "j.trigger.nextFireTime > :from " +
+                                                              "and j.trigger.nextFireTime < :to " +
+                                                              "and j.status:(" +  createStatusQuery(status) + ") " +
```
cristianonicolai: perhaps `j.status in ( value1, value2 )` ?

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanConfiguration.java
```diff
@@ -50,9 +65,37 @@ private Caches() {
...
+        return persistence
+                .filter("infinispan"::equals);
+    }
+
+    void initializeCaches(@Observes @Priority(Interceptor.Priority.PLATFORM_BEFORE) StartupEvent startupEvent,
+                          @ConfigProperty(name = PERSISTENCE_CONFIG_KEY) Optional<String> persistence,
+                          RemoteCacheManager remoteCacheManager,
+                          Event<InfinispanInitialized> initializedEvent) {
+        isEnabled(persistence)
+                .map(c -> remoteCacheManager.administration().getOrCreateCache(JOB_DETAILS, getCacheTemplate()))
```
cristianonicolai: @tiagodolphine would be important to try to have a consistent behaviour here compared to other services, see https://github.com/kiegroup/kogito-apps/blob/master/persistence-commons/persistence-commons-infinispan/src/main/java/org/kie/kogito/persistence/infinispan/cache/InfinispanCacheManager.java#L79.
Point is that the user should be able to define a template. That template allows users to take over the configuration on Infinispan side to match their need. Then if thats not provided, the default cache configuration would be in place as you specified here.

jobs-service/src/main/resources/META-INF/library.proto
tiagodolphine: when running the ispn 11 and executing the query string I was getting the error that it was missing the `analyze` int the attribute so basically when I added it solved the issue... that's why I added, the query in this case use the `IN` operation.

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanJobRepository.java
tiagodolphine: I think so, I'll add  `IN` and test, I just used the `: ` because it was on the some example on the doc, but they should be the same operation.

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanConfiguration.java
tiagodolphine: My approach it quite different, on the InfinispanConfiguration I´m running all the requirement to initialize infinispan, like creating the cache and healthcheck, ... but we can migrate maybe to use the same approach of `persistence-commons` with the `InfinispanCacheManager`, IMO we could to it on a new JIRA, just for this. The other thing is, I created the template approach in this PR because I need to add more configs to work on ispn11, so the user may use a different template if he changes the `.xml` template, before it was not even using any template, it was using the `default` one.

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanJobRepository.java
tiagodolphine: well, I tried to change and use `IN` instead of `:` but I got an error on the query...

tiagodolphine:
@cristianonicolai the problem I see using infinispan 11 just on jobs-service is that for the IT it is ok because we can override the version but running it on the BDD tests we may have a problem using infinispan 10, I tested here and it seems to not work after the changes for 11 =/

cristianonicolai:
@tiagodolphine, in that case, we need to sync and commit the data index part together I'm afraid :/

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanConfiguration.java
cristianonicolai: We can take advantage of the template system inheritance, see https://infinispan.org/docs/11.0.x/titles/configuring/configuring.html#templates-configuring. That way you only specify the indexing class that is needed but gives room for the user to customize the base template. But that's fine if you want to create a JIRA for reviewing that alone.

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanJobRepository.java
cristianonicolai: hm, I think  `:` uses index, that's why you also needed the `analyze` part. But `in` should work, I'm using that in data index, whats the error?

jobs-service/src/main/resources/META-INF/library.proto
cristianonicolai: yeah, I think thats because of the operator you're using, using `in` should avoid that.

jobs-service/src/main/java/org/kie/kogito/jobs/service/repository/infinispan/InfinispanJobRepository.java
tiagodolphine: it was working with `IN` with the query dsl, but when I changed the query to ickle, it didn't work... so basically I changed the operator `:` anyway, I think we will not be able to merge this PR because it is not compatible with ispn10, so or we find a way to make it work with both or we need to wait for data-index, wdyt?

tiagodolphine:
jenkins retest this

tiagodolphine:
@cristianonicolai so, I finally got it to work. Now I'm using the `IN` operator instead of `: ` as you suggested, for this, I needed to change the cache template I was using removing the indexing entity configuration, and the query syntax, that was causing the error because it is different like `value1 value2 ...` to `value1, value2...` with IN it is necessary to put a comma between the values and `:` no. In the end, it worked, and plus now it works with both ispn 10 and 11 :) \o/

cristianonicolai (approved):

jobs-service/docker/infinispan/infinispan.xml
```diff
@@ -0,0 +1,27 @@
+<infinispan xmlns='urn:infinispan:config:10.0' xsi:schemaLocation='urn:infinispan:config:10.0 http://www.infinispan.org/schemas/infinispan-config-10.0.xsd                             urn:infinispan:server:10.0 http://www.infinispan.org/schemas/infinispan-server-10.0.xsd' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>
+  <server xmlns='urn:infinispan:server:10.0'>
```
ricardozanini: Don't we want to change the schema to https://infinispan.org/schemas/infinispan-server-11.0.xsd?

ricardozanini (commented):
Just a minor observation regarding the Infinispan XSD.

jobs-service/docker/infinispan/infinispan.xml
tiagodolphine: updated it to 11

ricardozanini (approved):

tiagodolphine:
Hi @sutaakar, if you have some time to take a look at this PR, basically is to upgrade jobs-service to use infinspan 11 and to be compatible with 10 as well.
Thanks.

sutaakar (approved):
Looks good to me.

PR #471, opened 2020-10-02
KOGITO-3452: uniform error handling of build/encode/decode in CloudEventUtils

kostola:
- [JIRA task](https://issues.redhat.com/browse/KOGITO-3452)

### DIFF
--- a/explainability/explainability-service-messaging/src/main/java/org/kie/kogito/explainability/messaging/ExplainabilityMessagingHandler.java
+++ b/explainability/explainability-service-messaging/src/main/java/org/kie/kogito/explainability/messaging/ExplainabilityMessagingHandler.java
@@ -21,6 +21,7 @@ import java.net.URI;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+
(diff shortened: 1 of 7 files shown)


ruromero (approved):

danielezonca (approved):
LGTM

kostola:
jenkins retest this please

ricardozanini (commented):
@ruromero this makes sense in other parts of the code you refactored, like in the knative addon?

danielezonca:
@ricardozanini 
Wdym?
This PR is related to the refactor of [this class](https://github.com/kiegroup/kogito-runtimes/blob/master/addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java) (now it returns an `Optional`) that is used here. So I don't expect any impact on knative addon.
Btw that class is quite generic so we can consider it to move it to a shared lib somewhere

ruromero:
I agree with @danielezonca I don't think it affects the addon.

ricardozanini:
> Btw that class is quite generic so we can consider it to move it to a shared lib somewhere


That's why I asked 😀

Looked like something we already had, I'm sorry if I caused any mess.

Thanks for the clarification guys!

PR #475, opened 2020-10-06
KOGITO-2092 Task inbox filters building block

AjayJagan:

Description: this PR adds a toolbar to the task inbox page. The toolbar consists of filters which can be used to filter out the user task instances.
JIRA: https://issues.redhat.com/browse/KOGITO-2092


ui-packages/packages/task-console/server/server.js
```diff
@@ -59,37 +59,82 @@ function timeout(ms) {
...
-        } else {
-          // querying tasks assigned to current user
-          const actualOwnerClause = args['where'].or[0];
-          if(actualOwnerClause.actualOwner.equal === datum.actualOwner) {
-            return true;
+        console.log('args', args)
+        
+        if (args['where'].and) {
+          // if filter available
+          if(!checkTaskAssignment(datum,args['where'].and[0])){
```
jstastny-cz: white space formatting (missing space after if).

ui-packages/packages/task-console/src/components/Molecules/TaskInboxToolbar/TaskInboxToolbar.tsx
```diff
@@ -0,0 +1,209 @@
...
+    } else {
+      _.remove(selectedStatus, (status: string) => {
+        return status === selection;
+      });
+    }
+
+    forceUpdate();
+  };
+
+  const onDelete = (type: string = '', id: string = ''): void => {
```
jstastny-cz: I would rename the `id` argument into something more general, as in both cases below, the value is not an id really, it's either state or taskName ... so I would rename id->value or similar.
That might also apply to the type argument ... it's rather a categoryName passed to the ToolbarFilter component, right? Also in case of this attribute and checking strings below, I'd like to see a more tighter relation, not just string comparison - can we have an enum Category with values Status and TaskName for example? those values would be passed as categoryName to ToolbarFilter and then matched here below (possibly by a switch instead of two independent ifs).

ui-packages/packages/task-console/src/components/Molecules/TaskInboxToolbar/tests/TaskInboxToolbar.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+      .find(Select)
+      .find('button')
+      .simulate('click');
+    wrapper = wrapper.update();
+    // check length to be 5
+    expect(wrapper.find(Select).find(SelectOption).length).toBe(5);
+    wrapper
+      .find(Select)
+      .find(SelectOption)
+      .at(2)
```
jstastny-cz: can't we filter the options based on the status string rather than relying on fixed offset?

ui-packages/packages/task-console/src/components/Organisms/TaskInbox/tests/TaskInbox.test.tsx
```diff
@@ -324,4 +469,188 @@ describe('TaskInbox tests', () => {
       direction: 'asc'
     });
   });
+
+  it('show no filters selected state and select reset', async () => {
```
jstastny-cz: We need to mock the TaskInboxToolbar in all but this test to prevent the huge snapshot file.
Easiest way might be:
1. move this test to a separate file.
2. in TaskInbox.test.tsx mock the TaskInboxToolbar component.
3. in the new file don't mock it (for exactly this test).

jstastny-cz (changes requested):

pefernan (approved):
Works fine and code looks well. After @jstastny-cz suggestions +1

ui-packages/packages/task-console/server/server.js
AjayJagan: done :)

ui-packages/packages/task-console/src/components/Molecules/TaskInboxToolbar/TaskInboxToolbar.tsx
AjayJagan: done :)

ui-packages/packages/task-console/src/components/Molecules/TaskInboxToolbar/tests/TaskInboxToolbar.test.tsx
AjayJagan: done :)

ui-packages/packages/task-console/src/components/Organisms/TaskInbox/tests/TaskInbox.test.tsx
AjayJagan: @jstastny-cz , please tell me if the file name looks ok

jstastny-cz (approved):
LGTM, thanks for the changes.

pefernan:
Thanks @AjayJagan

PR #479, opened 2020-10-07
FAI-254 Implement counterfactual search using OptaPlanner

ruivieira:
[FAI-254] Implement counterfactual search using OptaPlanner

Implementation of a counterfactual search using OptaPlanner as the engine.

Current limitations at the moment:

- Only numerical (integer, double) and boolean features supported
- No feature weighting (this relates to a broader discussion on how to provide domain metadata to local explanations)
- No trade-off tuning parameter (primary goal for OptaPlanner's solution is the desired outcome)

> 
> 
> - [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
> - [x] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
> - [x] Pull Request contains link to any dependent or related Pull Request

### DIFF
--- a/explainability/explainability-core/pom.xml
+++ b/explainability/explainability-core/pom.xml
@@ -10,7 +10,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <properties>
-        <optaplanner.version>8.0.0-SNAPSHOT</optaplanner.version>
+        <version.org.optaplanner>8.0.0-SNAPSHOT</version.org.optaplanner>
     </properties>
 
     <dependencyManagement>
@@ -18,7 +18,7 @@
             <dependency>
                 <groupId>org.optaplanner</groupId>
                 <artifactId>optaplanner-core</artifactId>
-                <version>${optaplanner.version}</version>
+                <version>${version.org.optaplanner}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/explainability/explainability-core/pom.xml
+++ b/explainability/explainability-core/pom.xml
@@ -1,39 +1,59 @@
 <?xml version="1.0"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-  <parent>
-    <groupId>org.kie.kogito</groupId>
-    <artifactId>explainability</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
-  </parent>
-  <modelVersion>4.0.0</modelVersion>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>explainability</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
 
-  <artifactId>explainability-core</artifactId>
-  <name>Kogito :: Explainability Core</name>
-  <dependencies>
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.assertj</groupId>
-      <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-  </dependencies>
+    <properties>
+        <optaplanner.version>8.0.0-SNAPSHOT</optaplanner.version>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.optaplanner</groupId>
+                <artifactId>optaplanner-core</artifactId>
+                <version>${optaplanner.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+
+    <artifactId>explainability-core</artifactId>
+    <name>Kogito :: Explainability Core</name>
+    <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.optaplanner</groupId>
+            <artifactId>optaplanner-core</artifactId>
+        </dependency>
+    </dependencies>
 
 </project>
--- a/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterFactualScoreCalculator.java
+++ b/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterFactualScoreCalculator.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.explainability.local.counterfactual;
+
+
+import org.kie.kogito.explainability.local.counterfactual.entities.CounterfactualEntity;
+import org.kie.kogito.explainability.model.Feature;
+import org.kie.kogito.explainability.model.Output;
+import org.kie.kogito.explainability.model.PredictionInput;
+import org.kie.kogito.explainability.model.PredictionOutput;
+import org.optaplanner.core.api.score.buildin.bendablebigdecimal.BendableBigDecimalScore;
+import org.optaplanner.core.api.score.calculator.EasyScoreCalculator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.math.BigDecimal;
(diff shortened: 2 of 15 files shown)


explainability/explainability-core/pom.xml
```diff
@@ -1,39 +1,59 @@
...
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.assertj</groupId>
-      <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-  </dependencies>
+    <properties>
+        <optaplanner.version>8.0.0-SNAPSHOT</optaplanner.version>
```
danielezonca: We usually use a different naming convention: `version.groupId` so in this case `version.org.optaplanner`

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
```diff
@@ -0,0 +1,141 @@
...
+                    final DoubleEntity doubleEntity = new DoubleEntity(feature, featureDistribution.getStart(), featureDistribution.getEnd(), isConstrained);
+                    entities.add(doubleEntity);
+                } else if (feature.getValue().getUnderlyingObject() instanceof Integer) {
+                    final IntegerEntity integerEntity = new IntegerEntity(feature, (int) featureDistribution.getStart(), (int) featureDistribution.getEnd(), isConstrained);
+                    entities.add(integerEntity);
+                }
+            } else if (feature.getType() == Type.BOOLEAN) {
+                final BooleanEntity booleanEntity = new BooleanEntity(feature, isConstrained);
+                entities.add(booleanEntity);
+            }
```
danielezonca: What about move this code in an utility class so that it will be easier to test and extend with other data types?
Something like
```java
class CounterfactualEntityFactory {
    public static CounterfactualEntity from(Feature feature) {
         if (feature.getType() == Type.NUMBER) {
             ....
         } else if (feature.getType() == Type.BOOLEAN) {
             ....
         } else {
             throw new IllegalArgumentException("Unsupported feature type: " + feature.getType());
         } 
    }
}
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            // Wait until the solving ends
+            solution = solverJob.getFinalBestSolution();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new IllegalStateException("Solving failed: {}", e);
+        }
+
+        solverManager.close();
+
+        return CompletableFuture.completedFuture(solution.getEntities());
```
danielezonca: I think we can wrap the logic and make it async like
```suggestion
        return CompletableFuture.supplyAsync(() -> {
            SolverJob<CounterfactualSolution, UUID> solverJob = solverManager.solve(problemId, problem);
            CounterfactualSolution solution;
            try {
                // Wait until the solving ends
                solution = solverJob.getFinalBestSolution();
                return solution.getEntities();
            } catch (InterruptedException | ExecutionException e) {
                throw new IllegalStateException("Solving failed: {}", e);
            } finally {
                solverManager.close();
            }
        }, getExecutor());
```

Note: add an `Executor` field + constructor with `Executor`. Then the method `getExecutor()` should return that value or `ForkJoinPool.commonPool()` if no `Executor` has been explicitly provided

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualSolution.java
```diff
@@ -0,0 +1,82 @@
...
+public class CounterfactualSolution {
+
+    @PlanningEntityCollectionProperty
+    private List<CounterfactualEntity> entities;
+
+    @JsonIgnore
+    private List<Output> goal;
+
+    @JsonIgnore
+    private PredictionProvider model;
```
danielezonca: What is the reason for these `JsonIgnore`? Is this class intended to be serialized?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/entities/BooleanEntity.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Returns the {@link BooleanEntity} as a {@link Feature}
+     *
+     * @return {@link Feature}
+     */
+    @Override
+    public Feature asFeature() {
+        return FeatureFactory.newBooleanFeature(feature.getName(), this.value);
```
danielezonca: Why don't return `feature` instead of create a new one?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/entities/DoubleEntity.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Returns the {@link BooleanEntity} as a {@link Feature}
+     *
+     * @return {@link Feature}
+     */
+    @Override
+    public Feature asFeature() {
+        return FeatureFactory.newNumericalFeature(feature.getName(), this.value);
```
danielezonca: Same as above

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/entities/IntegerEntity.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * Returns the {@link IntegerEntity} as a {@link Feature}
+     *
+     * @return {@link Feature}
+     */
+    @Override
+    public Feature asFeature() {
+        return FeatureFactory.newNumericalFeature(feature.getName(), this.value);
```
danielezonca: Same as above

danielezonca (commented):
Thanks for the PR, very well done
I have the impression some of the code of `IntegerEntity`, `BooleanEntity` and `DoubleEntity` can be moved to a common abstract class but we can review this in the future

ge0ffrey:
OptaPlanner 8 is likely going to be part of the 0.17 kogito release train, which means that afterwards it's safe for this work to depend on the 8.0.0-SNAPSHOT if this work is part of the kogito release train too (or otherwise depend on that release).
@rsynek FYI ^^ trustyAI depends on optaplanner

ruivieira:
@danielezonca thanks for the comments!

This commit implements:

- Change POM properties name
- Extract `CounterfactualEntity` creation to a utility method
- Use `supplyAsync` with `Executor` for counterfactual calculation
- Remove JSON annotations from `CounterfactualSolution`

The `asFeature` method was not changed, explained in the conversations above.

explainability/explainability-core/pom.xml
ruivieira: Changed

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
ruivieira: Implemented

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
ruivieira: Implemented

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualSolution.java
ruivieira: Good catch! It was used in the PoC (REST endpoint returned the solution), but it's not needed here.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/entities/BooleanEntity.java
ruivieira: @danielezonca For a `CounterfactualEntity`, the constructor takes a `feature` as a means of passing the original input.
However, during the counterfactual search, we take the `value` of an entity (populated by OptaPlanner) which will be different from `feature`.
Although it is correct, I agree it can be confusing. Perhaps some solutions can be:

- `CounterfactualEntity` returns simple the `value` (current solution value) and the solution builds the feature itself
- Rename it from `asFeature` to `getSolutionFeature` or similar

wdyt?

danielezonca: Ok thanks for the explanation, now it is clear to me the overall flow.
From my point of view the most confusing thing is to have a `feature` field and a `value` initialized starting from it. This because only if you know how OP works and notice the `@PlanningVariable` annotation you realize that `value` could be changed.

As alternative proposal I suggest to remove `Feature` from the class as follow:
- Make constructor accept `Value<?> originalValue, String featureName, ...` so no `Feature` is stored in the class 
- Create in all *Entity class a `.from(Feature originalFeature, ...)` factory method that internally invokes the new constructor and then update `CounterfactualEntityFactory` accordingly
- Update *Entity fields removing `Feature feature`, adding `Value<?> originalValue` and renaming `value` as `proposedValue` / `newValue`

In this way we preserve the current behavior but we remove `feature` field ambiguity
Wdyt?
As alternative with lower impact what about just renaming `feature` as `originalFeature` and `value` as `proposedValue` / `newValue`?

ruivieira: Completely agree that it's much clearer that way. After a second iteration on the code, I've also realised that we are always using the underlying object from the `Value` (for distance calculation and equality comparison), so the new commit actually gets rid of the `Value` type and just stores,`T proposedValue`/`T originalValue` for the supported types (`Integer`, `Double`, `Boolean`).
wdyt?

ruivieira:
- Refactor nomenclature of `CounterfactualEntity`: `value`/`feature` into `proposedValue`/`originalValue`
- Remove `Value` type
- Make counterfactual match unit test more explicit (using specific feature values)

danielezonca (approved):
Thanks for your changes, approving :+1:

danielezonca:
This PR cannot be merged until release pipeline is updated accordingly (because it depends on OptaPlanner). This means after 1.0

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterFactualScoreCalculator.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    private double predictionDistance(List<PredictionOutput> predictions, List<Output> goal) {
+        final List<Output> outputs = predictions.get(0).getOutputs();
+        double distance = 0.0;
+        if (outputs.size() != predictions.size()) {
+            throw new IllegalArgumentException("Prediction size must be equal to goal size");
+        }
+        for (int i = 0; i < outputs.size(); i++) {
+            final Output output = outputs.get(i);
+            distance += goal.get(i).getValue().asNumber() - output.getValue().asNumber();
```
r00ta, line 64: Hi @ruivieira , just surfing around to have a look :) I was wondering, shouldn't this be `distance += Math.abs(goal.get(i).getValue().asNumber() - output.getValue().asNumber())`?

ruivieira: Hi @r00ta, thanks and you are right of course! :) But since this PR was converted to draft, development moved to the [sandbox](https://github.com/kiegroup/trusty-ai-sandbox/tree/master/explainability/core/counterfactuals) and this is corrected (distance is [calculated](https://github.com/kiegroup/trusty-ai-sandbox/blob/9dd991da19461aed9abadb28d2254898f40fac10/explainability/core/counterfactuals/src/main/java/org/kie/kogito/explainability/local/counterfactual/entities/DoubleEntity.java#L147) feature-wise).

cc @danielezonca should I delete this PR so a fresh one can be made?

danielezonca: Fine for me to close it but wait to create a new one. I raised the point here https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/repo.20reordering.20-.20optaplanner.2Fkogito-apps/near/218547261 so I hope to get an answer soon

r00ta:
Hi @ruivieira , can this be closed now?

ruivieira:
> Hi @ruivieira , can this be closed now?

Hi @r00ta, yes definitely. This is now being done at https://github.com/kiegroup/kogito-apps/pull/602

PR #486, opened 2020-10-15
KOGITO-3602 - Add aggregate LIME global explainer

tteofili:
This PR adds the `AggregatedLimeExplainer` which aggregates local explanations (mean feature importance) from `LimeExplainer` across several `Predictions`.

see https://issues.redhat.com/browse/KOGITO-3602


explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,61 @@
...
+            Prediction prediction = new Prediction(input, model.predictAsync(List.of(input)).get().get(0));
+            saliencies.addAll(limeExplainer.explainAsync(prediction, model).get().values());
+        }
+        List<Saliency> merged = Saliency.merge(saliencies.toArray(new Saliency[0]));
+        Map<String, Saliency> meanSaliencyMap = new HashMap<>();
+        for (Saliency saliency: merged) {
+            meanSaliencyMap.put(saliency.getOutput().getName(), saliency);
+        }
+        return meanSaliencyMap;
+    }
```
danielezonca: Some notes:
- You can send to `predictAsync` the whole list of `PredictionInput` instead of one by one
- You should remove all `get()` and use `thenApply` / `thenCompose` to combine all the async calls
- Make number of samples as constructor param
- Make this method returns `CompletableFuture<Map<String, Saliency>>`

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainer.java
```diff
@@ -81,12 +82,12 @@ public PartialDependencePlotExplainer() {
...
 
-        List<FeatureDistribution> featureDistributions = dataDistribution.getFeatureDistributions();
+        List<FeatureDistribution> featureDistributions = dataDistribution.asFeatureDistributions();
         for (int featureIndex = 0; featureIndex < noOfFeatures; featureIndex++) {
             for (int outputIndex = 0; outputIndex < metadata.getOutputShape().getOutputs().size(); outputIndex++) {
                 // generate samples for the feature under analysis
-                double[] featureXSvalues = DataUtils.generateSamples(featureDistributions.get(featureIndex).getMin(),
-                                                                     featureDistributions.get(featureIndex).getMax(), seriesLength);
+                FeatureDistribution featureDistribution = featureDistributions.get(featureIndex);
+                double[] featureXSvalues = featureDistribution.sample(seriesLength).stream().map(Value::asNumber).map(Number::doubleValue).mapToDouble(d -> d).sorted().toArray();
```
danielezonca: Don't abuse of stream :) 
```suggestion
                double[] featureXSvalues = featureDistribution.sample(seriesLength).stream().mapToDouble(v -> v.asNumber().doubleValue()).sorted().toArray();
```
Why do you `sort` them? Is it needed?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainer.java
```diff
@@ -175,9 +176,8 @@ public PartialDependencePlotExplainer() {
     private double[][] generateDistributions(int noOfFeatures, List<FeatureDistribution> featureDistributions) {
         double[][] trainingData = new double[noOfFeatures][seriesLength];
         for (int i = 0; i < noOfFeatures; i++) {
-            double[] featureData = DataUtils.generateData(featureDistributions.get(i).getMean(),
-                                                          featureDistributions.get(i).getStdDev(), seriesLength,
-                                                          random);
+            double[] featureData = featureDistributions.get(i).sample(seriesLength).stream()
+                    .map(Value::asNumber).map(Number::doubleValue).mapToDouble(d -> d).sorted().toArray();
```
danielezonca: Same as above
```suggestion
            double[] featureData = featureDistributions.get(i).sample(seriesLength).stream()
                    .mapToDouble(v -> v.asNumber().doubleValue()).sorted().toArray();
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,61 @@
...
+    private final LimeExplainer limeExplainer;
+
+    public AggregatedLimeExplainer(LimeExplainer limeExplainer) {
+        this.limeExplainer = limeExplainer;
+    }
+
+    @Override
+    public Map<String, Saliency> explain(PredictionProvider model, PredictionProviderMetadata metadata)
+            throws InterruptedException, ExecutionException {
+        List<PredictionInput> inputs = metadata.getDataDistribution().sample(100);
```
ruivieira: Perhaps the sample size could be customisable (with a fallback default)?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,61 @@
...
+import org.kie.kogito.explainability.model.PredictionInput;
+import org.kie.kogito.explainability.model.PredictionProvider;
+import org.kie.kogito.explainability.model.PredictionProviderMetadata;
+import org.kie.kogito.explainability.model.Saliency;
+
+/**
+ * Global explainer aggregating LIME explanations over a number of inputs by reporting the mean feature importance for
+ * each feature.
+ */
+public class AggregatedLimeExplainer implements GlobalExplainer<Map<String, Saliency>> {
```
danielezonca: General comments to this global explainer class:
I think it should be useful to support a scenario where we already have `Prediction`s (aka both input and output) and we just need to apply `limeExplainer`. For example this is true if we have training/historical data.
What about extend `PredictionProviderMetadata` to support a similar scenario? It will probably require to change the `explain` method too (like create a method `CompletableFuture<List<Prediction>> extractPrediction(model, metadata)` that get prediction data if available or calculate them)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
```diff
@@ -0,0 +1,66 @@
...
+        return feature;
+    }
+
+    @Override
+    public Value<?> sample() {
+        return sample(1).get(0);
+    }
+
+    @Override
+    public List<Value<?>> sample(int sampleSize) {
```
ruivieira, line 67: `PredictionInputsDataDistribution` has a nice optimisation (returning all samples when `sampleSize > inputs.size()`) with the added benefit that it won't given an index error if we ask for more samples than inputs.
Could it be added here too?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
```diff
@@ -0,0 +1,63 @@
...
+        return feature;
+    }
+
+    @Override
+    public Value<?> sample() {
+        return sample(1).get(0);
+    }
+
+    @Override
+    public List<Value<?>> sample(int sampleSize) {
```
ruivieira, line 64: Same comment as in `NumericFeatureDistribution.sample()`

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Output.java
```diff
@@ -25,6 +27,13 @@
     private final double score;
     private final String name;
 
+    public Output(String name, Type type) {
+        this.name = name;
+        this.type = type;
+        this.value = new Value<>(null);
+        this.score = 0d;
```
danielezonca:
```suggestion
        this(name, type, new Value<>(null), 0d);
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
```diff
@@ -165,7 +167,7 @@ else if (upperBound > lowerBound) {
                 perturbationSize = perturbationContext.getRandom().ints(lowerBound, 1 + upperBound).findFirst().orElse(1);
             }
             if (perturbationSize > 0) {
-                int[] indexesToBePerturbed = perturbationContext.getRandom().ints(1, newFeatures.size())
+                int[] indexesToBePerturbed = perturbationContext.getRandom().ints(0, newFeatures.size())
```
danielezonca: What's the reason of this change?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
```diff
@@ -98,7 +97,7 @@ private void assertGraph(PartialDependenceGraph pdp) {
         for (int i = 0; i < pdp.getX().length; i++) {
             assertNotEquals(Double.NaN, pdp.getY()[i]);
             if (i > 0) {
-                assertTrue(pdp.getX()[i] > pdp.getX()[i - 1]);
+                assertTrue(pdp.getX()[i] >= pdp.getX()[i - 1]);
```
danielezonca, line 107: What's the reason of this change?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    public Map<String, Saliency> explain(PredictionProvider model, PredictionProviderMetadata metadata)
+            throws InterruptedException, ExecutionException {
+        List<PredictionInput> inputs = metadata.getDataDistribution().sample(100);
+        List<Saliency> saliencies = new ArrayList<>();
+        for (PredictionInput input : inputs) {
+            Prediction prediction = new Prediction(input, model.predictAsync(List.of(input)).get().get(0));
+            saliencies.addAll(limeExplainer.explainAsync(prediction, model).get().values());
+        }
+        List<Saliency> merged = Saliency.merge(saliencies.toArray(new Saliency[0]));
```
danielezonca: What about make `merge` accept a `List<Saliency>` so you can avoid to convert to array and then convert again to list?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
```diff
@@ -0,0 +1,80 @@
...
+    private final List<PredictionInput> inputs;
+
+    public PredictionInputsDataDistribution(List<PredictionInput> inputs) {
+        this.inputs = Collections.unmodifiableList(inputs);
+    }
+
+    @Override
+    public PredictionInput sample() {
+        List<PredictionInput> copy = new java.util.ArrayList<>(inputs);
+        Collections.shuffle(copy);
```
ruivieira: IIUC when considering a large list, perhaps get a random element directly by index could be better than shuffling the entire collection, since it's _O_(1) vs _O_(N)?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/SaliencyTest.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.jupiter.api.Test;
+import org.kie.kogito.explainability.TestUtils;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class SaliencyTest {
```
ruivieira, line 31: I don't know if it's worth to test when multiple features have the same importance, perhaps not...

ruivieira (commented):
@tteofili lGtm! I've just made some minor comments.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
danielezonca: Please add a log message in similar scenarios like "required 100 samples but only 10 available"

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
tteofili: yeah, definitely. Same as per `LimeExplainer`.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
tteofili: agree with both comments, thanks!

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
tteofili: agreed

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
tteofili: the reason for not having done that already is that I'm starting to dislike having `Random` objects all over the API.
But I guess the performance difference is worth the annoyance :-)

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/SaliencyTest.java
tteofili: good idea.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainer.java
tteofili: yes, because the related "plot" will then show the output trend as feature value increases (where it makes sense, e.g. with numbers).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainer.java
tteofili: here we can avoid the sorting actually.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
tteofili: I didn't have the fix for KOGITO-3564.
I can drop it now.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
```diff
@@ -0,0 +1,76 @@
...
+        if (sampleSize >= doubles.length) {
+            LOGGER.warn("required {} samples, but only {} are available", sampleSize, doubles.length);
```
jiripetrlik: Shouldn't we rather throw an exception or something. User may not notice this in logs.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
```diff
@@ -0,0 +1,86 @@
...
+        return copy.get(0);
+    }
+
+    @Override
+    public List<PredictionInput> sample(int sampleSize) {
+        if (sampleSize >= inputs.size()) {
+            LOGGER.warn("required {} samples, but only {} are available", sampleSize, inputs.size());
```
jiripetrlik: Again shouldn't we throw exception here?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainerTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+class AggregatedLimeExplainerTest {
+
+    @Test
+    void testExplain() throws ExecutionException, InterruptedException {
+        PredictionProvider evenFeatureModel = TestUtils.getSumSkipModel(1);
+        PredictionProviderMetadata metadata = new PredictionProviderMetadata() {
+            @Override
+            public DataDistribution getDataDistribution() {
+                return DataUtils.generateRandomDataDistribution(3, 100, new Random());
```
jiripetrlik: I think it would be better to use fixed seed here.

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
tteofili: I have adjusted the test to better reflect the semantics of what one can expect from PDP graphs there.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
tteofili: I've made it possible to sample a no. of samples bigger than the sample size (so generated data will contain duplicates).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
tteofili: I've made it possible to sample a no. of samples bigger than the sample size (so generated data will contain duplicates).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
tteofili: I think that's a good idea, generally speaking. I think it would be better to let the user decide whether using existing `Predictions` or generating new ones via `model.predictAsync`, so I would probably create a different interface for that purpose (and `AggregatedLimeExplainer` would implement that too).

explainability/explainability-core/pom.xml
```diff
@@ -19,6 +19,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
```
danielezonca: Where is it used? I cannot find any usage

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public Feature getFeature() {
+        return feature;
+    }
+
+    @Override
+    public Value<?> sample() {
+        return sample(1).get(0);
```
danielezonca: This could produce to index out of bound if `values` is an empty list

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public Feature getFeature() {
+        return feature;
+    }
+
+    @Override
+    public Value<?> sample() {
+        return sample(1).get(0);
```
danielezonca: Same as above

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,69 @@
...
+ * each feature.
+ */
+public class AggregatedLimeExplainer implements GlobalExplainer<CompletableFuture<Map<String, Saliency>>>,
+                                                ExistingPredictionsGlobalExplainer<CompletableFuture<Map<String, Saliency>>> {
+
+    private final LimeExplainer limeExplainer;
+    private final int sampleSize;
+
+    public AggregatedLimeExplainer(LimeExplainer limeExplainer) {
+        this(limeExplainer, 100);
```
r00ta: `private static final int DEFAULT_SAMPLE_SIZE = 100`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
```diff
@@ -0,0 +1,62 @@
...
+        List<Value<?>> copy = new java.util.ArrayList<>(values);
+        List<Value<?>> samples = new ArrayList<>(sampleSize);
+        for (int i = 0; i < sampleSize; i++) {
+            if (i % values.size() == 0) {
+                Collections.shuffle(copy);
+            }
+            samples.add(copy.get(i));
+        }
+        return samples;
+    }
```
r00ta:
```
public static List<Value<?>> sample(int sampleSize) {
        List<Value<?>> copy = new java.util.ArrayList<>(values);
        Collections.shuffle(copy);
        return copy.subList(0, sampleSize);
    }
```
?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/IndependentFeaturesDatatDistribution.java
```diff
@@ -0,0 +1,81 @@
...
+        Set<List<Feature>> cartesianProduct = Sets.cartesianProduct(featureEnumerations);
+        List<PredictionInput> inputs = new ArrayList<>(cartesianProduct.size());
+        for (List<Feature> features : cartesianProduct) {
+            inputs.add(new PredictionInput(features));
+        }
+        return inputs;
+    }
+
+    @Override
+    public List<FeatureDistribution> asFeatureDistributions() {
```
r00ta: Minor comment: `getFeatureDistributions`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public List<Value<?>> sample(int sampleSize) {
+        List<Double> copy = Arrays.stream(doubles).boxed().collect(Collectors.toList());
+        List<Value<?>> samples = new ArrayList<>(sampleSize);
+        for (int i = 0; i < sampleSize; i++) {
+            if (i % doubles.length == 0) {
+                Collections.shuffle(copy);
+            }
+            samples.add(new Value<>(copy.get(i % doubles.length)));
+        }
```
r00ta: Same as before

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public List<Value<?>> getAllSamples() {
+        return values;
```
r00ta: I see that in other implementations `getAllSample` shuffles the collection before it is returned. Should we do the same here?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public List<PredictionInput> sample(int sampleSize) {
+        List<PredictionInput> samples = new ArrayList<>(sampleSize);
+        for (int i = 0; i < sampleSize; i++) {
+            int index = random.nextInt(inputs.size() - 1);
+            index = index % inputs.size();
+            samples.add(inputs.get(index));
+        }
+        return samples;
```
r00ta: In the other implementations of `DataDistribution` the method `sample` is without resampling, the same element can't be picked more than one time (and `sampleSize` must be <= `thelist.size()`). Here the same method works with `sampleSize > inputs.size()` and is with resampling, is this intended?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Saliency.java
```diff
@@ -58,8 +65,41 @@ public Output getOutput() {
...
+            for (Map.Entry<Feature, List<FeatureImportance>> entry : collect.entrySet()) {
+                double meanScore = entry.getValue().stream().map(FeatureImportance::getScore).flatMapToDouble(DoubleStream::of).average().orElse(0);
+                result.add(new FeatureImportance(entry.getKey(), meanScore));
+            }
+            result.sort(Comparator.comparing(f -> f.getFeature().getName()));
+            finalResult.put(saliencyEntry.getKey().getName(), new Saliency(saliencyEntry.getKey(), result));
+        }
+        return finalResult;
+    }
+
```
r00ta, line 103: remove newline :D

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
```diff
@@ -54,7 +53,7 @@
     PredictionProviderMetadata metadata = new PredictionProviderMetadata() {
         @Override
         public DataDistribution getDataDistribution() {
-            return DataUtils.generateRandomDataDistribution(3, 100, new FakeRandom());
+            return DataUtils.generateRandomDataDistribution(3, 100, new Random());
```
r00ta: Should we set the seed here as well?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/GenericFeatureDistributionTest.java
```diff
@@ -0,0 +1,118 @@
...
+        GenericFeatureDistribution numericFeatureDistribution = new GenericFeatureDistribution(feature, values);
+        List<Value<?>> samples = numericFeatureDistribution.getAllSamples();
+        assertNotNull(samples);
+        assertEquals(10, samples.size());
+        for (Value<?> sample : samples) {
+            assertNotNull(sample);
+            assertNotNull(sample.getUnderlyingObject());
+            assertThat(sample.asNumber()).isBetween(0d, 10d);
+        }
+    }
```
r00ta: What do you think if we aggregate the three tests in one? At the end they are checking that `sample()`, `sample(int)` and `getAllSamples` returns something, we might do something like 
```
    @Test
    void testNumericGetAllSamples() {
        Feature feature = TestUtils.getMockedNumericFeature();
        double[] doubles = DataUtils.generateSamples(0, 10, 10);
        List<Value<?>> values = Arrays.stream(doubles).mapToObj(Value::new).collect(Collectors.toList());
        GenericFeatureDistribution numericFeatureDistribution = new GenericFeatureDistribution(feature, values);
        Assertions.assertEquals(10, numericFeatureDistribution.getAllSamples().size());
        Assertions.assertEquals(3, numericFeatureDistribution.sample(3).size());
        Assertions.assertTrue(numericFeatureDistribution.sample().asNumber()).isBetween(0d, 10d));
    }
```
Assuming that ` DataUtils.generateSamples` is already unit tested I don't think it's necessary to test that every element is between `0` and `10`. 

I would put a unit test for the case `numericFeatureDistribution.sample(11)` instead

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/GenericFeatureDistributionTest.java
```diff
@@ -0,0 +1,118 @@
...
+        GenericFeatureDistribution numericFeatureDistribution = new GenericFeatureDistribution(feature, values);
+        List<Value<?>> samples = numericFeatureDistribution.getAllSamples();
+        assertNotNull(samples);
+        assertEquals(26, samples.size());
+        for (Value<?> sample : samples) {
+            assertNotNull(sample);
+            assertNotNull(sample.getUnderlyingObject());
+            assertThat(sample.asString()).isBetween("a", "z");
+        }
+    }
```
r00ta: Same for this

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/IndependentFeaturesDatatDistributionTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.kie.kogito.explainability.utils.DataUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class IndependentFeaturesDatatDistributionTest {
+
+    @Test
+    void testSample() {
```
r00ta: Same as before

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/NumericFeatureDistributionTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.jupiter.api.Test;
+import org.kie.kogito.explainability.TestUtils;
+import org.kie.kogito.explainability.utils.DataUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.*;
+
+class NumericFeatureDistributionTest {
+
+    @Test
```
r00ta, line 32: same as before

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/PredictionInputsDataDistributionTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.kie.kogito.explainability.TestUtils;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class PredictionInputsDataDistributionTest {
+
+    @Test
+    void testSample() {
```
r00ta, line 31: same as before

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Saliency.java
```diff
@@ -58,8 +65,41 @@ public Output getOutput() {
...
-}
 
+    /**
+     * Merge saliencies so that they are aggregated by {@code Output} and the resulting list of associated
+     * {@code FeatureImportance}s is derived by calculating the mean score for each feature across all such saliencies.
+     *
+     * @param saliencies a collection of saliency maps
+     * @return a map of merged saliencies, one for each output appearing in the input saliencies
+     */
+    public static Map<String, Saliency> merge(Collection<Map<String, Saliency>> saliencies) {
```
r00ta, line 80: Is this unit tested?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Map<String, Saliency>> explain(PredictionProvider model, Collection<Prediction> predictions) {
+        return CompletableFuture.completedFuture(predictions)
+                .thenApply(p -> p.stream().map(prediction -> limeExplainer.explainAsync(prediction, model)) // extract saliency for each input
+                        .map(CompletableFuture::join) // aggregate all the saliencies
+                        .reduce(Collections.emptyMap(), (m1, m2) -> Saliency.merge(List.of(m1, m2)))); // merge all the saliencies together
+    }
+}
```
r00ta: Maybe I missed them, but do we have tests for those?

r00ta (commented):
Hi @tteofili , looks good! Some comments from my side

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
tteofili: sure

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/IndependentFeaturesDatatDistribution.java
tteofili: not a strong opinion, but I preferred _as_  because the `DataDistribution` interface might or might not be made up of independent feature distributions, so this method would actually transform the underlying data to such a `List`.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
tteofili: I like the suggested change, I need to also make it possible to return a `List` whose size is bigger than the original `values` `List` (hence a `subList` wouldn't always work here).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
tteofili: in `IndependentFeaturesDatatDistribution` this is not explicitly done because there's no fixed list of `PredictionInputs` to sample from, in there `sample()` is called `n` times regardless of the dimensionality of the underlying `FeatureDistributions` therefore repetition might happen and there's no need to explicitly re-sample.
I'll add a test in  `IndependentFeaturesDatatDistributionTest` to cover this scenario there as well.

tteofili: anyway in general it should be possible to draw as many samples as one wish, without requiring knowledge about the underlying data / feature distribution size. I'll double check this can be done in all the affected classes.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
ruivieira: @tteofili I'm probably wrong here, but if it's sampling with replacement would something like

```
List<Value<?>> samples = new Random()
   .ints(sampleSize, 0, values.size()-1)
   .mapToObj(i -> values.get(i))
   .collect(Collectors.toList());
```

work?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Saliency.java
tteofili: oki

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/global/pdp/PartialDependencePlotExplainerTest.java
tteofili: yes, good point.

ruivieira (commented):
@tteofili lgtm! Left a minor comment and nothing to add in addition to the other reviews.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
tteofili: all in all I think the best solution is to just call `sample()` `n` times avoiding any unnecessary guards on the sample size.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Saliency.java
tteofili: yes, in `SaliencyTest`

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
tteofili: yes, they are tested in `AggregatedLimeExplainerTest`.

tteofili:
> @tteofili I'm probably wrong here, but if it's sampling with replacement would something like
> List<Value<?>> samples = new Random()
>   .ints(sampleSize, 0, values.size()-1)
>   .mapToObj(i -> values.get(i))
>   .collect(Collectors.toList());
> work?

thanks @ruivieira I think it would work nicely.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
tteofili: yes, the behavior (shuffling) should be consistent across the implementations.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
tteofili: will do sampling with replacament in a common matter as per @ruivieira's soggestion

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
tteofili: agreed, I'll abstract inside an utility method in `DataUtils`, thanks!

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/GenericFeatureDistributionTest.java
tteofili: ok, I will unify them as suggested and create a new one for the "bigger than distribution size" case.

tteofili:
jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
```diff
@@ -408,4 +400,32 @@ private static void linearizeFeature(List<Feature> flattenedFeatures, Feature f)
...
+     *
+     * @param values     the list to sample from
+     * @param sampleSize the no. of samples to draw
+     * @param random     a random instance
+     * @param <T>        the type of values to sample
+     * @return a list of sampled values
+     */
+    public static <T> List<T> sampleWithReplacement(List<T> values, int sampleSize, Random random) {
+        return random
+                .ints(sampleSize, 0, values.size() - 1)
```
r00ta: If I'm not wrong, according to the doc if you do `value.size() - 1` the last value is never picked 
https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#ints-long-int-int-

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
```diff
@@ -0,0 +1,71 @@
...
+        return sample(1).get(0);
+    }
+
+    @Override
+    public List<Value<?>> sample(int sampleSize) {
+        return DataUtils.sampleWithReplacement(toValuesList(), sampleSize, random);
+    }
+
+    private List<Value<?>> toValuesList() {
+        return Arrays.stream(doubles).boxed().map(Value::new).collect(Collectors.toList());
```
r00ta: `Arrays.asList(doubles)` is enought isnt't?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/PredictionInputsDataDistributionTest.java
```diff
@@ -0,0 +1,90 @@
...
+        inputs.add(new PredictionInput(List.of(TestUtils.getMockedTextFeature("bar"))));
+        inputs.add(new PredictionInput(List.of(TestUtils.getMockedTextFeature("asd"))));
+        PredictionInputsDataDistribution predictionInputsDataDistribution = new PredictionInputsDataDistribution(inputs);
+        List<PredictionInput> samples = predictionInputsDataDistribution.sample(12);
+        assertNotNull(samples);
+        assertEquals(12, samples.size());
+        for (PredictionInput sample : samples) {
+            assertTrue(inputs.contains(sample));
+        }
+    }
```
r00ta, line 71: Are these two tests covering the same scenario?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
tteofili: correct, thanks for spotting it out!

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
tteofili: this is used to map a `double[]` into a `List<Value<Double>>`, it seems to me it's needed.

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/PredictionInputsDataDistributionTest.java
tteofili: the second one covers the case for the `sampleSize` being bigger than the actual underlying data distribution size.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
r00ta: my bad, your're right

tteofili:
jenkins retest this please

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/utils/DataUtilsTest.java
```diff
@@ -400,4 +401,14 @@ void testDropLinearizedFeature() {
...
     }
+
+    @Test
+    void testSampleWithReplacement() {
+        List<Double> values = Arrays.stream(DataUtils.generateData(0, 1, 100, random)).boxed().collect(Collectors.toList());
+        int sampleSize = random.nextInt(100);
+        List<Double> samples = DataUtils.sampleWithReplacement(values, sampleSize, random);
+        assertNotNull(samples);
+        assertEquals(sampleSize, samples.size());
+        assertThat(values).contains(samples.get(random.nextInt(sampleSize - 1)));
```
r00ta: It would be better to make this deterministic, wdyt? 
you might want to test if `sampleWithReplacement` works when `values.size()` < `sampleSize` for example

tteofili: sure, makes sense.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
```diff
@@ -0,0 +1,71 @@
...
+    public Value<?> sample() {
+        return sample(1).get(0);
+    }
+
+    @Override
+    public List<Value<?>> sample(int sampleSize) {
+        return DataUtils.sampleWithReplacement(toValuesList(doubles), sampleSize, random);
+    }
+
+    private List<Value<?>> toValuesList(double[] doubles) {
```
r00ta, line 71: I was wondering: since the array `doubles` is private and when it is used it's always converted to a list, is it better to convert it to a `List<Value<?>>` in the constructor so that it's done only once and not repeated every time a method is called?

r00ta (commented):
Last comment from my side

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
tteofili: +1

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,71 @@
...
+                .thenApply(os -> DataUtils.getPredictions(inputs, os)) // generate predictions from inputs and outputs
+                .thenCompose(ps -> explain(model, ps)); // explain predictions
+    }
+
+    @Override
+    public CompletableFuture<Map<String, Saliency>> explain(PredictionProvider model, Collection<Prediction> predictions) {
+        return CompletableFuture.completedFuture(predictions)
+                .thenApply(p -> p.stream().map(prediction -> limeExplainer.explainAsync(prediction, model)) // extract saliency for each input
+                        .map(CompletableFuture::join) // aggregate all the saliencies
+                        .reduce(Collections.emptyMap(), (m1, m2) -> Saliency.merge(List.of(m1, m2)))); // merge all the saliencies together
```
danielezonca, line 59: Please create a ticket to review this code, we should be able to compose requests without the `CompletableFuture::join`

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
```diff
@@ -0,0 +1,71 @@
...
+import org.kie.kogito.explainability.utils.DataUtils;
+
+/**
+ * Global explainer aggregating LIME explanations over a number of inputs by reporting the mean feature importance for
+ * each feature.
+ */
+public class AggregatedLimeExplainer implements GlobalExplainer<CompletableFuture<Map<String, Saliency>>>,
+                                                ExistingPredictionsGlobalExplainer<CompletableFuture<Map<String, Saliency>>> {
+
+    private static final int DEFAULT_SAMPLE_SIZE = 100;
```
danielezonca: What about review it to use `LimeConfig`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/ExistingPredictionsGlobalExplainer.java
```diff
@@ -0,0 +1,39 @@
...
+import org.kie.kogito.explainability.model.Prediction;
+import org.kie.kogito.explainability.model.PredictionProvider;
+
+/**
+ * A global explainability method that leverages existing predictions (e.g. from historical data) instead of generating
+ * new ones for the purpose of explainability.
+ *
+ * @param <T> the type of global explanation generated
+ */
+public interface ExistingPredictionsGlobalExplainer<T> {
```
danielezonca: What about merge this interface with `GlobalExplainer` and have two different methods `explainFromPrediction` and `explainFromMetadata`?
You can also `default` implement `explainFromMetadata`
```java
    default public T explainFromMetadata(PredictionProvider model, PredictionProviderMetadata metadata) {
        List<PredictionInput> inputs = metadata.getDataDistribution().sample(sampleSize); // sample inputs from the data distribution

        return model.predictAsync(inputs) // execute the model on the inputs
                .thenApply(os -> DataUtils.getPredictions(inputs, os)) // generate predictions from inputs and outputs
                .thenCompose(ps -> explain(model, ps)); // explain predictions
    }
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PredictionInputsDataDistribution.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public PredictionInputsDataDistribution(List<PredictionInput> inputs, Random random) {
+        this.inputs = Collections.unmodifiableList(inputs);
+        this.random = random;
+    }
+
+    @Override
+    public PredictionInput sample() {
+        return sample(1).get(0);
```
danielezonca: Please verify if this could throw an index out of bound

explainability/explainability-core/pom.xml
tteofili: leftover from the initial draft, will remove it, thanks for spotting it.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/GenericFeatureDistribution.java
tteofili: correct, thanks, I'll fix it.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/NumericFeatureDistribution.java
tteofili: correct, thanks, I'll fix it.

explainability/explainability-core/pom.xml
tteofili: ah no! it's actually used in `IndependentFeatureDistribution` to generate the cartesian product of the feature values.

jiripetrlik (approved):

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/IndependentFeaturesDatatDistribution.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import com.google.common.collect.Sets;
+
+/**
+ * Data distribution based on list of {@code FeatureDistributions}.
+ */
+public class IndependentFeaturesDatatDistribution implements DataDistribution {
```
ruivieira: @tteofili sorry I've missed this in the first review, is `IndependentFeaturesDatatDistribution` a typo? (i.e. should be `IndependentFeaturesDataDistribution`).

```suggestion
public class IndependentFeaturesDataDistribution implements DataDistribution {
```

tteofili: thanks a lot Rui for having spotted this :+1:

explainability/explainability-core/pom.xml
tteofili: I'm dropping guava as it's only used here, hence reimplementing cartesian product.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
tteofili: agreed, it'll rely on the underlying `LimeExplainer#getLimeConfig()` settings.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/ExistingPredictionsGlobalExplainer.java
tteofili: agree on unifying the two interfaces, for now I'd not add a `default` method since we only have two impls, also the `PDPExplainer` works explicitly with feature distributions independently and would rather not make use of the existing outputs (as the inputs need to be "crafted" specifically and outputs need to be recalculated).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/global/lime/AggregatedLimeExplainer.java
tteofili: I've created https://issues.redhat.com/browse/FAI-334 to track this.

danielezonca (approved):
Thanks for the PR, approving :+1:

ruivieira (approved):
lGtm!

r00ta (approved):
Thanks! :+1:

PR #495, opened 2020-10-21
KOGITO-3118 Infinispan 11 upgrade

cristianonicolai:
Assembly:
- https://github.com/kiegroup/kogito-runtimes/pull/791
- https://github.com/kiegroup/kogito-apps/pull/495
- https://github.com/kiegroup/kogito-examples/pull/407
- https://github.com/kiegroup/kie-docs/pull/2952

### DIFF
--- a/data-index/data-index-service/data-index-service-infinispan/src/main/resources/META-INF/kogito-cache-default.xml
+++ b/data-index/data-index-service/data-index-service-infinispan/src/main/resources/META-INF/kogito-cache-default.xml
@@ -1,11 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<infinispan
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="urn:infinispan:config:10.0 https://infinispan.org/schemas/infinispan-config-10.0.xsd"
-  xmlns="urn:infinispan:config:10.0">
+<infinispan xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="urn:infinispan:config:11.0 https://infinispan.org/schemas/infinispan-config-11.0.xsd"
+            xmlns="urn:infinispan:config:11.0">
   <cache-container statistics="true" shutdown-hook="DEFAULT">
     <local-cache name="${cache_name}">
       <indexing index="ALL">
+        <!-- Workaround for https://issues.redhat.com/browse/ISPN-12203 -->
+        <property name="hibernate.search.index_uninverting_allowed">true</property>
         <property name="default.directory_provider">local-heap</property>
       </indexing>
     </local-cache>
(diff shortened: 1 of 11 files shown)


cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

radtriste (approved):

ricardozanini (approved):

tiagodolphine (approved):

data-index/data-index-service/data-index-service-infinispan/src/main/resources/META-INF/kogito-cache-default.xml
```diff
@@ -6,6 +6,8 @@
   <cache-container statistics="true" shutdown-hook="DEFAULT">
```
MarianMacik, line 5: Just an idea, shouldn't we upgrade also schema version in the xml? Currently it is 10.0.

Side question - this XML is used as a basic configuration for Infinispan instance that Data Index uses? I wonder why we don't have something similar for kogito runtime service by default, probably because we need more flexibility from the user in Kogito Runtime??

MarianMacik (approved):

data-index/data-index-service/data-index-service-infinispan/src/main/resources/META-INF/kogito-cache-default.xml
cristianonicolai: @MarianMacik just updated the schema version. As for the configuration XML, correct that is used for these different services, including data index. The reason is there is because indexing must be enabled, otherwise it restricts some of our searching capabilities. As opposed to runtimes, where we use it as a simple key-value store, so default config from Infinispan is fine. We can look into doing something like this for runtimes if there is some specific need.

cristianonicolai:
@danielezonca any impact of this in the trusty service?

danielezonca (approved):

danielezonca (changes requested):
Sorry I forgot this file
https://github.com/kiegroup/kogito-apps/blob/master/trusty/trusty-storage/trusty-storage-infinispan/src/main/resources/META-INF/decision.proto
I think it should be updated too @kostola can you please double check?

kostola:
The [Upgrading Infinispan 11.0 document](https://infinispan.org/docs/stable/titles/upgrading/upgrading.html) says:

> It is now required to annotate a field that requires sorting with @SortableField, both for embedded and remote queries.

In the Trusty storage protobuf definition, the only field we're currently sorting by (`Decision.executionTimestamp`) is already annotated with `@SortableField` ([link](https://github.com/kiegroup/kogito-apps/blob/0f49bdabc5a1b10fd91e8342f0e9f6c06e6032d0/trusty/trusty-storage/trusty-storage-infinispan/src/main/resources/META-INF/decision.proto#L18)).

Imho that file is ok at the moment. We may add the annotation to other fields in the future in case we need to sort by them.

@cristianonicolai and @MarianMacik what do you think?

danielezonca (approved):
@kostola 
Ok thanks for the verification, I think we are fine to proceed :+1:

r00ta (approved):

kostola (approved):

PR #512, opened 2020-10-30
KOGITO-3328 Allow using not operator to negate a filter

cristianonicolai:


cristianonicolai:
@Sara4994 this seems to break the domain explorer, could you please take a look.

Sara4994:
> @Sara4994 this seems to break the domain explorer, could you please take a look.

@cristianonicolai sure

persistence-commons/persistence-commons-infinispan/src/main/java/org/kie/kogito/persistence/infinispan/query/InfinispanQuery.java
```diff
@@ -138,6 +138,8 @@ protected InfinispanQuery(QueryFactory qf, String rootType) {
                     return getRecursiveString(filter, OR);
                 case AND:
                     return getRecursiveString(filter, AND);
+                case NOT:
+                    return format("not %s", filterStringFunction().apply((AttributeFilter<?>) filter.getValue()));
```
jstastny-cz, line 142: filter can be even complex expression, right? Do all complex expressions (containing OR, AND) come with surrounding parentheses? Looking at getRecursiveString they probably do, could you confirm?

jstastny-cz (approved):

persistence-commons/persistence-commons-infinispan/src/main/java/org/kie/kogito/persistence/infinispan/query/InfinispanQuery.java
cristianonicolai: @jstastny-cz at this point the support for _not_ operator is for single filter, not complex ones, that need might be something we can revisit in the future if there are use cases for that. For now, I went with a more basic approach that covers the use case needed by the task console.

pefernan: I guess that's fine, if you need grouping conditions inside of a not, you may use an and/or that will add the parenthesis..

Sara4994 (approved):

pefernan:
@cristianonicolai code looks well but I cannot tested locally!!

nmirasch (approved):

pefernan (approved):

PR #521, opened 2020-11-04
KOGITO-3252: management console configuration to use the process SVG …

nmirasch:
…addon

Jira: https://issues.redhat.com/browse/KOGITO-3252
Depends on https://github.com/kiegroup/kogito-runtimes/pull/846

### DIFF
--- a/management-console/pom.xml
+++ b/management-console/pom.xml
@@ -127,6 +127,21 @@
       <build>
         <plugins>
           <plugin>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-maven-plugin</artifactId>
+            <version>${version.io.quarkus}</version>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>native-image</goal>
+                </goals>
+                <configuration>
+                  <enableHttpUrlHandler>true</enableHttpUrlHandler>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
             <artifactId>maven-failsafe-plugin</artifactId>
             <executions>
               <execution>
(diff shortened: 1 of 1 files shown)


management-console/pom.xml
```diff
@@ -107,6 +112,22 @@
...
             </configuration>
           </execution>
+          <execution>
+            <id>copy-svg-resources</id>
+            <phase>process-resources</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${basedir}/target/classes/META-INF/processSVG/</outputDirectory>
```
cristianonicolai: @nmirasch I though we were only going to point to a external folder in the management console. Not sure we should package the svgs in the jar, as the users wouldnt be able to do the same as we provide the generated app. Not sure if Im missing something.

nmirasch: yes, that was and currently is  the idea. I have revert this  resource copy into the jar. I have added some test and resources with  the needed svg at least for the kogito-travel-agency.

cristianonicolai (approved):

nmirasch:
@cristianonicolai @jstastny-cz Just kept  the process-svg-addon dependency and added the native-imagen plugin

cristianonicolai (approved):

nmirasch:
jenkins please retest this

management-console/pom.xml
```diff
@@ -121,6 +125,21 @@
       </activation>
       <build>
         <plugins>
+          <plugin>
```
jstastny-cz, line 129: Do we actually need it? it's already defined in the common section in project / build, so just setting the property `<quarkus.package.type>native</quarkus.package.type>` below should be enough.

jstastny-cz: it's not enough :-)

jstastny-cz (approved):

PR #530, opened 2020-11-09
KOGITO-3763 - Improved numeric feature handling

tteofili:
This PR includes the following changes which target higher stability explanations when numeric features are involved:
- less strict filtering for sparse encoding of numeric features (this is useful to address stability for negative impact on regression problems) 
- sample weighting is stricter, this means that only very similar samples will receive high sample weights (this generally generates better fitting linear models wrt the prediction to explain)
- reduced standard deviation to 1% of the original value when perturbing a number (so a new "perturbed" number is generated from a normal distribution centered on the original feature value with stdDev equals to 1% of the original value)
- reduced variance when perturbing `int` values
- included the original feature value in min / max scaling (which is performed during sparse encoding), this fixes cases where all the original value doesn't fall in the value range of the sampled perturbed values (which is rare but can still happen)
- small bug in feature perturbation size (a missing +1 in `DataUtils#perturbFeature`)
- introduced local saliency stability metric: this measures the fraction of times that the top K positive/negative features are exactly the same across a no. of explaimability runs on the same `Prediction`
- added missing `Math.abs` in sparse encoding check that the generated value is within the required threshold  

See https://issues.redhat.com/browse/KOGITO-3763


tteofili:
jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/SampleWeighter.java
```diff
@@ -27,7 +27,7 @@
  */
 class SampleWeighter {
 
-    private static final double SIGMA = 0.75;
+    private static final double SIGMA = 0.675;
```
danielezonca, line 31: Is this based on some euristic/rule?

tteofili:
jenkins please retest this

tteofili:
jenkins please retest this

tteofili:
jenkins please retest this

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
```diff
@@ -162,7 +162,7 @@ static Feature doubleToFeature(double d) {
                 perturbationSize = lowerBound;
             }
             else if (upperBound > lowerBound) {
-                perturbationSize = perturbationContext.getRandom().ints(1, lowerBound, upperBound).findFirst().orElse(1);
+                perturbationSize = perturbationContext.getRandom().ints(1, lowerBound, 1 + upperBound).findFirst().orElse(1);
```
danielezonca, line 171: Can you add a test to cover this branch?

danielezonca:
jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -164,20 +165,28 @@
...
             // find maximum and minimum values
-            double[] doubles = new double[values.length];
+            double[] doubles = new double[values.length + 1];
             int i = 0;
             for (Value<?> v : values) {
                 doubles[i] = v.asNumber();
                 i++;
             }
             double originalValue = target.asNumber();
+            doubles[i] = originalValue; // include target number in feature scaling
```
danielezonca: Why are you adding `originalValue` to feature scaling? Can you please clarify?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -164,20 +165,28 @@
         @Override
         public List<double[]> encode(Value<?> target, Value<?>... values) {
             // find maximum and minimum values
-            double[] doubles = new double[values.length];
+            double[] doubles = new double[values.length + 1];
             int i = 0;
```
danielezonca:
```suggestion
            int valueIndex = 0;
```

tteofili:
jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -132,4 +139,109 @@ public static double classificationFidelity(List<Pair<Saliency, Prediction>> pai
...
+        Map<String, List<Saliency>> saliencies = getMultipleSaliencies(model, prediction, saliencyLocalExplainer, runs);
+
+        LocalSaliencyStability saliencyStability = new LocalSaliencyStability(saliencies.keySet());
+        // for each decision, calculate the stability rate for the top k important feature set, for each k < topK
+        for (Map.Entry<String, List<Saliency>> entry : saliencies.entrySet()) {
+            for (int k = 1; k <= topK; k++) {
+                String decision = entry.getKey();
+                List<Saliency> perDecisionSaliencies = entry.getValue();
+
+                int finalK = k;
```
r00ta, line 167: do we need `finalK`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -132,4 +139,109 @@ public static double classificationFidelity(List<Pair<Saliency, Prediction>> pai
...
+        long max = 0L;
+        Pair<List<String>, Long> topK = Pair.of(Collections.emptyList(), 0L);
+        for (Map.Entry<List<String>, Long> entry : collect.entrySet()) {
+            if (entry.getValue() >= max) {
+                topK = Pair.of(entry.getKey(), entry.getValue());
+                max = entry.getValue();
+            }
+        }
+        return topK;
+    }
```
r00ta: If I've got it properly this can be replaced by something like
```
Map.Entry maxEntry = Collections.max(collect.entrySet(), Map.Entry.comparingByValue());
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/LocalSaliencyStability.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Local {@code Saliency} stability evaluation result.
+ */
+public class LocalSaliencyStability {
```
r00ta, line 27: Do we need `public` or can we keep it internal?

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
```diff
@@ -86,17 +89,30 @@ void testPMMLRegression() throws Exception {
...
                 assertEquals(1d, v);
             }
+            int topK = 1;
+            LocalSaliencyStability stability = ExplainabilityMetrics.getLocalSaliencyStability(model, prediction, limeExplainer, topK, 10);
+            for (int i = 1; i <= topK; i++) {
+                for (String decision : stability.getDecisions()) {
+                    double positiveStabilityScore = stability.getPositiveStabilityScore(decision, i);
+                    double negativeStabilityScore = stability.getNegativeStabilityScore(decision, i);
+                    assertThat(positiveStabilityScore).isGreaterThanOrEqualTo(0.5);
+                    assertThat(negativeStabilityScore).isGreaterThanOrEqualTo(0.5);
```
r00ta: I have some questions around the `getPositiveStabilityScore` and `getNegativeStabilityScore`, what do they represent? Since they are rates how can they be both greater than `0.5`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -132,4 +139,109 @@ public static double classificationFidelity(List<Pair<Saliency, Prediction>> pai
...
+     * Such an evaluation is intended to measure how stable the explanations are in terms of "are the top k most important
+     * positive/negative features always the same for a single prediction?".
+     *
+     * @param model                  a model to explain
+     * @param prediction             the prediction on which explanation stability will be evaluated
+     * @param saliencyLocalExplainer a local saliency explainer
+     * @param topK                   no. of top k positive/negative features for which stability report will be generated
+     * @return a report about stability of all the decisions/predictions (and for each {@code k < topK})
+     */
+    public static LocalSaliencyStability getLocalSaliencyStability(PredictionProvider model, Prediction prediction,
```
r00ta, line 154: Do we need `public` or can we keep it internal?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
tteofili: it seems so, as we use `k` in the stream, but `k` value changes across iterations `for (k=0; k < ...`

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
tteofili: much better, thanks!

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
tteofili: being a metric, I supposed it would have made sense to keep it public, however at this stage it's not "used" by other components, so we can do both.

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
```diff
@@ -86,17 +89,30 @@ void testPMMLRegression() throws Exception {
                 }
                 return outputs;
             });
-            PredictionOutput output = model.predictAsync(List.of(input))
-                    .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit())
-                    .get(0);
+            List<PredictionOutput> predictionOutputs = model.predictAsync(List.of(input))
+                    .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit());
+            assertNotNull(predictionOutputs);
```
jiripetrlik: I think it would be good to use assertj also here. It provides much better messages when test fails.

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
```diff
@@ -86,17 +89,30 @@ void testPMMLRegression() throws Exception {
...
+            PredictionOutput output = predictionOutputs.get(0);
+            assertNotNull(output);
             Prediction prediction = new Prediction(input, output);
             Map<String, Saliency> saliencyMap = limeExplainer.explainAsync(prediction, model)
                     .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit());
             for (Saliency saliency : saliencyMap.values()) {
                 assertNotNull(saliency);
-                double v = ExplainabilityMetrics.impactScore(model, prediction, saliency.getPositiveFeatures(2));
+                double v = ExplainabilityMetrics.impactScore(model, prediction, saliency.getTopFeatures(2));
                 assertEquals(1d, v);
```
jiripetrlik: Please use assertj.

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/model/TypeTest.java
```diff
@@ -294,6 +294,27 @@ void testEncode(Type type) {
...
+            Value<?>[] values = new Value<?>[6];
+            for (int i = 0; i < values.length / 2; i++) {
+                values[i] = new Value<>(target.asNumber() + target.asNumber() * (1 + i) / 100d);
+                values[values.length - 1 - i] = new Value<>(target.asNumber() - target.asNumber() * (1 + i) / 100d);
+            }
+            List<double[]> vectors = Type.NUMBER.encode(target, values);
+            assertNotNull(vectors);
+            assertEquals(values.length, vectors.size());
+            for (int i = 0; i < vectors.size() / 2; i++) {
+                assertEquals(vectors.get(i)[0], vectors.get(vectors.size() - 1 - i)[0]);
```
jiripetrlik: Please use assertj also here.

jiripetrlik (commented):
Hello @tteofili ,
thank you for the PR. It looks good to me and it is nice to see that you started to use assertj. I think it provides much better explanation than junit. However, would it be possible to use assertj for all new asserts?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
r00ta: Sorry, I did not see it was used in the lambda

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
r00ta: np, if it was intented to be a public api let's keep it as it is. Just wanted to be sure it was not just for our internal testing

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/LocalSaliencyStability.java
tteofili: same reasoning as `ExplainabilityMetrics#getLocalSaliencyStability`, being this the result of the evaluation, it should be public for evaluators to consume it.

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
tteofili: data about positive and negative stability are kept separate because we want to separately check that the top `k` negative features are always the same as well as the top `k` positive features are always the same; it can happen that the negatives are stable whereas this is not true for the positive features.
However I think _rate_ is wrong here, _ratio_ is more accurate (it's basically the number of times the most frequent positive/negative features appear divided by the number of explanations drawn for that decision).

tteofili:
jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -130,7 +130,7 @@
     },
 
     NUMBER("number") {
-        private static final double CLUSTER_THRESHOLD = 1e-3;
+        private static final double CLUSTER_THRESHOLD = 1e-1;
```
danielezonca, line 133: Magic number?

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
```diff
@@ -131,16 +144,29 @@ void testPMMLRegressionCategorical() throws Exception {
...
+        }
+        int topK = 1;
+        LocalSaliencyStability stability = ExplainabilityMetrics.getLocalSaliencyStability(model, prediction, limeExplainer, topK, 10);
+        for (int i = 1; i <= topK; i++) {
+            for (String decision : stability.getDecisions()) {
+                double positiveStabilityScore = stability.getPositiveStabilityScore(decision, i);
+                double negativeStabilityScore = stability.getNegativeStabilityScore(decision, i);
+                assertThat(positiveStabilityScore).isGreaterThanOrEqualTo(0.5);
+                assertThat(negativeStabilityScore).isGreaterThanOrEqualTo(0.5);
+            }
```
danielezonca: I see the same "pattern" in all the tests (not only PMML but all the others too) with only number of samples, topK and positive/negative stability score as difference.
Does it make sense to move it to a common method?
Are all of them magic numbers? Like why 300/500 samples? Why the last assertion is >= 0.5/0.8?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -164,20 +165,28 @@
...
-            List<Double> encodedValues = DoubleStream.of(doubles).map(d -> (d - min) / (max - min))
-                    .map(d -> Double.isNaN(d) ? 1 : d).boxed().map(d -> DataUtils.gaussianKernel(d, 0, 1))
-                    .map(d -> (d - threshold < CLUSTER_THRESHOLD) ? 1d : 0d).collect(Collectors.toList());
+
+            // feature scaling
+            List<Double> scaledValues = DoubleStream.of(doubles).map(d -> (d - min) / (max - min)).boxed().collect(Collectors.toList());
+            double scaledOriginalValue = scaledValues.remove(i); // extract the scaled original value (it must not appear in encoded values)
+
+            // kernel based clustering
+            double sigma = 1;
```
danielezonca, line 184: Why fixed value?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
```diff
@@ -132,4 +139,102 @@ public static double classificationFidelity(List<Pair<Saliency, Prediction>> pai
...
+                    } else {
+                        saliencies.put(saliencyEntry.getKey(), List.of(saliencyEntry.getValue()));
+                    }
+                } else {
+                    LOGGER.warn("skipping empty / zero saliency for {}", saliencyEntry.getKey());
+                    skipped++;
+                }
+            }
+        }
+        LOGGER.debug("skipped {} useless saliencies", skipped);
```
danielezonca: What about make both of them `warn` or `debug`?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
tteofili: as per issue description: 

> included the original feature value in min / max scaling (which is performed during sparse encoding), this fixes cases where the original value doesn't fall in the value range of the sampled perturbed values (which is rare but can still happen)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
tteofili: ok

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/SampleWeighter.java
tteofili: this softens the proximity function "strictness", some of the changes in this PR are somewhat correlated as we have improved the locality of the generated samples (in `Type.NUMBER`), so that the algo benefits from having less strict proximity function (here the width controls how strict the similarity between two vectors is) and similarity threshold (the `CLUSTER_THRESHOLD` in `Type.NUMBER`).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
tteofili: see comment above for `SampleWeighter.SIGMA`

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
tteofili: Sure it makes sense to have a common method in `TestUtils`, I'll do that.
For the specific settings: I've set a minimum expected stability rate to 0.5 for the models having more than 5 features (when linearized) including ones of `Type.NUMBER` since they're harder. 
I've set a high minimum (0.8) for _OpenNLP_ ITs since the input only consists of 4 words that can only be set to empty String when perturbed, resulting in a low no. of possible perturbed inputs and hence an higher stability.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/ExplainabilityMetrics.java
tteofili: ok

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
tteofili: this has not changed, previously the gaussian filter was centered in 0 with a stdDev of 1, now its centered around the original value with the same stdDev (as per PR description).
this way the encoding mechanism sets to 1 those scaled numbers that fall close to the center of the gaussian (= the original feature value).

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/DataUtils.java
tteofili: actually this is already covered in `DataUtils#testPerturbDropNumericTwo` and `DataUtils#testPerturbDropNumericThree`.
we can't control the actual number being generated unless we induce the generation with a custom `Random` test-impl (e.g. the `FakeRandom` class we had).

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlLimeExplainerTest.java
tteofili: a practical problem with such a test method is that we need to inject a test classified version of `explainability-core` into the IT modules, which I think it's a bit overkill given the fact that we just need a few lines of code to check the expected stability scores.

tteofili: I've moved the boilerplate to validate local saliency stability in a `ValidationUtils#validateLocalSaliencyStability` method so that it can be re-used easily in ITs as well.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -150,12 +150,13 @@
             // sample from a standard normal distribution and center around feature value
             double normalDistributionSample = perturbationContext.getRandom().nextGaussian();
             if (originalFeatureValue != 0d) {
-                normalDistributionSample = normalDistributionSample * originalFeatureValue + originalFeatureValue;
+                double stDev = originalFeatureValue * 0.01; // set std dev at 1% of feature value
```
jiripetrlik, line 153: Would it be possible make "0.01" a constant and move it somewhere up to be more visible. I think it is better to have constants visible rather than hidden inside mothods etc.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -150,12 +150,13 @@
...
+                normalDistributionSample = normalDistributionSample * originalFeatureValue + stDev;
             }
             if (intValue) {
                 normalDistributionSample = (int) normalDistributionSample;
                 if (normalDistributionSample == originalFeatureValue) {
-                    normalDistributionSample = (int) normalDistributionSample * 10d;
+                    normalDistributionSample = (int) normalDistributionSample + 1d;
```
jiripetrlik, line 159: Same here. Can we make "1d" a constant and move it to more visible place?

explainability/explainability-integrationtests/explainability-integrationtests-opennlp/pom.xml
```diff
@@ -25,6 +25,11 @@
       <artifactId>quarkus-junit5</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
jiripetrlik, line 28: +1 for using assertj instead of junit asserts.

jiripetrlik (commented):
I have few minor comments. It also seems to me that we are tuning various parameters like SIGMA, CLUSTER_THRESHOLD etc. Can we make these parameters somehow configurable with reasonable default? To be honest I have only a limited knowledge about these algorithms, but I would guess that user may sometimes need a different settings. What do you think?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
tteofili: I am thinking of placing all those parameters inside `LimeConfig` so that they are centralized, configurable and visible frome one place. I will make a new PR for that.

tteofili: here's the Jira issue: https://issues.redhat.com/browse/FAI-342

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
tteofili: I'd resolve this also as part of https://issues.redhat.com/browse/FAI-342

danielezonca (approved):
LGTM well done :+1:

jiripetrlik (approved):
@tteofili 
I agree to solve the configuration of parameters in a separate PR. Thank you for creating of https://issues.redhat.com/browse/FAI-342.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
```diff
@@ -164,20 +165,28 @@
...
+
+            // feature scaling
+            List<Double> scaledValues = DoubleStream.of(doubles).map(d -> (d - min) / (max - min)).boxed().collect(Collectors.toList());
+            double scaledOriginalValue = scaledValues.remove(valueIndex); // extract the scaled original value (it must not appear in encoded values)
+
+            // kernel based clustering
+            double sigma = 1;
+            double threshold = DataUtils.gaussianKernel(scaledOriginalValue, scaledOriginalValue, sigma);
+            List<Double> clusteredValues = scaledValues.stream()
+                    .map(d -> Double.isNaN(d) ? 0 : d).map(d -> DataUtils.gaussianKernel(d, scaledOriginalValue, sigma)).collect(Collectors.toList());
```
r00ta: just wondering.. if `Double.isNan(d)` is true the previous call at line 180
```
            List<Double> scaledValues = DoubleStream.of(doubles).map(d -> (d - min) / (max - min)).boxed().collect(Collectors.toList());
```
should have crashed right?

explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/PrequalificationDmnLimeExplainerTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class PrequalificationDmnLimeExplainerTest {
+
+    @Test
+    void testPrequalificationDMNExplanation1() throws ExecutionException, InterruptedException, TimeoutException {
```
r00ta:
```suggestion
    void testPrequalificationDMNExplanation() throws ExecutionException, InterruptedException, TimeoutException {
```

r00ta: And rename `Prequalification-1` file?

r00ta (approved):
Just a couple of minor things :+1:

explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/PrequalificationDmnLimeExplainerTest.java
tteofili: sure, thanks :)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/Type.java
tteofili: good catch, it doesn't actually catch but the fact that there's a `NaN` makes all the values become `NaN` too in the feature scaling.
I've fixed it and added a test to cover this case.

PR #553, opened 2020-11-24
KOGITO-3877 - Parse date expression when using a query without variables

cristianonicolai:


nmirasch (approved):

wmedvede (approved):
Was tested locally in both query variants (with and without variables) and worked fine.
LGTM
Thanks!

PR #570, opened 2020-12-15
PLANNER-2247: Task events consumption / data index subscriptions

wmedvede:


wmedvede:
Hi @ge0ffrey , @rsynek  @cristianonicolai , would you mind review this PR?
Thanks

For the record, sonar complains about duplication are because there are two different pojos with similar field, etc,, but it's ok for this use case.

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEvent.java
```diff
@@ -0,0 +1,262 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class UserTaskEvent {
```
rsynek, line 29: While I favor using a separate classes for each purpose, maybe the duplication `SonarCloud` complains about could be reduced by introducing an abstract parent class?

task-assigning/task-assigning-process-service-client/src/test/java/org/kie/kogito/taskassigning/messaging/ReactiveMessagingEventConsumerTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+class ReactiveMessagingEventConsumerTest {
+
+    @Test
+    @Timeout(2)
```
rsynek: Nitpicking: in the optaplanner repository, we usually use 10 seconds for test timeouts. Even if the test runs on desperately slow machine, there is enough buffer for it to pass. It's just a timeout, so it will take the 10 seconds only if there is an issue and then we know it's worth investigating.

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEventMessage.java
```diff
@@ -0,0 +1,65 @@
...
+
+package org.kie.kogito.taskassigning.messaging;
+
+import java.net.URI;
+import java.time.ZonedDateTime;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
```
rsynek, line 25: The `RegisterForReflection` annotation is there because of the (de)serialization  to/from JSON?

task-assigning/task-assigning-process-service-client/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -1,8 +1,17 @@
...
+
+quarkus.kafka.bootstrap-servers=localhost:9092
+
+mp.messaging.incoming.kogito-usertaskinstances-events.connector=smallrye-kafka
+mp.messaging.incoming.kogito-usertaskinstances-events.topic=kogito-usertaskinstances-events
+mp.messaging.incoming.kogito-usertaskinstances-events.value.deserializer=org.kie.kogito.taskassigning.messaging.UserTaskEventDeserializer
+mp.messaging.incoming.kogito-usertaskinstances-events.group.id=kogito-task-assigning-service
+mp.messaging.incoming.kogito-usertaskinstances-events.enable.auto.commit=false
+mp.messaging.incoming.kogito-usertaskinstances-events.auto.offset.reset=earliest
+mp.messaging.incoming.kogito-usertaskinstances-events.isolation.level=read_committed
```
rsynek, line 17: How does this microprofile-config.properties relate to Quarkus (and its application.properties)?

task-assigning/task-assigning-process-service-client/src/test/java/org/kie/kogito/taskassigning/messaging/UserTaskEventDeserializerTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.kie.kogito.taskassigning.TestUtil.parseZonedDateTime;
+
+class UserTaskEventDeserializerTest {
```
rsynek, line 33: I think this test case can be simplified a bit, depending on what you want to check:
A) to check only the deserialization, prepare the expected object and use [assertIsEqualToComparingFieldByFieldRecursively](https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/internal/Objects.html#assertIsEqualToComparingFieldByFieldRecursively-org.assertj.core.api.AssertionInfo-java.lang.Object-java.lang.Object-java.util.Map-org.assertj.core.internal.TypeComparators-) to make sure it's equal to the object deserialized from JSON
B) in case you want to check also serialization, you can deserialize and again serialize the JSON and compare it as a String with the original JSON String. This way you check also formatting (not sure how much relevant it is in this case).

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEventDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+    public UserTaskEvent deserialize(String topic, byte[] data) {
+        try {
+            if (data == null) {
+                return null;
+            }
+            UserTaskEventMessage message = OBJECT_MAPPER.readValue(data, UserTaskEventMessage.class);
+            UserTaskEvent event = message.getData();
+            event.setEventTime(message.getTime());
+            event.setLastUpdate(message.getTime().truncatedTo(ChronoUnit.MILLIS));
+            event.setEndpoint(buildEndpoint(message.getSource(), event.getProcessInstanceId(), event.getName(), event.getId()));
```
rsynek, line 42: What is the role of the endpoint in the event?

task-assigning/task-assigning-process-service-client/pom.xml
```diff
@@ -35,7 +35,19 @@
...
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-reactive-messaging-kafka</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
```
rsynek, line 47: If we use guava just to do HTML escaping, may I suggest using https://mvnrepository.com/artifact/org.apache.commons/commons-lang3 instead? It's memory footprint is ~5 times smaller.

rsynek (commented):
A couple of suggestions and a couple of questions to help me understand the architecture.

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEventMessage.java
wmedvede: This annotation is for making this class properly managed for the native image generation.

task-assigning/task-assigning-process-service-client/src/main/resources/META-INF/microprofile-config.properties
wmedvede: This properties are very close to the application.properties, since Quarkus is mostly based on eclipse microprofile/microprofile-config. So in general, many of the application.properties might define same things/configurations. 
So this properties in the microprofile-config represent a kind of by default configurations for this kind of "generic" module I've tried to define. But I really have no Quarkus application/service yet.

So, in future step, when I incorporate the module that really defines my Quarkus application (the module that e.g. will take care of starting the solver and put all the pieces working), this properties will be taken as a by default configuration for this generic module, and might be overridden in the application.properties if needed. Most of them won't need any override.

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEventDeserializer.java
wmedvede: I you remember, in my Kogito tour presentation, I've showed how the events carries the url at where the process service live.  So by using this url, we can know how to invoke the referred process, etc.

task-assigning/task-assigning-process-service-client/pom.xml
wmedvede: I prefer to not incorporate commons-lang3, since guava is already being used in the index-service and not commons-lang3. I've also checked that this this guava used class worked well in the native image.

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEvent.java
wmedvede: For this use case, I don't like to be forced to add an abstract class when I believe this classes has a different nature. But also, note that this two classes involves JSON deserialization from json strings that doesn't have exactly the same json field names. Since the UserTaskEvent is used for getting information in json format comming from the events, and the UserTaskInstance is used for getting information comming in the format returned by the data index.
While very similar, those two formats are not exactly the same.
e.g.
the event carries a json field:

    @JsonProperty("taskDescription")
    private String description;

while the data index query returns similar information but in a json field "description"
    //no spetial configuration is required here
    private String description;

Well, to finish a too long explanation. I don't like to have to change my natural way to see it, just because sonar complained about this use case.

task-assigning/task-assigning-process-service-client/src/test/java/org/kie/kogito/taskassigning/messaging/UserTaskEventDeserializerTest.java
wmedvede: I've been looking at the signature of the suggested alternative A) method, see below, but honestly I don't like it and don't see it as a huge simplification. The package "internal" see org.assertj.core.internal.Objects dosen't look inspiring to use too. O the other hand this format is never used in all kogito-runtimes, kogito-apps or even OptaPlanner.

I think current test is ok, what we are testing is a Deserializer, so it's intended to basically take a json  and return a pojo, and this is exactly what the test does.

I'll will go with this format.


void assertIsEqualToComparingFieldByFieldRecursively(AssertionInfo info,
                                                                Object actual,
                                                                Object other,
                                                                Map<String,Comparator<?>> comparatorByPropertyOrField,
                                                                TypeComparators comparatorByType)

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEventMessage.java
rsynek: Yep, I was more interested in what is the reason for reflection there - and I guess it's the JSON serialization itself (through e.g. the jackson library).

task-assigning/task-assigning-process-service-client/src/main/resources/META-INF/microprofile-config.properties
rsynek: Thanks for the explanation.

task-assigning/task-assigning-process-service-client/src/test/java/org/kie/kogito/taskassigning/messaging/UserTaskEventDeserializerTest.java
rsynek: Sorry for a wrong link; it is the implementation hidden by the fluent API, where a similar method is available too: https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractObjectAssert.html#isEqualToComparingFieldByField-java.lang.Object-

task-assigning/task-assigning-process-service-client/pom.xml
rsynek: If we are going to use it for other purposes too, let's keep it. If the only reason is the html escaping, I think it's little value it brings for the memory footprint it has.

task-assigning/task-assigning-process-service-client/src/main/java/org/kie/kogito/taskassigning/messaging/UserTaskEvent.java
rsynek: Sounds fair.

rsynek (approved):
Looks ok, but please reconsider adding the guava dependency from the perspective of the value it brings and its cost.

wmedvede:
> Looks ok, but please reconsider adding the guava dependency from the perspective of the value it brings and its cost.

I will, let's see if in upcoming PRs i have the need to use commons-lang3 in other parts, and if it's compatible with the native image generation. Even so, I believe that for this integration we won't get too much benefit from the potential minimized footprint, see that by one side there won't be a huge decrease by just using commons-lang3, but also, even we the footprint could be a good thing in terms of the application start-up, this is somethinig this integration won't benefit too much I guess. Since each time the application starts, we need to 1) cosume the tasks, 2) start and feed the solver and 3) wait for a solution :(....

wmedvede:
looks like the jenkins job died by getting no response

stderr: fatal: unable to access 'https://github.com/kiegroup/kogito-runtimes.git/': Failed connect to github.com:443; Connection timed out
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2372)

	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:1985)

wmedvede:
Jenkins retest this please

PR #574, opened 2020-12-22
KOGITO-3948 - Redis persistence and new trusty-redis

r00ta:
Jira: https://issues.redhat.com/browse/KOGITO-3948

This PR aims to add a new `persistence-commons` module: `persistence-commons-redis`. 
At the moment, only some operations are supported. 

## Redis Search
Since Redis itself does not support secondary indexing, my proposal is to use the extension Redis Search https://oss.redislabs.com/redisearch/ . In particular, the consumers can create an index on some properties that can be used in the queries. 
Redis search can only create indexes on some basic types (Tag, FullText, Geo, Numeric), and can't handle complex/nested structured objects very well. My workaround is to serialize the entire object and store it into the document under the new property `rawObject`, together with only the indexed properties. For example, if the POJO is 
```
public class Person {
    @JsonProperty("name")
    public String name;

    @JsonProperty("age")
    public Integer age;
} 
```
and the index is created for the property `name` (and the index name is "persons"), then the document that will be stored for the object `new Person("pippo", 18)` will look like
```
{"name" : "pippo",
 "indexName": "persons",
 "rawObject" : "{\"name\": \"pippo\", \"age\": 18}"
}
```
With this strategy, when the consumer queries the storage using some indexed properties, the document is fetched and the `rawObject` deserialization is returned instead. With this strategy it's possible to store and retrieve and handle complex/nested objects. 

The trusty modules have been restructured so that we generate a trusty image that uses infinispan and another one that uses redis. 

Screenshot from 2021-01-04 14-36-41

## Related Pull requests

1. Kogito-runtimes: https://github.com/kiegroup/kogito-runtimes/pull/949
    * update bom
    * add redis test container
2. kogito-images: https://github.com/kiegroup/kogito-images/pull/305
    * add trusty infinispan image
    * add trusty redis image
3. kogito-cloud-operator: https://github.com/kiegroup/kogito-cloud-operator/pull/705
    * use the trusty infinispan image by default 


trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/messaging/incoming/AbstractTraceEventConsumerIT.java
```diff
@@ -65,30 +59,30 @@ public void setup() {
...
         await()
                 .atMost(5, SECONDS)
-                .untilAsserted(() -> assertDoesNotThrow(() -> trustyService.getDecisionById(CLOUDEVENT_WITH_ERRORS_ID)));
+                .untilAsserted(() -> assertDoesNotThrow(() -> trustyService.getDecisionById(TrustyServiceTestUtils.CLOUDEVENT_WITH_ERRORS_ID)));
 
-        Decision storedDecision = trustyService.getDecisionById(CLOUDEVENT_WITH_ERRORS_ID);
+        Decision storedDecision = trustyService.getDecisionById(TrustyServiceTestUtils.CLOUDEVENT_WITH_ERRORS_ID);
         assertNotNull(storedDecision);
-        TraceEventTestUtils.assertDecision(buildDecisionWithErrors(), storedDecision);
+        JSONAssert.assertEquals(MAPPER.writeValueAsString(TrustyServiceTestUtils.buildDecisionWithErrors()), MAPPER.writeValueAsString(storedDecision), true);
```
r00ta, line 86: Removed `TraceEventTestUtils` and used `JSONAssert` to check that the two (very nested/complex) objects are equals.

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/messaging/incoming/TraceEventConverterTest.java
```diff
@@ -14,52 +14,62 @@
...
-    void testCorrectTraceEvent() {
-        doTest(buildCorrectTraceEvent(CORRECT_CLOUDEVENT_ID), buildCorrectDecision(CORRECT_CLOUDEVENT_ID));
+    void testCorrectTraceEvent() throws JsonProcessingException, JSONException {
+        doTest(TrustyServiceTestUtils.buildCorrectTraceEvent(TrustyServiceTestUtils.CORRECT_CLOUDEVENT_ID), TrustyServiceTestUtils.buildCorrectDecision(TrustyServiceTestUtils.CORRECT_CLOUDEVENT_ID));
     }
 
     @Test
-    void testTraceEventWithError() {
-        doTest(buildTraceEventWithErrors(), buildDecisionWithErrors());
+    @Disabled("This test is broken by design, should be changed.")
```
r00ta: The current implementation of `TraceEventTestUtils` was not checking a particular field, with the new assertion method this test is failing. Disabling for the time being

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/KeycloakTrustyServiceIT.java
```diff
@@ -25,26 +32,30 @@
...
     @Test
     void shouldReturnUnauthorized() {
         given().get(TRUSTY_ENDPOINT)
-               .then().statusCode(HttpStatus.SC_UNAUTHORIZED);
+                .then().statusCode(HttpStatus.SC_UNAUTHORIZED);
     }
 
     @Test
     void shouldReturnOkWhenValidUser() {
+        when(trustyService.getExecutionHeaders(any(OffsetDateTime.class), any(OffsetDateTime.class), anyInt(), anyInt(), anyString())).thenReturn(new MatchedExecutionHeaders(new ArrayList<>(), 0));
```
jstastny-cz, line 63: mocking in integration tests seems suspicious. Together with removal of %keycloak.quarkus.oidc props above I am afraid if we're testing the actual integration with keycloak now.

jstastny-cz: ah, didn't notice the rename to microprofile-config.properties while below there's application.properties with retained keycloak config. So just this mocking line is interesting.

trusty/trusty-service/trusty-service-infinispan/src/test/resources/application.properties
```diff
@@ -3,9 +3,7 @@ quarkus.http.cors=true
 #oidc
 quarkus.oidc.enabled=true
 quarkus.oidc.tenant-enabled=false
-%keycloak.quarkus.oidc.tenant-enabled=true
-%keycloak.quarkus.oidc.client-id=kogito-app
-%keycloak.quarkus.oidc.credentials.secret=secret
+quarkus.oidc.auth-server-url=none
```
jstastny-cz, line 5: what happens by this change?

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/KeycloakTrustyServiceIT.java
r00ta: Hi @jstastny-cz , this is needed since I created a new `trusty-service-common` module, and then `trusty-service-infinispan`/ `trusty-service-redis` modules that will provide the selected persistence layer. 
Since `QuarkusTest` will look for a `StorageService` bean, I created a new basic mock for that, but it's not enough for  `trustyService.getExecutionHeaders` since it requires some more complex mocking. That's why I mocked the call directly to return an empty list. The test is about the integration between the trusty application and keyclock. Being a `common` module, the persistence layer has to be mocked.  From my perspective, this is fair..

The alternative would be to make this test `abstract` and execute it for the `infinispan` and `redis` module, but these two modules are not adding any logic to the integration between the application and keycloak, it would be just the same test executed twice.

trusty/trusty-service/trusty-service-infinispan/src/test/resources/application.properties
r00ta: The test project needs the `quarkus.oidc.auth-server-url` to be set, otherwise it fails with 
```
Caused by: javax.enterprise.inject.spi.DeploymentException: No config value of type [java.lang.String] exists for: quarkus.oidc.auth-server-url
```
The deletions are just an artifact of the renaming of the files, since the infinispan tests are not using keycloak I set the property to `none`. wdyt about that?

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/KeycloakTrustyServiceIT.java
jstastny-cz: I was not sure what behavior is actually being mocked. If it's unrelated to keycloak integration, it should be fine. Thanks.

trusty/trusty-service/trusty-service-infinispan/src/test/resources/application.properties
jstastny-cz: Could we actually disable the oidc extension by setting quarkus.oidc.enabled=false ? I think (and you indicated it above) there are no tests actually testing the integration with keycloak.

r00ta: Hi @jstastny-cz , I've refactored the keycloak configuration (now it's the same of the `data-index`). 
The only difference is that in the submodules `redis` and `infinispan` (where we don't have any keycloak specific tests) we still have to specify a `quarkus.oidc.auth-server-url=none` even if `quarkus.oidc.enabled=false`. 
Since we have a dependency to the common test package, the `KeycloakTrustyServiceIT` is annotated with `QuarkusTestResource` and it requires that property.

jstastny-cz: Ah, just wondering if conditional test resource could help here - first in common, the keycloak test would be annotated with `@QuarkusTestResource(value=KeycloakQuarkusTestResource.Conditional.class)` requiring the following system property to be set for the test resource to start `enable.resource.keycloak=true` ... if the property is not set, the container would not start at all.
So the solution might be (haven't checked) set in common the keycloak test resource as conditional, invoke Keycloak related tests in a separate failsafe execution (setting the property and filtering using exclude, include). Then in infinispan and redis modules, when not setting this prop, it could resolve the issue. WDYT?

jstastny-cz: IMO it's actually the config property in test 
```
class KeycloakTrustyServiceIT {

    private static final String VALID_USER = "jdoe";
    private static final String TRUSTY_ENDPOINT = "/executions";

    @ConfigProperty(name = KeycloakQuarkusTestResource.KOGITO_KEYCLOAK_PROPERTY)
    String keycloakURL;
```
enforcing the existence of that prop.

EDITED: so the conditional resource won't help IMO. I don't completely understand how test-jar works, but it seems that it bundles all tests from common module and runs it as part of redis/infinispan module build? If that's so, we're running the Keycloak IT test too? Meaning in redis/infinispan we actually run the Keycloak test even if we don't want to?

Solution could be to replicate the config from common also in redis and infinispan (since we would inherit all the test from there).

r00ta: Hi @jstastny-cz , the `KeycloakTrustyServiceIT` is not run for the infinispan and redis modules. I think that when the infinispan and redis test suite is run quarkus fetches the quarkus test resource `KeycloakQuarkusTestResource` and tries to inject the required property. In short.. `KeycloakTrustyServiceIT` is annotated with `QuarkusTestResource` and by consequence quarkus tries to resolve its dependency even if it is not run for that module. wdyt?

r00ta: btw the Conditional test resource doesn't help in this case as you pointed out

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/CacheManagerServiceMock.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.quarkus.test.Mock;
+import org.kie.kogito.persistence.api.Storage;
+import org.kie.kogito.persistence.api.StorageService;
+import org.kie.kogito.persistence.api.factory.StorageQualifier;
+
+@Mock
+@ApplicationScoped
+@StorageQualifier("mock")
+public class CacheManagerServiceMock implements StorageService {
```
danielezonca, line 32: Do we expect method of this interface to return null?
If not please make all of them return `mock(...)`

persistence-commons/persistence-commons-redis/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>persistence-commons-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.redislabs</groupId>
+      <artifactId>jredisearch</artifactId>
+      <version>2.0.0</version>
```
danielezonca: Use dependencManagement

persistence-commons/persistence-commons-redis/src/main/java/org/kie/kogito/persistence/redis/RedisStorage.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public void addObjectUpdatedListener(Consumer<V> consumer) {
+        throw new UnsupportedOperationException("addObjectUpdatedListener operation not supported for Redis.");
+    }
+
+    @Override
+    public void addObjectRemovedListener(Consumer<String> consumer) {
+        throw new UnsupportedOperationException("addObjectRemovedListener operation not supported for Redis.");
+    }
```
danielezonca: Are you sure Redis doesn't support a similar listener? Btw you can manually implement them with 3 `Collection` of `Consumer` manually invoked inside `put` / `remove`. Wdyt?

pom.xml
```diff
@@ -26,10 +26,11 @@
 
     <!-- external containers used for testing -->
     <container.image.infinispan>quay.io/infinispan/server:${version.org.infinispan}</container.image.infinispan>
+    <container.image.redis>redislabs/redisearch:${version.org.redis}</container.image.redis>
```
danielezonca, line 32: @cristianonicolai 
Is the internal docker image proxy automatic or should we manually configure all the proxed images?

persistence-commons/persistence-commons-redis/src/main/java/org/kie/kogito/persistence/redis/index/RedisIndexManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final RedisClientManager redisClientManager;
+    private Map<String, List<String>> indexes = new HashMap<>();
+
+    @Inject
+    public RedisIndexManager(RedisClientManager redisClientManager) {
+        this.redisClientManager = redisClientManager;
+    }
+
+    public void createIndex(RedisCreateIndexEvent event) {
```
danielezonca, line 49: Is this used? I don't fine any usage (same for firing `ReditCreateIndexEvent` event)

persistence-commons/persistence-commons-redis/src/main/java/org/kie/kogito/persistence/redis/RedisQuery.java
```diff
@@ -0,0 +1,118 @@
...
+    public Query<V> filter(List<AttributeFilter<?>> filters) {
+        this.filters = filters;
+        return this;
+    }
+
+    @Override
+    public Query<V> sort(List<AttributeSort> sortBy) {
+        if (!sortBy.isEmpty()) {
+            if (sortBy.size() > 1) { // TODO: implement backend side sorting on multiple attributes https://issues.redhat.com/browse/KOGITO-4072
+                throw new UnsupportedOperationException("Multiple sorting attributes not implemented yet.");
```
danielezonca, line 76: Can you please add a `README.md` file in the root of this module to list the features that are not supported yet?

pom.xml
r00ta: I think the docker hub proxy works like the following: 
1) an image is requested
2) If the image is in the local registry, send it back. If not, fetch it from docker-hub, store it and then return it

No need to do anything manual afaik

cristianonicolai: yep, it should just pull and cache for further requests

persistence-commons/persistence-commons-redis/src/main/java/org/kie/kogito/persistence/redis/index/RedisIndexManager.java
r00ta: Hi @danielezonca , yes this is the bean that the consumers should use to create an index when the application starts up. For example, in trusty we use it in `trusty-storage-redis` https://github.com/kiegroup/kogito-apps/blob/6cc510181b1c4bb0d6fe1befd30ac9d937aa3d06/trusty/trusty-storage/trusty-storage-redis/src/main/java/org/kie/kogito/trusty/storage/redis/IndexProvider.java

persistence-commons/persistence-commons-redis/src/main/java/org/kie/kogito/persistence/redis/RedisStorage.java
r00ta: Hi @danielezonca , I did not investigate much on these three operations. I had a quick look now and yes, this _seems_ to be  be possible. We need to address this in another ticket in case we want to invest on such operations for some apps that need them https://issues.redhat.com/browse/KOGITO-4258

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/messaging/incoming/TraceEventConverterTest.java
danielezonca: Please create a ticket and link it in the disabled message

persistence-commons/persistence-commons-redis/pom.xml
r00ta: Moved to kogito-apps pom :+1:

trusty/trusty-service/trusty-service-common/src/test/java/org/kie/kogito/trusty/service/common/CacheManagerServiceMock.java
r00ta: We need this mock just to make Quarkus test suite happy with a `StorageService` bean to be injected in the application when the `KeycloakTrustyServiceIT` starts. This is why this class is not doing anything. Added the mock anyway.

trusty/trusty-storage/trusty-storage-api/src/main/java/org/kie/kogito/trusty/storage/api/TrustyStorageServiceImpl.java
```diff
@@ -27,9 +27,9 @@
 @ApplicationScoped
 public class TrustyStorageServiceImpl implements TrustyStorageService {
 
-    private static final String DECISIONS_STORAGE = "decisions";
-    private static final String EXPLAINABILITY_RESULTS_STORAGE = "explainability-results";
-    private static final String MODELS_STORAGE = "models";
+    public static final String DECISIONS_STORAGE = "decisions";
+    public static final String EXPLAINABILITY_RESULTS_STORAGE = "explainability-results";
+    public static final String MODELS_STORAGE = "models";
```
danielezonca: Why don't move these constants to `TrustyStorageService` so that `IndexProvider` can avoid to depends on this Impl class? (or another class of constants)

jitexecutor/README.md
```diff
@@ -60,7 +60,7 @@ If you are interested in the full DMN result, you can use the endpoint `/jitdmn/
   ]
 }
 ```
-
+.git.
```
danielezonca: To be removed?

danielezonca (approved):

r00ta:
Jenkins retest this please

r00ta:
Hi @jiripetrlik , any update on this?

jiripetrlik (approved):
LGTM

r00ta:
Jenkins retest this please

r00ta:
Jenkins retest this please

PR #614, opened 2021-01-26
FAI-376 WeightedLinearRegression

RobGeada:
JIRA: https://issues.redhat.com/browse/FAI-376

Added WeightedLinearRegression class to explainability.utils, as well as corresponding unit tests. This class adds methods to perform weighted linear regressions and recover metrics about the fitted model. 


explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,286 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
```
danielezonca:
```suggestion
 * Copyright 2021 Red Hat, Inc. and/or its affiliates.
```

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/utils/WeightedLinearRegressionTest.java
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
```
danielezonca:
```suggestion
 * Copyright 2021 Red Hat, Inc. and/or its affiliates.
```

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/utils/WeightedLinearRegressionTest.java
```diff
@@ -0,0 +1,247 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+
+class WeightedLinearRegressionTest {
+    // Setup some consistent random generators for various parameters
+    Random rn = new Random();
+    int nRandomTests = 100;
+    int generateNFeatures(){
+        return this.rn.nextInt(50)+1;
+    }
```
r00ta: HI @RobGeada , thanks for the PR! 

Unit tests should be atomic and reproducible. For example, if a test is doing

```java
public int sum(int a, int b) { return a + b; };

@Tets
public void test(){
  Random random = new Random();
  int a = random.nextInt(100);
  int b = random.nextInt(100);
  Assertions.assertEquals(a + b, sum(a, b));
}
```
And after some time it starts to fail because somebody changes `sum` to 
```java
public int sum(int a, int b) { 
  if (a == 42){
    throw new RuntimeException("gg");
  }
  return a + b;
}
```

 you don't know how to reproduce it because the test case is based on random data. Now, your algorithm is much more complicated respect to this `sum` example, so it's going to be impossible to catch the root cause in case something goes wrong. 

I think the algorithm should be tested on some given parameters (alias test case), usually bugs are in the boundaries (for example, if `nfeatures.length = 0, 1` or similar). 

For tests that use a lot of data (for example nfeatures = 50), I would suggest to generate this data randomly once and store it in a file (in case we don't have a concrete a concrete test case to reuse). Then your test simply reads it. wdyt?

r00ta (changes requested):

ruivieira:
Hi @RobGeada, thanks for the PR, looking good!

I only have a few comments, along with some annotations in the code:

- Personally, I would place the model separate to the WLR implementation. This way, if we need to reuse the data in another algorithm, we can just pass the model, instead of passing a `WeightedLinearRegression`. Basically, I would just encapsulate it in a class and change, e.g., `fit(double[][] x, double[] y, double[] sample_weights)`.
- I would also place the matrix utilities (such as as the modified Gauss-Jordan in a separate class). They are useful on their own.
- The rest I've added as inline comments above

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+ * https://www.codeproject.com/Articles/25335/An-Algorithm-for-Weighted-Linear-Regression?msg=4467580#xx4467580xx
+ * We use Dr. Debabrata DasGupta's In-Place Matrix Inversion by Modified Gauss-Jordan algorithm to perform matrix
+ * inversion, as described here:
+ * https://www.researchgate.net/publication/271296470_In-Place_Matrix_Inversion_by_Modified_Gauss-Jordan_Algorithm,
+ * */
+public class WeightedLinearRegression {
+    boolean intercept;
+    boolean fitPerformed;
+
+    private double [][] features;
```
ruivieira: Model outside WLR, perhaps?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+     * @param x An {@code nsamples x nfeatures} array of doubles; each row contains one data point of size [nfeatures]
+     * @param y An {@code nsamples} array, where y[n] is the observation for data point n.
+     * @param sample_weights An {@code nsamples} array, where sample_weights[n] is the weighting of data point n.
+     *
+     * @return C, an {@code nfeatures} array of coefficients as computed by the regression.
+     * In the case where {@code intercept} is true, the last value of {@code C} is the intercept.
+     *
+     */
+    public double[] fit(double[][] x, double[] y, double[] sample_weights) throws ArithmeticException {
+        // if we want to compute an intercept, add a dummy feature at last column.
```
ruivieira, line 48: I would add some initial checking here regarding dimensions (rows(x)=len(weights), etc) and throw an error if not the case.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+                    }
+                }
+            }
+        }
+
+        // invert the coefficient matrix X in-place
+        boolean inversionSuccessful = false;
+        double[][] xInv;
+        int jitterTries = 0;
+        while (! inversionSuccessful && jitterTries < 10) {
```
ruivieira: would it be worth to allow specifying the number of retries (10 could still be the default), instead of hard-coding it?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+
+    // MODEL METRICS ===================================================================================================
+    /**
+     * Recover the goodness-of-fit of the WLR model. This is the coefficient of determination, as per:
+     * https://en.wikipedia.org/wiki/Multiple_correlation
+     * @return the coefficient of determination
+     */
+    public Double getGoodnessOfFit(){
```
ruivieira: Pass a model, perhaps?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+        if (! this.fitPerformed) {
+            return Double.NaN;
```
ruivieira: Personally, I wouldn't return a `Double.NaN`. I would just throw an exception, it wouldn't escape unnoticed.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+            }
+            return  1 - (residualSquareSum/totalSquareSum);
+        }
+    }
+
+    /**
+     * Recover the mean square error of the WLR model.
+     * @return the mean squared error of the model
+     */
+    public Double getMSE(){
```
ruivieira: Same as previous, perhaps pass a model.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+                }
+                double residual = (this.observations[i] - f_i);
+                totalResidual += this.weights[i] * (residual * residual);
+                weightSum += this.weights[i];
+            }
+            return totalResidual/weightSum;
+        }
+    }
+
+    // MATRIX UTILITIES ================================================================================================
```
ruivieira: I would move the matrix utilities into a separate class.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * Inverts the square, non-singular matrix X
+     * @param X a square, non-singular double[][] X
+     * @return the inverted matrix
+     *
+     * Dr. Debabrata DasGupta's description of the algorithm is here:
+     * https://www.researchgate.net/publication/271296470_In-Place_Matrix_Inversion_by_Modified_Gauss-Jordan_Algorithm
+     */
+    public static double[][] invertSquareMatrix(double[][] X) {
```
ruivieira: Perhaps doing an initial check whether this is really a square matrix, throw an exception if not.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+        int size = X.length;
```
ruivieira: I would also some test for the inversion, one for a hardcoded known result where it succeeds (for a small matrix, for instance) and perhaps some edge, cases where it is expected to fail.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+        double[][] copy = new double[size][size];
+        for (int i = 0; i < size; i++) {
+            copy[i] = Arrays.copyOf(X[i], size);
+        }
+
+        double[] pivotsUsed = new double[size];
```
ruivieira: Pivots used could be a `boolean[]` here.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
```diff
@@ -0,0 +1,288 @@
...
+        return copy;
+    }
+
+    /**
+     * Jitters the values of a matrix IN-PLACE by a random number in range (0, delta)
+     * @param X the matrix to be jittered
+     * @param delta the scale of the jittering
+     *
+     */
+    private static void jitterMatrix(double[][] X, double delta){
```
ruivieira: Same as above, separate matrix utility class?

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/utils/WeightedLinearRegressionTest.java
RobGeada: Okay, that makes a lot of sense; I'll refactor the tests to be more granular and non-randomly generated. Thanks!

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
RobGeada: Yeah, there was some reason I was doing that but I've changed how the method works so I definitely need to get rid of return NaN.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
RobGeada: I'm not quite sure how the model would be meaningfully separable? Are you envisioning a class that would basically store the features, observations, and sample weights, and then we'd change WLR.fit and the metric functions to accept that class?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
danielezonca: In general I think we can/should avoid this problem entirely splitting this class in two different objects:
- `WeightedLinearRegressionResult` with all `final` fields (because the result is immutable) with the result (+ intermediate results) of `fit` procedure
- `WeightedLinearRegression` with private constructor and only static methods: `fit` that accept also `intercept` param (or two methods `fit` and `fitWithIntercept`) + the other existing methods with `WeightedLinearRegressionResult` as param

Wdyt?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/WeightedLinearRegression.java
ruivieira: @RobGeada Yes, exactly. I was thinking of how to avoid keeping all the state (features and observations) outside the WLR class, but it's just a suggestion. wdyt?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/Matrix.java
```diff
@@ -0,0 +1,223 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.kogito.explainability.utils;
+
+import java.util.Arrays;
+import java.util.stream.IntStream;
+
+
+public class Matrix {
```
ruivieira: Usually `MatrixUtils` by convention.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/Matrix.java
```diff
@@ -0,0 +1,223 @@
...
+            throw new IllegalArgumentException("# columns of matrix A must match # rows of matrix B" +
+                    String.format("Matrix A shape:  %d x %d, ", aShape[0], aShape[1]) +
+                    String.format("Matrix B shape:  %d x %d,", bShape[0], bShape[1])
+            );
+        }
+
+        double[][] product = new double[aShape[0]][bShape[1]];
+        for (int i = 0; i < aShape[0]; i++) {
+            for (int j = 0; j < bShape[1]; j++) {
+                product[i][j] = 0;
```
ruivieira: Is this necessary? `product` is initialised with zeros anyway.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/Matrix.java
```diff
@@ -0,0 +1,223 @@
...
+        for (int iterations = 0; iterations < size; iterations++) {
+            // OPERATION 1
+            // find the pivot
+            // the pivot idx is the idx of the diagonal element with largest absolute value
+            // that hasn't already been used as a pivot
+            int pivot = Matrix.findPivot(copy, pivotsUsed);
+            double pivotVal = copy[pivot][pivot];
+
+            // check if pivotVal is 0, allowing for some floating point error
+            if (Math.abs(pivotVal) < 1e-9) {
```
ruivieira: Perhaps move the threshold to a static constant.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/Matrix.java
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Attempt to invert the given matrix in-place.
+     * If the matrix is singular, jitter the values slightly to break singularity
+     * @param x a square double[][]; the matrix to be inverted
+     * @param numRetries the number of times to attempt jittering before giving up
+     * @return boolean, whether or not the inversion was successful
+     *
+     */
+    public static double[][] jitterInvert(double[][] x, int numRetries) throws ArithmeticException {
```
ruivieira: I would expect an in-place method to not return anything.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/Matrix.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+        return copy;
+    }
+
+    /**
+     * Attempt to invert the given matrix in-place.
+     * If the matrix is singular, jitter the values slightly to break singularity
+     * @param x a square double[][]; the matrix to be inverted
+     * @param numRetries the number of times to attempt jittering before giving up
+     * @return boolean, whether or not the inversion was successful
```
ruivieira: Change to either `double[][]` or remove (see https://github.com/kiegroup/kogito-apps/pull/614#discussion_r573629008).

ruivieira (changes requested):
@RobGeada thanks for the changes, looking good to me!
I've made a few (minor) inlined comments.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/Matrix.java
RobGeada: Oops, that's a documentation error!

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/MatrixUtils.java
```diff
@@ -0,0 +1,225 @@
...
+ */
+
+package org.kie.kogito.explainability.utils;
+
+import java.util.Arrays;
+import java.util.stream.IntStream;
+
+
+public class MatrixUtils {
+    static double zeroThreshold = 1e-9;
```
danielezonca:
```suggestion
    static final double zeroThreshold = 1e-9;
```

danielezonca (approved):

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/MatrixUtils.java
```diff
@@ -0,0 +1,225 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.kogito.explainability.utils;
+
+import java.util.Arrays;
+import java.util.stream.IntStream;
+
+
+public class MatrixUtils {
```
jiripetrlik, line 22: Maybe it would be good to consider if we can utilize some library for linear algebra such as EJML instead of implement our own code for matrix multiplication etc.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/MatrixUtils.java
```diff
@@ -0,0 +1,225 @@
...
+    static final double zeroThreshold = 1e-9;
```
jiripetrlik: Please make this constant configurable or provide some info how you have found out this exact number. Please also use capital letters for constants.

jiripetrlik (commented):
I have few comments/suggestions.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/MatrixUtils.java
RobGeada: Okay, done; it's now an argument to the relevant methods.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/utils/MatrixUtils.java
RobGeada: Hi Jiri, thanks for the review! I've talked to Daniele, Rui, and Tomasso and the concern with using a linear algebra is the lack of ubiquitous library for it, as well as the burden of maintaining anything that we end up using. The reasoning is that the stuff we need is relatively small compared to what's offered in a full library, so we'll just write our own bits and pieces.

jiripetrlik: Ok, understand. Makes sense.

jiripetrlik (approved):

danielezonca:
jenkins retest this please

PR #620, opened 2021-02-01
KOGITO-3613 update patternfly

AjayJagan:
This Pr is intended to update patternfly dependencies and uniforms, so as to use composable tables and to align the patternfly versions with kogito-tooling repository.


ui-packages/yarn.lock
```diff
@@ -2831,116 +2777,111 @@
...
+"@patternfly/patternfly@4.65.5":
+  version "4.65.5"
+  resolved "https://registry.yarnpkg.com/@patternfly/patternfly/-/patternfly-4.65.5.tgz#22917237ea8b9cbfe52be1dfd00a258a6bfdb324"
+  integrity sha512-10dXyBt4fo55Wtm7v+Vmn9eVp4rD25fR296uif0nxGX3ZXhEuyyPOfo2DgNmCFHu42Exo6N8DP7W472Pp9VkJw==
 
-"@patternfly/react-charts@6.6.0":
-  version "6.6.0"
-  resolved "https://registry.yarnpkg.com/@patternfly/react-charts/-/react-charts-6.6.0.tgz#e41097d4c761d034a6bf70a9266378107a2dc0d6"
-  integrity sha512-biCVvwus5AHNNC8NKlAdwLLlPQaWhSYcBQDX8ZBIRdBejoV/8r8OqXMUMWaaIA5NrTKyrzo//RLOtmtBmm4shA==
+"@patternfly/patternfly@4.70.2":
```
jstastny-cz, line 2780: why do we have two versions now?

ui-packages/yarn.lock
```diff
@@ -2831,116 +2777,111 @@
...
+    focus-trap "6.2.2"
     react-dropzone "9.0.0"
     tippy.js "5.1.2"
+    tslib "1.13.0"
 
-"@patternfly/react-core@4.32.1", "@patternfly/react-core@^4.32.1":
-  version "4.32.1"
-  resolved "https://registry.yarnpkg.com/@patternfly/react-core/-/react-core-4.32.1.tgz#9fd2dc024eda51b8d8e58d014cbb9d69c3a78604"
-  integrity sha512-4FrKJvMfjHjWtmvGu1QVxo/nCdUgePlkdNzMs91r0wdL16CpfoQVtZcfZe4343fRAQ2ObeYfZ2GuiwvS1sw8Og==
+"@patternfly/react-core@^4.79.2":
```
jstastny-cz: Seems we're combining libraries supporting different versions of patternfly, please check that.

jstastny-cz (changes requested):

jstastny-cz (approved):
LGTM. Thanks.
* yarn.lock changes look nice a tidy.
* OUIA related changes are inflicted by increasing support in PF components (good news).

cristianonicolai (approved):

pefernan (approved):
Task Console seems to work fine, thank you @AjayJagan!

kelvah (commented):
Hi @AjayJagan, thank you. Trusty package works great after the update!

I am just curious about the duplicated patternfly dependencies. I see the changes to `package.json` introduced with [c6d151e](https://github.com/kiegroup/kogito-apps/pull/620/commits/c6d151e5fd7d0cc59f00d8164a06f1b21afa337b) are minimal. So how did you manage to solve the duplication in the lock file?

jstastny-cz:
@kelvah we've resorted to manual "merging" of the matching version selectors in yarn.lock(exact vs. caret version). Wondering if there's a better way to make deps like "1.0.0" and "^1.0.0" to be resolved as a single yarn.lock entry so that we have just a single version there.

kelvah:
@jstastny-cz thank you for the clarification. 

I would advise to not change `yarn.lock` manually. Altering it could break things at some point given it's automatically generated/updated by yarn.
I guess certain duplicated deps are the price to pay for having fixed version numbers inside `package.json`.
If we want to reduce duplication (whenever possible) I suppose we could add a `resolutions` field to `package.json`.
Something like:
```
{
   "dependencies": {
      ...
   },
   "resolutions": {
    "@patternfly/react-core": "4.84.4",
    ...
  }
}
```
Of course when updating deps in the future, it will need to be updated too.

Resolutions are [usually used for more critical scenarios](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-why-would-you-want-to-do-this) because they have limitations. But I think it's the only way if we want to keep fixed PF versions in the dependency list.

jstastny-cz:
@kelvah Hm, I just don't understand, why Yarn does the following:
```
"@patternfly/react-core@4.84.4":
  version "4.84.4"
  resolved "https://registry.yarnpkg.com/@patternfly/react-core/-/react-core-4.84.4.tgz#e3061596f0ac02c5cb181f41201293ba073013a1"
  integrity sha512-GIv6zJl+NGYIrSm2pATQFggRt4ZFYUKVNbLSsna2x8+eYeZscqIUFSPmBM+eNpQkGaoJk254Lu09pgm4hM/b4g==
  dependencies:
    "@patternfly/react-icons" "^4.7.22"
    "@patternfly/react-styles" "^4.7.22"
    "@patternfly/react-tokens" "^4.9.22"
    focus-trap "6.2.2"
    react-dropzone "9.0.0"
    tippy.js "5.1.2"
    tslib "1.13.0"

"@patternfly/react-core@^4.84.4":
  version "4.90.2"
  resolved "https://registry.yarnpkg.com/@patternfly/react-core/-/react-core-4.90.2.tgz#d03346d8f3a8d785e076b93df296ce4fa79d9dbe"
  integrity sha512-Ld4HCXYrVZR+ycWKvygmOEDyyMpb/pfZFklJiT/9ktSSSU9SfKwH2XmXotgjwBy1lYUoRYJ58xTpV19duouEoA==
  dependencies:
    "@patternfly/react-icons" "^4.8.4"
    "@patternfly/react-styles" "^4.7.29"
    "@patternfly/react-tokens" "^4.9.26"
    focus-trap "6.2.2"
    react-dropzone "9.0.0"
    tippy.js "5.1.2"
    tslib "1.13.0"
```
when the first entry is already matching the caret version below - am I missing something?

kelvah:
@jstastny-cz 
The first entry is the `@patternfly/react-core` version we listed in our `package.json`, so it has to be there.
The second one is the version requested by `@patternfly/react-table` (another of our deps), which has `@patternfly/react-core: ^4.84.4` as dependency in its own `package.json`. That is resolved by yarn to version `4.90.2`.

So basically we need a certain (older) version of something, while one of our deps wants a newer version of it. Both are installed and we end up with two entries.

jstastny-cz:
@kelvah but it's not wrong per se to merge those in such case, right? Cause yarn docs state following:

Version range | Expanded version range
-- | --
^3.1.4 | >=3.1.4 <4.0.0

So Yarn takes the newest possible for caret version, rather than considering requirements of other packages, fair enough.

With manual merging of these entries - we should not risk anything, cause the requirements can be fulfilled by a single version, that is the 4.84.4 in our case.

To me it seems that it's safer to manipulate yarn.lock than having multiple versions of a single dependency.

kelvah:
@jstastny-cz You're right, I agree the requirements can be satisfied by a single version of the package.

I just don't think it's a good idea to manually edit a generated file. Doing it is discouraged in yarn docs, reasonably I think.
The `resolutions` option means to avoid that.

AjayJagan:
@jstastny-cz so do you think we can add resolutions or can we ignore duplicates and push it ?

AjayJagan:
with this commit f7425e9 , I have added resolutions with the patternfly dependencies. This creates an auto-generated yarn.lock file with the merged versions. Thanks for the info @kelvah :) . Please have look when you have some time. @jstastny-cz Please let me know what you think about this change :)

jstastny-cz:
@AjayJagan looks good. Though I would rely here on @kelvah opinion so that I'm not enforcing something which is against good practice.

ui-packages/package.json
```diff
@@ -123,5 +123,12 @@
     "webpack-dev-server": "^3.8.0",
     "webpack-merge": "^4.2.2",
     "yarn": "^1.22.4"
+  },
+  "resolutions": {
+    "@patternfly/react-core": "4.84.4",
+    "@patternfly/react-icons": "4.7.22",
+    "@patternfly/react-styles": "4.7.22",
+    "@patternfly/react-table": "4.19.45",
```
kelvah: I don't see any duplication of `@patternfly/react-table`, so it's safe to remove it from `resolutions`.

ui-packages/packages/trusty/package.json
```diff
@@ -23,6 +23,9 @@
     "typescript": "~3.8.3",
     "use-react-router-breadcrumbs": "^1.0.4"
   },
+  "resolutions": {
+    "@patternfly/react-charts": "6.12.12"
+  },
```
kelvah: Same here, no need for a `@patternfly/react-charts` resolution.

kelvah (commented):
Thank you @AjayJagan and @jstastny-cz for considering my observations.

I left a couple of comments about the new `resolutions`, I would list only packages affected by duplication and therefore requiring attention.

ui-packages/package.json
AjayJagan: @kelvah removed :+1:

ui-packages/packages/trusty/package.json
AjayJagan: @kelvah removed :+1:

nmirasch (approved):

kelvah (approved):
Looks good. Thank you.

Sara4994 (approved):

PR #621, opened 2021-02-03
KOGITO-4311 Remove data index native tests as the service is not inte…

cristianonicolai:
…nded to run in native mode


r00ta:
Hi @cristianonicolai , if the data-index is intended to be used in native mode we could even remove it from the modules to be compiled with the native profile `kogito-apps/pom.xml`

r00ta (approved):
Hi @cristianonicolai , if the data-index is intended to be used in native mode we could even remove it from the modules to be compiled with the native profile kogito-apps/pom.xml

radtriste (approved):

cristianonicolai:
@r00ta makes sense, I just updated the PR to also remove data index from native builds.

radtriste (approved):

PR #673, opened 2021-02-23
KOGITO-4532 Disable org.kie.kogito.it.trusty.QuarkusTrustyExplainabilityEnd2EndIT

evacchi:
I am disabling this test because it loops indefinitely very often. Please take a look in the meantime


danielezonca (approved):
Thanks for the PR, I agree to disable so we can check why it randomly fails

PR #693, opened 2021-03-05
KOGITO-4596 Process instance details breaks when runtime kogito.service.url is undefined

AjayJagan:
Description : Added service url null check to show/hide the node trigger card.

JIRA : https://issues.redhat.com/browse/KOGITO-4596

### DIFF
--- a/ui-packages/packages/management-console/server/MockData/graphql.js
+++ b/ui-packages/packages/management-console/server/MockData/graphql.js
@@ -950,6 +950,113 @@ module.exports ={ ProcessInstanceData : [
     ]
   },
   {
+    id: '8035b580-6ae4-4aa8-9ec0-e18e19809e0b11111',
+    processId: 'travels',
+    businessKey: null,
+    parentProcessInstanceId: null,
+    parentProcessInstance: null,
+    processName: 'travels(no-service-url)',
+    roles: [],
+    state: 'ACTIVE',
+    rootProcessInstanceId: null,
+    serviceUrl: null,
+    endpoint: 'http://localhost:4000',
+    addons: ['jobs-management', 'prometheus-monitoring', 'process-management'],
+    error: {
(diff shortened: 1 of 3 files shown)


nmirasch:
@AjayJagan we need to add some control on the process diagram panel. When the app don't find it it goes to the runtime endpoint to look for it here: https://github.com/kiegroup/kogito-apps/blob/0b95239a79b105f903da672abc06a7d9e2e5cce6/ui-packages/packages/management-console/src/utils/Utils.tsx#L453
Even the details page is not breaking it shows an empty diagram, could you take a look? Thanks!

jstastny-cz (approved):

AjayJagan:
> @AjayJagan we need to add some control on the process diagram panel. When the app don't find it it goes to the runtime endpoint to look for it here:
> 
> https://github.com/kiegroup/kogito-apps/blob/0b95239a79b105f903da672abc06a7d9e2e5cce6/ui-packages/packages/management-console/src/utils/Utils.tsx#L453
> 
> 
> Even the details page is not breaking it shows an empty diagram, could you take a look? Thanks!

@nmirasch  - I have added  service url check now :+1:

nmirasch:
@AjayJagan Please sync with @Sara4994 about when show the error.
When the kogito.service.url is empty, the process diagram is just hidden. We will display the error when the call to the endpoint can be done and the error is something different from NOT_FOUND.

AjayJagan:
> @AjayJagan Please sync with @Sara4994 about when show the error.
> When the kogito.service.url is empty, the process diagram is just hidden. We will display the error when the call to the endpoint can be done and the error is something different from NOT_FOUND.

@nmirasch  removed the `else` part which triggers the pop-up, thanks :)

nmirasch (approved):
thanks @AjayJagan

Sara4994 (approved):

PR #694, opened 2021-03-06
KOGITO-4337: PostgreSQL persistence support for Job Service

xieshenzh:
Jira: https://issues.redhat.com/browse/KOGITO-4337

Description: PostgreSQL persistence support for Job Service. Build three separate artifacts for in-memory, Infinispan and PostgreSQL. 


r00ta:
Hi @xieshenzh , is there any plan to implement the PostgreSQL persistence layer in the `persistence-commons` module where we have the `infinispan` and the `redis` ones?

xieshenzh:
> Hi @xieshenzh , is there any plan to implement the PostgreSQL persistence layer in the `persistence-commons` module where we have the `infinispan` and the `redis` ones?

Hi @r00ta , @cristianonicolai is implementing PostgreSQL persistence for data index. I think he will look into that. 
The job service uses reactive apis. So @tiagodolphine suggested me to use quarkus reactive sql clients directly. 
In addition, there are other discrepancies between job service persistence and `persistence-commons`. I think we should figure out that in a new thread.

tiagodolphine (commented):
@xieshenzh nice work! Just feedback, I ran your branch locally and it seems to be working fine in general. I need to look into more details in your changes.
Thanks

jobs-service/jobs-service-common/src/main/resources/db/migration/V2.0.0__Create_Table.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TYPE JOB_STATUS AS ENUM
```
cristianonicolai: @xieshenzh @tiagodolphine we probably need a standard way to handle create/update scripts.

jobs-service/jobs-service-pg/src/main/java/org/kie/kogito/jobs/service/repository/pg/PgJobRepository.java
```diff
@@ -0,0 +1,190 @@
...
+                        job.getScheduledId(),
+                        Optional.ofNullable(job.getPayload()).map(p -> new JsonObject(p.toString())).orElse(null),
+                        Optional.ofNullable(job.getType()).map(Enum::name).orElse(null),
+                        job.getPriority(),
+                        recipientMarshaller.marshall(job.getRecipient()),
+                        triggerMarshaller.marshall(job.getTrigger()))
+                        .collect(toList())))
+                .onItem().transform(RowSet::iterator)
+                .onItem().transform(iterator -> iterator.hasNext() ? from(iterator.next()) : null)
+                .emitOn(Infrastructure.getDefaultExecutor()) // Workaround for Quarkus Reactive Client issue with GZip: https://github.com/quarkusio/quarkus/issues/8152
```
cristianonicolai: @xieshenzh maybe mark this workaround as TODO so we can track once is fixed? Or even a JIRA would do.

jobs-service/jobs-service-pg/src/main/java/org/kie/kogito/jobs/service/repository/pg/marshaller/TriggerMarshaller.java
```diff
@@ -0,0 +1,189 @@
...
+                                .put(CLASS_TYPE, t.getClass().getName())),
+                Map.entry(
+                        (Predicate<Trigger>) PointInTimeTrigger.class::isInstance,
+                        (Function<Trigger, JsonObject>) t -> JsonObject.mapFrom(new PointInTimeTriggerAccessor((PointInTimeTrigger) t))
+                                .put(CLASS_TYPE, t.getClass().getName())))
+                .filter(e -> e.getKey().test(trigger))
+                .map(e -> e.getValue().apply(trigger))
+                .findAny()
+                .orElseGet(() -> {
+                    LOGGER.debug("Cannot marshall Trigger to Json: {}", trigger);
```
cristianonicolai: should we throw an exception if this happens? Same for unmarshall. @xieshenzh @tiagodolphine wdyt?

jobs-service/jobs-service-pg/src/main/resources/application.properties
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.username=kogito
+quarkus.datasource.password=kogito
+quarkus.datasource.reactive.url=postgresql://localhost:5432/kogito
+quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/kogito
```
cristianonicolai: I guess this is not in use as we only use reactive?

jobs-service/jobs-service-pg/src/main/resources/application.properties
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.username=kogito
```
cristianonicolai: @xieshenzh @tiagodolphine I think in general we leave configs empty for prod.

jobs-service/jobs-service-pg/src/test/java/org/kie/kogito/jobs/service/repository/pg/PgJobRepositoryTest.java
```diff
@@ -0,0 +1,340 @@
...
+        String query = "SELECT id, correlation_id, status, last_update, retries, execution_counter, scheduled_id, " +
+                "payload, type, priority, recipient, trigger FROM job_details " +
+                "WHERE status IN ('SCHEDULED', 'RETRY') AND (trigger->>'nextFireTime')::INT8 > " + from.toInstant().toEpochMilli() +
+                " AND (trigger->>'nextFireTime')::INT8 < " + to.toInstant().toEpochMilli() + " ORDER BY priority DESC";
+
+        assertEquals(query, queryCaptor.getValue());
+    }
+
+    @Test
+    void findByStatusBetweenDatesOrderByPriority_noCondition() {
```
cristianonicolai:
```suggestion
    void findByStatusBetweenDatesOrderByPriorityNoCondition() {
```

jobs-service/jobs-service-pg/src/test/java/org/kie/kogito/jobs/service/repository/pg/marshaller/RecipientMarshallerTest.java
```diff
@@ -0,0 +1,67 @@
...
+        Recipient recipient = new Recipient.HTTPRecipient("test");
+        JsonObject jsonObject = marshaller.marshall(recipient);
+        assertEquals(new JsonObject()
+                        .put("endpoint", "test")
+                        .put("classType", Recipient.HTTPRecipient.class.getName()),
+                jsonObject);
+    }
+
+    @Test
+    void marshall_null() {
```
cristianonicolai:
```suggestion
    void marshallNull() {
```

jobs-service/jobs-service-pg/src/test/java/org/kie/kogito/jobs/service/repository/pg/marshaller/RecipientMarshallerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    void unmarshall() {
+        JsonObject jsonObject = new JsonObject()
+                .put("endpoint", "test")
+                .put("classType", Recipient.HTTPRecipient.class.getName());
+        Recipient recipient = marshaller.unmarshall(jsonObject);
+        assertEquals(new Recipient.HTTPRecipient("test"), recipient);
+    }
+
+    @Test
+    void unmarshall_invalid() {
```
cristianonicolai:
```suggestion
    void unmarshallInvalid() {
```

jobs-service/jobs-service-pg/src/test/java/org/kie/kogito/jobs/service/repository/pg/marshaller/RecipientMarshallerTest.java
```diff
@@ -0,0 +1,67 @@
...
+        JsonObject jsonObject = new JsonObject().put("endpoint", "test");
+        Recipient recipient = marshaller.unmarshall(jsonObject);
+        assertNull(recipient);
+    }
+
+    @Test
+    void unmarshall_null() {
```
cristianonicolai:
```suggestion
    void unmarshallNull() {
```

jobs-service/jobs-service-pg/src/test/java/org/kie/kogito/jobs/service/repository/pg/marshaller/TriggerMarshallerTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+class TriggerMarshallerTest {
+
+    TriggerMarshaller marshaller = new TriggerMarshaller();
+
+    @Test
+    void marshall_IntervalTrigger() {
```
cristianonicolai:
```suggestion
    void marshallIntervalTrigger() {
```

jobs-service/jobs-service-pg/src/test/resources/application.properties
```diff
@@ -0,0 +1,39 @@
...
+%keycloak.quarkus.oidc.enabled=true
+%keycloak.quarkus.oidc.tenant-enabled=true
+%keycloak.quarkus.oidc.client-id=kogito-app
+%keycloak.quarkus.oidc.credentials.secret=secret
+
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.username=test
+quarkus.datasource.password=test
+quarkus.datasource.reactive.url=postgresql://localhost:5432/test
+quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/test
```
cristianonicolai: not needed?

jobs-service/pom.xml
```diff
@@ -1,283 +1,33 @@
...
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <modules>
+        <module>jobs-service-common</module>
+        <module>jobs-service-infinispan</module>
+        <module>jobs-service-pg</module>
```
cristianonicolai: this is what Im using for data index, should we adopt the same pattern? @xieshenzh @tiagodolphine 
```suggestion
        <module>jobs-service-postgres</module>
```

jobs-service/jobs-service-common/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -1,5 +1,5 @@
 #
```
MarianMacik: Why was the renaming needed? [Quarkus documentation](https://quarkus.io/guides/config-reference#configuration_sources) doesn't even mention this file, just application.properties.

jobs-service/jobs-service-common/src/main/resources/db/migration/V2.0.0__Create_Table.sql
MarianMacik: Moreover, shouldn't this be in the jobs-service-pg module?

jobs-service/jobs-service-common/src/test/resources/application.properties
```diff
@@ -0,0 +1,32 @@
...
+#
+
+kogito.jobs-service.maxIntervalLimitToRetryMillis=60000
+kogito.jobs-service.backoffRetryMillis=1000
+kogito.service.url=http://localhost:8080
+kogito.jobs-service.schedulerChunkInMinutes=10
+kogito.jobs-service.loadJobIntervalInMinutes=10
+kogito.jobs-service.loadJobFromCurrentTimeIntervalInMinutes=0
+kogito.jobs-service.forceExecuteExpiredJobs=false
+kogito.jobs-events-topic=kogito-jobs-events
```
MarianMacik, line 24: A lot of these properties is already defined in microprofile-config.properties. Maybe we could leave here just test-specific ones.

jobs-service/jobs-service-common/src/test/resources/application.properties
```diff
@@ -0,0 +1,32 @@
...
+
+# Keycloak oidc
+quarkus.oidc.enabled=true
```
MarianMacik, line 27: Do we need it enabled also in non-keycloak profile?

xieshenzh:
Assembly:
https://github.com/kiegroup/kogito-runtimes/pull/1136
https://github.com/kiegroup/kogito-apps/pull/694

jobs-service/jobs-service-postgres/src/main/java/org/kie/kogito/jobs/service/repository/postgres/PostgresJobRepository.java
```diff
@@ -0,0 +1,190 @@
...
+                        .emitOn(Infrastructure.getDefaultExecutor())); // TODO Workaround for Quarkus Reactive Client issue with GZip: https://github.com/quarkusio/quarkus/issues/8152
+    }
+
+    @Override
+    public PublisherBuilder<JobDetails> findByStatusBetweenDatesOrderByPriority(ZonedDateTime from, ZonedDateTime to, JobStatus... status) {
+        Optional<String> statusQuery = createStatusQuery(status);
+        Optional<String> timeQuery = createTimeQuery(from, to);
+        String query = statusQuery
+                .flatMap(s -> timeQuery.map(t -> s + " AND " + t)
+                        .or(() -> Optional.of(s)))
```
MarianMacik: `or()` is from Java 9 feature set, do we support that already? Based on [this](https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-build-parent/pom.xml#L52) we don't.
It's also quite difficult to read, maybe a simple solution without Optional would be better, what do you think?
Based on the implementation the from, to and status are never null.
cc @cristianonicolai

jobs-service/jobs-service-postgres/src/main/java/org/kie/kogito/jobs/service/repository/postgres/PostgresJobRepository.java
```diff
@@ -0,0 +1,190 @@
...
+                        .onItem().transform(this::from)
+                        .emitOn(Infrastructure.getDefaultExecutor())); // TODO Workaround for Quarkus Reactive Client issue with GZip: https://github.com/quarkusio/quarkus/issues/8152
+    }
+
+    @Override
+    public PublisherBuilder<JobDetails> findByStatusBetweenDatesOrderByPriority(ZonedDateTime from, ZonedDateTime to, JobStatus... status) {
+        Optional<String> statusQuery = createStatusQuery(status);
+        Optional<String> timeQuery = createTimeQuery(from, to);
+        String query = statusQuery
+                .flatMap(s -> timeQuery.map(t -> s + " AND " + t)
```
MarianMacik: Here you could use map + orElse on the next line.

jobs-service/jobs-service-postgres/src/main/java/org/kie/kogito/jobs/service/repository/postgres/marshaller/RecipientMarshaller.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(RecipientMarshaller.class);
+
+    private static final String CLASS_TYPE = "classType";
+
+    public JsonObject marshall(Recipient recipient) {
+        return Stream.of(
+                Map.entry(
+                        (Predicate<Recipient>) HTTPRecipient.class::isInstance,
+                        (Function<Recipient, JsonObject>) r -> JsonObject.mapFrom(new HTTPRecipientAccessor((HTTPRecipient) r))
+                                .put(CLASS_TYPE, r.getClass().getName())))
```
MarianMacik: What's the point of creating a one-element stream here? This can be avoided I think. Also Map.entry is JDK 9+, not sure we are allowed to use it currently.

jobs-service/jobs-service-postgres/src/test/resources/application.properties
```diff
@@ -0,0 +1,39 @@
...
+# limitations under the License.
+#
+
+kogito.jobs-service.maxIntervalLimitToRetryMillis=60000
+kogito.jobs-service.backoffRetryMillis=1000
+kogito.service.url=http://localhost:8080
+kogito.jobs-service.schedulerChunkInMinutes=10
+kogito.jobs-service.loadJobIntervalInMinutes=10
+kogito.jobs-service.loadJobFromCurrentTimeIntervalInMinutes=0
+kogito.jobs-service.forceExecuteExpiredJobs=false
```
MarianMacik: Same here, a lot of this is already defined in microprofile-config.properties. Maybe we could leave here just test-specific ones. Or maybe microprofile-config is not read by tests?

jobs-service/jobs-service-common/pom.xml
```diff
@@ -0,0 +1,268 @@
...
+              <goal>build</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
```
cristianonicolai: do we need to declare these?

jobs-service/jobs-service-common/src/main/resources/META-INF/microprofile-config.properties
cristianonicolai: I believe the reason was to keep share common properties? I actually think I would prefer to duplicate these across two application.properties even if that means a bit more work to maintain.
@xieshenzh @tiagodolphine @MarianMacik wdyt?

jobs-service/jobs-service-infinispan/pom.xml
```diff
@@ -0,0 +1,216 @@
...
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
cristianonicolai: probably not needed?

jobs-service/jobs-service-postgres/src/main/docker/Dockerfile.native
```diff
@@ -0,0 +1,22 @@
+####
```
cristianonicolai: @xieshenzh are these dockerfile actually used/needed?

jobs-service/jobs-service-postgres/src/main/java/org/kie/kogito/jobs/service/repository/postgres/marshaller/RecipientMarshaller.java
cristianonicolai: yeah, good point, to be honest, I think standard code flow would actually be easier to understand whats going on here.

jobs-service/jobs-service-postgres/src/main/java/org/kie/kogito/jobs/service/repository/postgres/marshaller/RecipientMarshaller.java
```diff
@@ -0,0 +1,95 @@
...
+                .findAny()
+                .orElseGet(() -> {
+                    LOGGER.debug("Cannot marshall Recipient to Json: {}", recipient);
+                    return null;
+                });
+    }
+
+    public Recipient unmarshall(JsonObject jsonObject) {
+        String classType = Optional.ofNullable(jsonObject).map(o -> (String) o.remove(CLASS_TYPE)).orElse(null);
+        return Optional.ofNullable(classType).flatMap(ct -> Stream.of(
```
cristianonicolai: similar comment here

pom.xml
```diff
@@ -29,7 +29,7 @@
     <container.image.redis>redislabs/redisearch:${version.org.redis}</container.image.redis>
     <!-- internal containers used for testing (built by maven using quarkus-container-image-jib) -->
     <container.image.explainability-service-messaging>org.kie.kogito/explainability-service-messaging:${project.version}</container.image.explainability-service-messaging>
-    <container.image.jobs-service>org.kie.kogito/jobs-service:${project.version}</container.image.jobs-service>
+    <container.image.jobs-service>org.kie.kogito/jobs-service-common:${project.version}</container.image.jobs-service>
```
cristianonicolai, line 32: @xieshenzh @MarianMacik @tiagodolphine -common feels strange, is there a better naming we could apply, gessing that will also reflect in the Quay image name, or should we consider breaking into different distros ( inmemory, infinispan, postgres ) ?

jobs-service/jobs-service-common/src/main/resources/META-INF/microprofile-config.properties
MarianMacik: Yes, I think it is better to have just one type of a config file.

jobs-service/jobs-service-postgresql/src/main/java/org/kie/kogito/jobs/service/repository/postgresql/PostgreSqlJobRepository.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private static final String JOB_DETAILS_COLUMNS = "id, correlation_id, status, last_update, retries, " +
+            "execution_counter, scheduled_id, payload, type, priority, recipient, trigger";
+
+    private PgPool client;
+
+    @Inject
+    TriggerMarshaller triggerMarshaller;
+
+    @Inject
```
tiagodolphine: let's inject all attributes in the constructor, in this way we keep the same approach for all attributes, wdyt?

jobs-service/jobs-service-postgresql/src/main/resources/db/migration/V2.0.0__Create_Table.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TYPE JOB_STATUS AS ENUM
```
tiagodolphine, line 1: just a question, using TYPE would be better than just a varchar? performance? in the repository we kind of map the type as a string into a Java Enum. Wdyt?
An other point would be create indexes in postgre for the most used queries, but this is a plus, as an index with type + nextFireTime

pom.xml
tiagodolphine: +1

jobs-service/jobs-service-pg/src/main/java/org/kie/kogito/jobs/service/repository/pg/PgJobRepository.java
tiagodolphine: agree, we should change this as soon as we have the issue fixed, so a jira to track would be great.

pom.xml
MarianMacik: @cristianonicolai then the only "issue" would be that jobs-servicee-postgresql and jobs-service-infinispan modules would depend on jobs-service-inmemory, which may seem a little bit weird, but I am fine with it.

jobs-service/jobs-service-inmemory/src/main/resources/application.properties
```diff
@@ -0,0 +1,117 @@
...
+%events-support.quarkus.kafka.health.enabled=true
+%events-support.kafka.bootstrap.servers=localhost:9092
+%events-support.kogito.jobs-service.events-support=true
+%events-support.mp.messaging.outgoing.kogito-job-service-job-status-events.connector=smallrye-kafka
+%events-support.mp.messaging.outgoing.kogito-job-service-job-status-events.bootstrap.servers=localhost:9092
+%events-support.mp.messaging.outgoing.kogito-job-service-job-status-events.topic=${kogito.jobs-events-topic}
+%events-support.mp.messaging.outgoing.kogito-job-service-job-status-events.value.serializer=org.apache.kafka.common.serialization.StringSerializer
+%events-support.quarkus.arc.exclude-types=
+
+quarkus.oidc.enabled=true
```
MarianMacik: Does this need to be enabled also in the default profile?

pom.xml
tiagodolphine: yeah.. makes sense, it sounds strange... even because the in-memory repository is optional like it just works in case no other persistence is provided, using the @DefaultBean. So suggestions? 
`jobs-service-basic`?

jobs-service/jobs-service-postgresql/src/main/resources/db/migration/V2.0.0__Create_Table.sql
xieshenzh: Because `TYPE` is the way Postgres supports enumeration. If using `varchar`, what is the length we normally use in this case?

jobs-service/jobs-service-inmemory/src/main/resources/application.properties
xieshenzh: I am not sure. It is like this before. 
But the value of the property is fixed at build time, it cannot be changed at runtime: https://quarkus.io/guides/security-openid-connect#quarkus-oidc_quarkus.oidc.enabled
If we don't enable it by default, it cannot be enabled anymore.

jobs-service/jobs-service-common/src/main/resources/db/migration/V2.0.0__Create_Table.sql
xieshenzh: Thanks @MarianMacik . I have fixed this.

jobs-service/jobs-service-pg/src/main/java/org/kie/kogito/jobs/service/repository/pg/PgJobRepository.java
xieshenzh: Will create a jira once this is merged.

jobs-service/jobs-service-inmemory/src/main/resources/application.properties
tiagodolphine: maybe that's the reason, because the approach we had until now, was to have a single image and enable each feature at runtime. But now this is kind of changing with the new persistence modules.

jobs-service/jobs-service-pg/src/main/resources/application.properties
xieshenzh: This is used by Flyway. Reactive client and Flyway use different properties.

pom.xml
cristianonicolai: @MarianMacik @tiagodolphine shouldnt we remove this option then? I mean, people should use the specific image they need. If in memor use `-inmemory`, so in that case there wouldn't be a dependency.

jobs-service/jobs-service-infinispan/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+    </parent>
+
+    <artifactId>jobs-service-infinispan</artifactId>
+    <name>Jobs Service with Infinispan</name>
+    <description>Jobs Service (Timers and Async Jobs) with Infinispan</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>jobs-service-inmemory</artifactId>
```
cristianonicolai: probably not needed anymore as we move to different distros? @xieshenzh @tiagodolphine

jobs-service/jobs-service-infinispan/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>jobs-service</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jobs-service-infinispan</artifactId>
+    <name>Jobs Service with Infinispan</name>
```
cristianonicolai:
```suggestion
    <name>Kogito ::  Jobs Service Infinispan</name>
```

jobs-service/jobs-service-infinispan/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-query-dsl</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>infinispan-quarkus-health-addon</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
```
cristianonicolai, line 49: probably same here.

jobs-service/jobs-service-inmemory/pom.xml
```diff
@@ -0,0 +1,255 @@
...
+         xmlns="http://maven.apache.org/POM/4.0.0">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>jobs-service</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>jobs-service-inmemory</artifactId>
+  <name>Jobs Service In-Memory</name>
```
cristianonicolai:
```suggestion
  <name>Kogito :: Jobs Service In-Memory</name>
```

jobs-service/jobs-service-inmemory/src/main/resources/application.properties
```diff
@@ -0,0 +1,117 @@
...
+#Web Config
+quarkus.ssl.native=true
+quarkus.resteasy.gzip.enabled=true
+quarkus.resteasy.gzip.max-input=10M
+quarkus.http.cors=true
+quarkus.http.limits.max-body-size=10M
+quarkus.http.port=8080
+
+#Swagger
+quarkus.smallrye-openapi.path=/swagger
```
cristianonicolai: @tiagodolphine  @xieshenzh should we leave the default instead so it is consistent across all services?

jobs-service/jobs-service-postgresql/pom.xml
```diff
@@ -0,0 +1,193 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>jobs-service</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jobs-service-postgresql</artifactId>
+    <name>Jobs Service with PostgreSQL</name>
+    <description>Jobs Service (Timers and Async Jobs) with PostgreSQL</description>
```
cristianonicolai, line 14:
```suggestion
    <description>Kogito :: Jobs Service PostgreSQL</description>
```

jobs-service/jobs-service-postgresql/pom.xml
```diff
@@ -0,0 +1,193 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>jobs-service-inmemory</artifactId>
```
cristianonicolai: same comment about perhaps no need to keep this dependency

jobs-service/pom.xml
```diff
@@ -1,283 +1,33 @@
...
-  </profiles>
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-apps</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jobs-service</artifactId>
+    <name>Jobs Service</name>
```
cristianonicolai:
```suggestion
    <name>Kogito :: Jobs Service</name>
```

cristianonicolai (approved):

jobs-service/jobs-service-postgresql/pom.xml
```diff
@@ -0,0 +1,193 @@
...
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>jobs-service-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-pg-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-postgresql</artifactId>
```
cristianonicolai, line 27: @xieshenzh do we need this one as we only use reactive?

xieshenzh: This dependency is used by quarkus flyway. It is based on jdbc: https://quarkus.io/guides/flyway#setting-up-support-for-flyway

cristianonicolai: oh ok,thanks

jobs-service/jobs-service-postgresql/src/main/resources/db/migration/V2.0.0__Create_Table.sql
cristianonicolai: the only downside I see is extra maintenance to update the DB if we need to add a new value to any of these enums.

pom.xml
tiagodolphine: yes, we don't need to provide the image for the in-memory, as we discussed this option is only supposed to be used for testing and dev.

jobs-service/jobs-service-postgresql/src/main/resources/db/migration/V2.0.0__Create_Table.sql
tiagodolphine: yeah, this is the point, that's why I said we could simply and use varchar and keep the enum just in Java. I know for performance TYPE should be better but more difficult for maintenance. Anyway just minor IMO, a matter of choosing an option.

tiagodolphine: so let's keep the type for now, and we can revisit this in the future in case we find that's better a varchar..

tiagodolphine:
After this PR probably the BDD tests + Operator should be updated to the new images 
@radtriste @ricardozanini ^

radtriste:
@tiagodolphine I see that https://issues.redhat.com/browse/KOGITO-4534 and https://issues.redhat.com/browse/KOGITO-4564 are already existing. for that. 
Note that we will need some time to implement and test so please do not merge at the end of a sprint ...

about the different new apps, I see:
- jobs-service-infinispan
- jobs-service-postgresql

do we want as well a `in-memory` image ? cc @MarianMacik @ricardozanini @spolti @cristianonicolai

ricardozanini:
@radtriste actually we are trying to decrease the number of images we have. Maybe we should have only one JobsService image (same for Data Index) with all specific persistence binaries. The motivation to have separated images was to decrease the amount of space required, correct? cc @ibek

radtriste:
yeah the main problem if I remember correctly is that we would have one image with many different runner jar files. and that would be quite big ...
I do understand that we should have minimum number of images but for example, if you take the jobs-service current image, it is already 146MB => https://quay.io/repository/kiegroup/kogito-jobs-service?tab=tags
runner jar from latest [release](https://repository.jboss.org/org/kie/kogito/jobs-service/1.4.0.Final/) is 50MB. let's say you have 3 different storage types (for now), this brings to 150MB of jars in the image.
Also you would so download things you don't need, and we are trying to avoid that or ?

MarianMacik (approved):

pom.xml
MarianMacik: So what is the current status with this? I noticed it was renamed to `inmemory`, now it is back again `common`. I think common sound good from the Maven point of view that `postgresql` and `infinispan` variants depend on it. But for images, if we want to make it clear, we should name it `inmemory` in case we want to distribute it.

MarianMacik:
+1 I am for separate images, it makes it pretty clear which solution you are using.

I also hope that we won't end up supporting too many different persistence solutions for Jobs Service and Data Index. In case we don't keep the number low, we will have either one big image with all variants or a lot of small images. But even if this happens, it is easier to have more smaller images rather than one which is too big.

pom.xml
xieshenzh: @MarianMacik I think the image is just an intermediate product of the build which we don't upload or release. It is only used by the integration tests for kogito runtime/jobs addon and jobs service in the `integration-tests-jobs-service` modules. 
The integration tests for the jobs service and persistence are in the `jobs-service` modules which covers different types of persistence.

ricardozanini:
@MarianMacik @radtriste I agree with your points. Please bring this concern to our next Cloud meeting. :+1:

cristianonicolai:
@xieshenzh needs  rebase

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

tiagodolphine (commented):
LGTM, my last comment is basically the mashaller, maybe we could provide a generic Json marshaller that could be used by mongo and postgresql, but for this to work probably we would need to serialize as Json the whole JobDetails, in this way we would have just one table and a column to store the payload, I'm not sure about performance but it would be more flexible just using Json, but we could try/test test later on a different PR... WDYT @xieshenzh @cristianonicolai?

xieshenzh:
> LGTM, my last comment is basically the mashaller, maybe we could provide a generic Json marshaller that could be used by mongo and postgresql, but for this to work probably we would need to serialize as Json the whole JobDetails, in this way we would have just one table and a column to store the payload, I'm not sure about performance but it would be more flexible just using Json, but we could try/test test later on a different PR... WDYT @xieshenzh @cristianonicolai?

@tiagodolphine I have created the marshaller for JobDetails in the PR for MongoDB. 
I think it is better to have broader discussion and some performance evaluation before taking the one table one column approach with relationship DB.

tiagodolphine:
> > LGTM, my last comment is basically the mashaller, maybe we could provide a generic Json marshaller that could be used by mongo and postgresql, but for this to work probably we would need to serialize as Json the whole JobDetails, in this way we would have just one table and a column to store the payload, I'm not sure about performance but it would be more flexible just using Json, but we could try/test test later on a different PR... WDYT @xieshenzh @cristianonicolai?
> 
> @tiagodolphine I have created the marshaller for JobDetails in the PR for MongoDB.
> I think it is better to have broader discussion and some performance evaluation before taking the one table one column approach with relationship DB.

Yeah, I think we could discuss, and if makes sense, we could do it in a different PR.

tiagodolphine (approved):

xieshenzh:
Jenkins retest please

xieshenzh:
Jenkins retest this

PR #701, opened 2021-03-12
KOGITO-4674 Data Index integration tests with runtimes and using mult…

cristianonicolai:
…iples databases

New IT test modules for Data Index ( Mongo, Infinispan ) and runtime emitting Kafka events.
Will be extended with Postgres once available.


ricardozanini (approved):
Nicely done!

cristianonicolai:
jenkins please retest this

nmirasch (approved):
@cristianonicolai really good!

tiagodolphine (approved):

data-index/data-index-service/data-index-service-infinispan/pom.xml
```diff
@@ -22,6 +22,11 @@
       <artifactId>data-index-storage-infinispan</artifactId>
     </dependency>
 
+    <dependency>
```
radtriste, line 25: With that we enforce the build of a container in simple `package` phase, which is not needed outside integration-tests. Do we really want that ?

radtriste (commented):
One comment.

And also, sorry for nitpicking:
- package and artifacts naming is `mongodb`
- classes naming is `Mongo`
Why this difference ? I think classes and all related naming should be `[M|m]ongoDB`

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-common/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    <sonar.exclusions>**/it/*.java,**/testcontainers/*.java</sonar.exclusions>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <version>${version.org.testcontainers}</version>
+      <scope>compile</scope>
+    </dependency>
```
MarianMacik, line 24: Could this be defined for all integration tests at one place?

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-common/src/main/resources/approvals.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto2"; 
+package org.acme.travels.approvals; 
+import "kogito-index.proto";
+import "kogito-types.proto";
+option kogito_model = "Approvals";
+option kogito_id = "approvals";
+
```
MarianMacik: Can't we use the generated proto file? Just to have it even more integrated. The proto is generated during the build of the service before the tests are run, so it should be available when the Data Index container is created.

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus/src/test/java/org/kie/kogito/index/quarkus/InfinispanTestProfile.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.index.quarkus;
+
+import io.quarkus.test.junit.QuarkusTestProfile;
+
+public class InfinispanTestProfile implements QuarkusTestProfile {
```
MarianMacik, line 24: I guess this is created so Quarkus is restarted between different persistence backends. Why do we need a restart?

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-common/pom.xml
cristianonicolai: moved as depManagmeent at root pom

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus/src/test/java/org/kie/kogito/index/quarkus/InfinispanTestProfile.java
cristianonicolai: we need to spin up different data index, one with mongo, another with infinispan. In theory, this can be done without a restart but would be two data index instances running, I thought was a better approach to isolate the tests.

data-index/data-index-service/data-index-service-infinispan/pom.xml
cristianonicolai: agree, it might not be needed depending on what you're trying to do but I'm not sure what other alternative? do not build if it is run with `-DskipTests` ? would it be worth it? Jobs Service is already taking the same approach

cristianonicolai (commented):
> One comment.
> 
> And also, sorry for nitpicking:
> 
>     * package and artifacts naming is `mongodb`
> 
>     * classes naming is `Mongo`
>       Why this difference ? I think classes and all related naming should be `[M|m]ongoDB`

@radtriste fixed to consistent naming

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-common/src/main/resources/approvals.proto
cristianonicolai: @MarianMacik moved to get proto from generated source

cristianonicolai:
jenkins please retest this

data-index/data-index-service/data-index-service-infinispan/pom.xml
radtriste: I guess we would need to have a `integration-tests` profile with that library included. but that would be a different issue... gonna raise the discussion on Zulip

radtriste: https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/integration.20tests.20and.20.60quarkus-container-image-jib.60

radtriste (approved):

MarianMacik (approved):
Looks good!

One thing which crossed my mind is why we don't use invoker here in the integration tests but we use it for [kogito-quarkus extension](https://github.com/kiegroup/kogito-runtimes/tree/master/integration-tests/integration-tests-kogito-plugin). Maybe here we don't mind if the Kogito Service is not isolated?

PR #705, opened 2021-03-15
PLANNER-2334: Planning execution and problem fact changes generation

wmedvede:

- [X] You have read the [contributors guide](https://github.com/kiegroup/kogito-runtimes#contributing-to-kogito)
- [X] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [X] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [X] Pull Request contains link to the JIRA issue
- [X] Pull Request contains link to any dependent or related Pull Request
- [X] Pull Request contains description of the issue
- [X] Pull Request does not include fixes for issues other than the main ticket

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  
</details>


wmedvede:
Timeout failure somewhere

vms.eng.rdu2.redhat.com/job/KIE/job/kogito/job/pullrequest/job/kogito-apps/336/display/redirect and message: 'Build finished. 1983 tests run, 2 skipped, 0 failed.'

Timeout has been exceeded

Finished: ABORTED

wmedvede:
Jenkins retest this

wmedvede:
Hi @ge0ffrey @rsynek , would you mind review this PR?
Thanks

task-assigning/task-assigning-core/src/main/java/org/kie/kogito/taskassigning/core/model/ModelConstants.java
```diff
@@ -57,11 +51,38 @@ private ModelConstants() {
...
             "dummy-root-process-id",
             DUMMY_DATE,
             DUMMY_DATE,
             DUMMY_DATE,
             "dummy-endpoint"),
             false);
+
+    /**
+     * This task was introduced for dealing with situations where all tasks are pinned and avoid falling into
+     * https://issues.jboss.org/browse/PLANNER-241. Will be removed when issue is fixed.
```
rsynek: The issue has been long marked as resolved, is it really this one?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/resource/TaskAssigningResource.java
```diff
@@ -38,6 +37,6 @@
     @Path("/service/status")
     @Produces({ "application/json" })
     public String getServiceStatus() {
-        return "{\"result\": \"" + service.getStatus() + "\"}";
+        return "{\"result\": \"Not yet implemented!\"}";
```
rsynek, line 41: Would you mind adding a TODO note so that we don't forget to change it in the future?

task-assigning/task-assigning-core/src/test/java/org/kie/kogito/taskassigning/core/model/solver/realtime/TaskStateChangeProblemFactChangeTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    protected TaskStateChangeProblemFactChange createChange(TaskAssignment taskAssignment) {
+        return new TaskStateChangeProblemFactChange(taskAssignment, NEW_STATE);
+    }
+
+    @Override
+    protected Task createTask() {
+        Task task = super.createTask();
+        task.setPriority(CURRENT_STATE);
```
rsynek: How does a priority relate to a state?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/PlanningItem.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.kie.kogito.taskassigning.service;
+
+import org.kie.kogito.taskassigning.core.model.Task;
+
+public class PlanningItem {
+
+    private Task task;
+
+    private String targetUser;
```
rsynek, line 25: Could be made `final`.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/util/TaskUtil.java
```diff
@@ -16,45 +16,308 @@
...
+    /**
+     * Avoid falling into https://issues.redhat.com/browse/KOGITO-4537.
+     * TODO, remove all the usages of this method when fixed.
+     */
+    private static String sanitizeStringValue(String priority) {
+        if (priority != null && priority.isEmpty()) {
+            return null;
+        }
+        return priority;
+    }
```
rsynek, line 92: The method can be simplified a bit:
```suggestion
    private static String sanitizeStringValue(String value) {
        return value == null || value.isEmpty() ? null : value;
    }
```

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public void setLastEventTime(ZonedDateTime lastEventTime) {
+            this.lastEventTime = lastEventTime;
+        }
+    }
+
+    private long changeSetIds;
+    private long currentChangeSetId;
+    private long lastProcessedChangeSetId = -1;
```
rsynek: If we use `AtomicLong`, we can avoid synchronization in the methods below.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
```diff
@@ -0,0 +1,99 @@
...
+    private Map<String, TaskContext> taskContextMap = new HashMap<>();
```
rsynek: `ConcurrentHashMap` should help us avoid synchronization in the remaining methods.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/util/EventUtil.java
```diff
@@ -0,0 +1,71 @@
...
+     * Given a list of events finds the newest event per each task (in case if any) and it to the results list.
+     * 
+     * @param userTaskEvents a list of events to filter.
+     * @return a list of events were each event is the newest one that could be found for the given task.
+     */
+    public static List<UserTaskEvent> filterNewestTaskEvents(List<UserTaskEvent> userTaskEvents) {
+        Map<String, UserTaskEvent> lastEventPerTask = new HashMap<>();
+        UserTaskEvent previousEventForTask;
+        for (UserTaskEvent event : userTaskEvents) {
+            previousEventForTask = lastEventPerTask.get(event.getId());
```
rsynek: A question about the `UserTaskEvent`: is the `id` there an identifier of the event or the task the event relates to?

If it's an identifier of the task, I think renaming the field to the `taskId` would make it clearer.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/util/EventUtil.java
```diff
@@ -0,0 +1,71 @@
...
+     * The returned events are automatically marked as processed in the context.
+     *
+     * @param context the context instance that holds the processed events information.
+     * @param userTaskEvents a list of events to filter.
+     * @return a list of events were each event is the newest one that could be found for the given task that was never
+     *         processed in the current context before.
+     */
+    public static List<UserTaskEvent> filterNewestTaskEventsInContext(TaskAssigningServiceContext context, List<UserTaskEvent> userTaskEvents) {
+        List<UserTaskEvent> result = new ArrayList<>();
+        List<UserTaskEvent> newestTaskEvents = new ArrayList<>(filterNewestTaskEvents(userTaskEvents));
```
rsynek: The defensive copy of the `ArrayList` is redundant since the method `filterNewestTaskEvents` already does the same.

task-assigning/task-assigning-service/src/test/java/org/kie/kogito/taskassigning/service/util/TaskUtilTest.java
```diff
@@ -76,8 +83,149 @@ void fromUserTaskInstance() {
...
+        userTaskEvent.setLastUpdate(LAST_UPDATE);
+        userTaskEvent.setEndpoint(ENDPOINT);
+        List<TaskData> result = TaskUtil.fromUserTaskEvents(Collections.singletonList(userTaskEvent));
+        assertThat(result).hasSize(1);
+        TaskData taskData = result.get(0);
+        assertExpectedTaskData(taskData);
+    }
+
+    @Test
+    void fromTaskData() {
```
rsynek, line 122: Not sure this test brings enough value. Could we maybe use an existing `TaskData` implementation to reduce the boilerplate code, and, as a bonus, increase coverage of that implementation class?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -16,36 +16,56 @@
...
 @ApplicationScoped
 @Startup
 public class TaskAssigningService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(TaskAssigningService.class);
 
+    private static final Predicate<UserTaskEvent> IS_ACTIVE_TASK_EVENT = userTaskEvent -> !TaskState.isTerminal(userTaskEvent.getState());
+
     @Inject
     SolverFactory<TaskAssigningSolution> solverFactory;
```
rsynek, line 71: Can by simplified by injecting directly the `Solver` after we implement https://issues.redhat.com/browse/PLANNER-2369.

rsynek (commented):
There are still several files I have to go through (likely tomorrow).

task-assigning/task-assigning-core/src/main/java/org/kie/kogito/taskassigning/core/model/ModelConstants.java
wmedvede: I have seen the issue was marked as closed, however it looks like it was partially fixed, i.e. for the non chained models, since I can still see the classic already reported "Bailing out" messages.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: > There are still several files I have to go through (likely tomorrow).

Ok, let me know when you finish the review so I can go through all comments together.

wmedvede:
> There are still several files I have to go through (likely tomorrow).

Ok, let me know when you finish the review so I can go through all comments together.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: Will see, by now the solverFactory injection is the only thing we have. But additionally I would like also to have a more controlled way to check the solver instantiation at the service start validation, to print a proper message in cases where the solver instantiation fails, so I'll probably want to use the solverFactory anyway.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+
+        for (TaskAssignment removedTask : removedTasksSet) {
+            removedTaskChanges.add(new RemoveTaskProblemFactChange(removedTask));
+        }
+
+        List<ProblemFactChange<TaskAssigningSolution>> totalChanges = new ArrayList<>();
+
+        totalChanges.addAll(removedTaskChanges);
+        totalChanges.addAll(releasedTasksChanges);
+        assignToUserChangesByUserId.values().forEach(byUserChanges -> byUserChanges.forEach(change -> totalChanges.add(change.getElement())));
```
rsynek: At the first glance, it's not clear there is a nested loop in fact. Maybe different formatting would help?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+            if (!Objects.equals(taskAssignment.getTask().getPriority(), taskData.getPriority())) {
+                propertyChanges.add(new TaskPriorityChangeProblemFactChange(taskAssignment, taskData.getPriority()));
+            }
+            if (!Objects.equals(taskAssignment.getTask().getState(), taskData.getState())) {
+                propertyChanges.add(new TaskStateChangeProblemFactChange(taskAssignment, taskData.getState()));
+            }
+        }
+    }
+
+    private User getUser(Map<String, User> usersById, String userId) {
```
rsynek: If we are interested just in a single user, there is no need to pass the entire map. Maybe `Optional<User>` would make the method contract simpler?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+        this.solution = solution;
+        return this;
+    }
+
+    public SolutionChangesBuilder fromTasksData(List<TaskData> taskDataList) {
+        this.taskDataList = taskDataList;
+        return this;
+    }
+
+    public List<ProblemFactChange<TaskAssigningSolution>> build() {
```
rsynek, line 96: Several collections defined in this top-level method are passed from one method to another, often to get/add a single item from/to them. Making these collections fields of the class would reduce the complexity of the code.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+    private static void addChangeToUser(Map<String, List<IndexedElement<AssignTaskProblemFactChange>>> changesByUserId,
+            AssignTaskProblemFactChange change,
+            User user,
+            int index,
+            boolean pinned) {
+        final List<IndexedElement<AssignTaskProblemFactChange>> userChanges = changesByUserId.computeIfAbsent(user.getId(), key -> new ArrayList<>());
+        addInOrder(userChanges, new IndexedElement<>(change, index, pinned));
+    }
+
+    private void addTaskChanges(TaskAssignment taskAssignment,
```
rsynek, line 160: See the comment above; the method takes 7 parameters, which can be avoided by using fields.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +81,261 @@
...
+
+    private PlanningExecutor planningExecutor;
+
+    private TaskAssigningServiceContext context;
 
-    SolutionDataLoader solutionDataLoader;
+    private AtomicReference<TaskAssigningSolution> currentSolution = new AtomicReference<>(null);
 
-    AtomicReference<String> serviceStatus = new AtomicReference<>("Stopped");
+    private AtomicReference<Boolean> applyingPlanningExecutionResult = new AtomicReference<>(false);
```
rsynek: For this purpose, there is `AtomicBoolean`.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +81,261 @@
...
 
-    int totalChances = 3;
+    private AtomicReference<Boolean> startingFromEvents = new AtomicReference<>(false);
```
rsynek: ditto

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +81,261 @@
...
+                true, true,
+                config.getDataLoaderRetryInterval(), config.getDataLoaderRetries(), config.getDataLoaderPageSize());
     }
 
-    // use the observer instead of the @PreDestroy alternative.
-    // https://github.com/quarkusio/quarkus/issues/15026
-    void onShutDownEvent(@Observes ShutdownEvent ev) {
-        destroy();
+    void onSolutionDataLoad(SolutionDataLoader.Result result) {
+        lock.lock();
```
rsynek, line 148: We are effectively locking the entire method (and other methods using this `lock` instance, too). Is there a reason for not making the methods `synchronized`?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +81,261 @@
...
         }
     }
 
-    public String getStatus() {
-        return serviceStatus.get();
+    void processTaskEvents(List<UserTaskEvent> events) {
+        lock.lock();
+        try {
+            List<UserTaskEvent> newEvents = filterNewestTaskEventsInContext(context, events);
+            if (currentSolution.get() == null) {
```
rsynek, line 228: Would you please add a comment explaining what we try to achieve in this case?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -16,36 +16,56 @@
...
- * The only objective by now is to be sure the tasks can be consumed and the solver started.
- */
+import static org.kie.kogito.taskassigning.core.model.solver.TaskHelper.filterNonDummyAssignments;
+import static org.kie.kogito.taskassigning.service.util.EventUtil.filterNewestTaskEvents;
+import static org.kie.kogito.taskassigning.service.util.EventUtil.filterNewestTaskEventsInContext;
+import static org.kie.kogito.taskassigning.service.util.TaskUtil.fromUserTaskEvents;
+
 @ApplicationScoped
 @Startup
 public class TaskAssigningService {
```
rsynek, line 64: Would you please add Javadoc to the non-private methods? As this is the main service class, a facade over the entire task assigning, it's important to understand what actions it coordinates and why.

rsynek (commented):
The review is done, with the exception of the `TaskAssignmentService`, which is for me difficult to understand without explanatory comments in the code.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
wmedvede: can be done

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
wmedvede: with or without the ConcurrentHashMap this methods still needs to be synchronized.

task-assigning/task-assigning-service/src/test/java/org/kie/kogito/taskassigning/service/util/TaskUtilTest.java
wmedvede: we have by one side the Task, that is the thing that will end-up in the solution, and on the other hand the TaskData that is the task data information that comes from the external system. What we want to test here is that the Task produced from the TaskData has the expected information. The TaskData is a kind of border line between this service and events and queries information. This last two might some day change in favor of kogito provided frameworks in case they exists.
I think it's ok by now.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
wmedvede: If we pass an optional as parameter next thing we'll have so sonar complaining.
I think that even when not perfect it's ok.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: At some point I had a more fine grained access that was latter refactored to what I have now. But I'll keep it as is until I introduce the "waitForImprovedSolutionDuration" and "improveSolutionOnBackgroundDuration", since if I don't remember wrong for this use cases it was nice to have this more fine grained control.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: I'm not fun of writing text telling what the code does, when we have the code read. To keep synchronized the text story of the code and the code itself is a pain, but also a non recommended pattern.
But, yeah, I'll try to add some minimal text.

wmedvede:
@rsynek modifications are in.
Thanks.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
rsynek: In that case, just the `User` would be a sufficient parameter (with a null-check, which is already present). The simpler the method parameters are, the better: the code is easier to read and there is no risk of accidentally changing the map.

task-assigning/task-assigning-core/src/main/java/org/kie/kogito/taskassigning/core/model/ModelConstants.java
rsynek: @ge0ffrey do you by any chance know what remains to be fixed from the PLANNER-241? Is this the https://issues.redhat.com/browse/PLANNER-1738 we see here?

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
rsynek: Would you please show what scenario you have on your mind? I don't see it there, but I may be just missing something.

rsynek:
The code duplication reported by SonarCloud is irrelevant.

task-assigning/task-assigning-core/src/main/java/org/kie/kogito/taskassigning/core/model/ModelConstants.java
wmedvede: I've added a reproducer here

https://issues.redhat.com/browse/PLANNER-241

wmedvede: hmmm, wait, maybe my reproducer falls in the PLANNER-1738

wmedvede: yes, looks like my scenario falls into PLANNER-1738

An executable go reproducer is provided in the link below.

https://github.com/wmedvede/bailing-out-test/blob/master/src/test/java/org/acme/SolverTest.java

Will test all this a bit more add see if this workaround can be removed.

wmedvede: After spending some time on this, I believe that to remove move this second dummy task is even worst, since the probability of falling into the one entity for selection scenario increases, and in fact it can be seen.
However by moving the workaround from place and ensuring the solution to have always two dummy tasks, It looks like issue doesn't happen any more.

So the fix by now is to keep this second dummy entity and always add it to the solution when it's initially created.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
wmedvede: the ConcurrentHashMap only synchronizes the access to the task context, but not the task context it self. So there could still be the case of concurrent acces to the conetxt. Anyway. I'll introduce one last change I won't waste more time on this.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
wmedvede: was aded

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: comments added

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: was added

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: was added

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
wmedvede: was added

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
wmedvede: was added

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/SolutionChangesBuilder.java
wmedvede: was added

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
wmedvede: was added

wmedvede:
Last comments are in, PR/commit descritpion was changed from PLANNER-2234 to PLANNER-2334 that is the real JIRA number.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
```diff
@@ -0,0 +1,100 @@
...
+package org.kie.kogito.taskassigning.service;
+
+import java.time.ZonedDateTime;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class TaskAssigningServiceContext {
+
+    private static class TaskContext {
```
rsynek, line 26: I still don't understand what sequence of two or more threads are we trying to prevent. Right now, for example with threads T1 and T2:
T1 enters `isPublished()` and T2 wants to enter `setLastEventTime()`, but has to wait due to the synchronization.
There is a couple of other combinations, but none of them looks harmful.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +82,321 @@
...
-                userServiceConnector,
-                Duration.ofMillis(5000));
+        planningExecutor = createPlanningExecutor(clientServices, config);
+        managedExecutor.execute(planningExecutor);
+        solutionDataLoader = createSolutionDataLoader(taskServiceConnector, userServiceConnector);
         managedExecutor.execute(solutionDataLoader);
-        solutionDataLoader.start(this::processTaskLoadResult, 1);
+        solutionDataLoader.start(this::onSolutionDataLoad,
+                true, true,
+                config.getDataLoaderRetryInterval(), config.getDataLoaderRetries(), config.getDataLoaderPageSize());
```
rsynek, line 129: If we don't transform the configuration, we could either pass the entire `config` object or let the `SolutionDataLoader` inject it.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +82,321 @@
...
+                    + ", includeUsers: {}, tasks: {}, users: {}", startingFromEvents, result.hasErrors(),
+                    !startingFromEvents.get(), true, result.getTasks().size(), result.getUsers().size());
+            if (result.hasErrors()) {
+                solutionDataLoader.start(this::onSolutionDataLoad,
+                        !startingFromEvents.get(), true,
+                        config.getDataLoaderRetryInterval(), config.getDataLoaderRetries(), config.getDataLoaderPageSize());
+            } else {
+                TaskAssigningSolution solution;
+                List<TaskAssignment> taskAssignments;
+                if (Boolean.TRUE.equals(startingFromEvents.get())) {
```
rsynek:
```suggestion
                if (startingFromEvents.get()) {
```

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +82,321 @@
...
+                    if (hasQueuedEvents()) {
+                        List<UserTaskEvent> newEvents = filterNewestTaskEventsInContext(context, pollEvents());
+                        startingEvents.addAll(newEvents);
+                        startingEvents = filterNewestTaskEvents(startingEvents)
```
rsynek: Aren't we doing the filtering twice (at least for the events added on the previous line)? The `filterNewestTaskEventsInContext` internally calls `filterNewestTaskEvents`.
Maybe moving lines 162-167 to a private, well-named method that returns the filtered `List<UserTaskEvent>` would make it clear what the goal is.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +82,321 @@
...
 
-    public String getStatus() {
-        return serviceStatus.get();
+    /**
+     * Invoked when a set of events are received for processing.
+     * Two main scenarios might happen:
+     * a) A solution already exists and thus the proper problem fact changes are calculated and passed to the solver for
+     * execution. If there are no changes to apply, wait for more events.
+     *
+     * b) No solution exist. Instruct the solution data loader to read the users information and the solver will be
```
rsynek:
```suggestion
     * b) No solution exists. Instruct the solution data loader to read the users information and the solver will be
```

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
```diff
@@ -61,76 +82,321 @@
...
+     * stared when this information is returned plus the information collected from the events.
```
rsynek:
```suggestion
     * started when this information is returned plus the information collected from the events.
```

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
rsynek: If you need to lock only a particular block of code, let's do it - now or later. Otherwise, I would recommend just removing the `lock.lock()` together with the `try-finally` block and using just the `synchronized` keyword on the method declaration. Semantically, it's the same, just easier to understand.

rsynek (commented):
The last piece of the PR has been reviewed, with a couple of more suggestions. Once they are addressed, we are good to merge.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningServiceContext.java
wmedvede: Yes, in general, right now, there aren't any harmful edge cases.
The only thing I don't want is a potential (future) case where the TaskContext is being updated in parallel with no synchronization checks.
I wouldn't like to have two T1 doing setPublished and T2 doing setLastEventTime for the same task. So, by incorporating the ConcurrentHashMap and making the TaskContext synchornized this potential case is also covered.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: This is ok.
In this scenario we are starting the solution from the events.
So we have already loaded a set of startingEvents that where already filtered in context. (i.e. process an even for a task only once)

Now, when the solution data loader ends-up of loading the uses, control returns to this point. And a second events consumption is done in cases events arrived in the middle. Only the new received events must be filtered against the context, the already filtered in context staringEvent can't be filtered again since they have already been filtered.

Finally we combine them, and take only the newest events for each task, anyway.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: I could be, but really this component doesn't need to have access to the entire task assigning service configuration.
By now I will go with this, this start method variant also keeps open the door to invoke different readings on the same solution data loader instance by passing e.g. a different number of retries, or even change the period betwee sync and sync. I prefer this.

task-assigning/task-assigning-service/src/main/java/org/kie/kogito/taskassigning/service/TaskAssigningService.java
wmedvede: Yes, when I add the waitForImprovedSolution and improveSolutionOnBakground if  no fine grained control is required, this can be changed to synchronized.

rsynek (approved):
Approving; once the last comments are addressed and the build is green, feel free to merge. Thanks!

PR #707, opened 2021-03-16
[FAI-386] Implement counterfactuals ability to return intermediate results

ruivieira:
[FAI-386](https://issues.redhat.com/browse/FAI-386): Allow for services to register a consumer with `CounterfactualExplainer` in order to be notified of intermediate counterfactual results during the search.
This is particularly aimed at long running searches, where intermediate results provide an indication of progress.

If no consumers are provided, default ones will be used which will simply log the intermediate (and final results).

> 
> 
> - [ ] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
> - [ ] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
> - [ ] Pull Request contains link to any dependent or related Pull Request

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  
</details>

### DIFF
--- a/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
+++ b/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
@@ -22,6 +22,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ForkJoinPool;
+import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -60,6 +62,16 @@ public class CounterfactualExplainer implements LocalExplainer<CounterfactualRes
     private final SolverConfig solverConfig;
     private final Executor executor;
     private final DataDistribution dataDistribution;
+    private final Consumer<CounterfactualSolution> intermediateResultsConsumer;
+    private final Consumer<CounterfactualSolution> finalResultsConsumer;
+
+    public static Consumer<CounterfactualSolution> defaultIntermediateConsumer = counterfactual -> {
+        logger.debug("Intermediate counterfactual: {}", counterfactual.getEntities());
+    };
+
+    public static Consumer<CounterfactualSolution> defaultFinalConsumer = counterfactual -> {
+        logger.debug("Final counterfactual: {}", counterfactual.getEntities());
+    };
 
     /**
      * Create a new {@link CounterfactualExplainer} using OptaPlanner as the underlying engine.
@@ -69,18 +81,25 @@ public class CounterfactualExplainer implements LocalExplainer<CounterfactualRes
      * The desired outcome is passed using an {@link Output}, where the score of each feature represents the
      * minimum prediction score for a counterfactual to be considered.
      * A customizable OptaPlanner solver configuration can be passed using a {@link SolverConfig}.
+     * A {@link Consumer<CounterfactualSolution>} should be provided for the intermediate and final search results.
+     * Note that the final counterfactual is always returned by the prediction method itself, which means that the
+     * consumer for the final counterfactual can be used to other purposes.
      *
      * @param dataDistribution Characteristics of the data distribution as {@link DataDistribution}, if available
      * @param dataDomain A {@link DataDomain} which specifies the search space domain
      * @param contraints A list specifying by index which features are constrained
      * @param goal A collection of {@link Output} representing the desired outcome
      * @param solverConfig An OptaPlanner {@link SolverConfig} configuration
+     * @param intermediateConsumer A {@link Consumer<CounterfactualSolution>} for the search intermediate result
+     * @param finalConsumer A {@link Consumer<CounterfactualSolution>} for the final intermediate result
      */
     protected CounterfactualExplainer(DataDistribution dataDistribution,
             DataDomain dataDomain,
             List<Boolean> contraints,
             List<Output> goal,
             SolverConfig solverConfig,
+            Consumer<CounterfactualSolution> intermediateConsumer,
+            Consumer<CounterfactualSolution> finalConsumer,
             Executor executor) {
         this.dataDistribution = dataDistribution;
         this.dataDomain = dataDomain;
@@ -88,6 +107,8 @@ public class CounterfactualExplainer implements LocalExplainer<CounterfactualRes
         this.goal = goal;
         this.solverConfig = solverConfig;
         this.executor = executor;
+        this.intermediateResultsConsumer = intermediateConsumer;
+        this.finalResultsConsumer = finalConsumer;
     }
 
     public static Builder builder(List<Output> goal, List<Boolean> constraints, DataDomain dataDomain) {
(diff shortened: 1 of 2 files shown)


explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
```diff
@@ -116,24 +137,25 @@ public static Builder builder(List<Output> goal, List<Boolean> constraints, Data
 
         final UUID problemId = UUID.randomUUID();
 
+        Function<UUID, CounterfactualSolution> initial = (UUID uuid) -> new CounterfactualSolution(entities, model, goal);
```
danielezonca: It should work also without `UUID` explicit type
```suggestion
        Function<UUID, CounterfactualSolution> initial = uuid -> new CounterfactualSolution(entities, model, goal);
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
```diff
@@ -116,24 +137,25 @@ public static Builder builder(List<Output> goal, List<Boolean> constraints, Data
...
+                        solverManager.solveAndListen(problemId, initial, intermediateResultsConsumer, finalResultsConsumer,
+                                null);
                 CounterfactualSolution solution;
                 try {
                     // Wait until the solving ends
                     return solverJob.getFinalBestSolution();
                 } catch (ExecutionException e) {
                     logger.error("Solving failed: {}", e.getMessage());
-                    throw new IllegalStateException("Prediction returned an error", e);
+                    throw new IllegalStateException("Prediction returned an error {}", e);
```
danielezonca: As far as I know `IllegalStateException` constructor doesn't support format in this way: the second parameter is `Throwable` and it is used only for compose multiple cause. In other word I expect the final error message to contain `{}` and not a replaced text

danielezonca: As far as I know `IllegalStateException` constructor doesn't support format in this way: the second parameter is `Throwable` and it is used only for compose multiple cause. In other word I expect the final error message to contain `{}` and not a replaced text

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
```diff
@@ -116,24 +137,25 @@ public static Builder builder(List<Output> goal, List<Boolean> constraints, Data
...
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
-                    throw new IllegalStateException("Solving failed (Thread interrupted)", e);
+                    throw new IllegalStateException("Solving failed (Thread interrupted): {}", e);
```
danielezonca: Same as above

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainerTest.java
```diff
@@ -574,4 +576,64 @@ void testNoCounterfactualPossible(int seed)
...
+        final SolverConfig solverConfig = CounterfactualConfigurationFactory
+                .builder().withTerminationConfig(terminationConfig).build();
+        solverConfig.setRandomSeed((long) seed);
+        solverConfig.setEnvironmentMode(EnvironmentMode.REPRODUCIBLE);
+
+        final AtomicBoolean finalConsumerCalled = new AtomicBoolean(false);
+
+        final Consumer<CounterfactualSolution> assertIntermediateCounterfactualNotNull = counterfactual -> {
+            assertNotNull(counterfactual);
+            assertNotNull(counterfactual.getEntities());
```
danielezonca: Isn't this condition something that only OptaPlanner can guarantee? We don't have control of this 🤔

ruivieira: Yes, that's a good point. I'll perhaps just leave the check that the consumer is actually called.

explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainerTest.java
```diff
@@ -574,4 +576,64 @@ void testNoCounterfactualPossible(int seed)
...
+            finalConsumerCalled.set(true);
+        };
+
+        final CounterfactualExplainer counterfactualExplainer =
+                CounterfactualExplainer
+                        .builder(goal, constraints, dataDomain)
+                        .withSolverConfig(solverConfig)
+                        .withIntermediateConsumer(assertIntermediateCounterfactualNotNull)
+                        .withFinalConsumer(assertFinalCounterfactualNotNull)
+                        .build();
```
r00ta: I think this is not really testing that the consumers are invoked, right?

ruivieira: @r00ta I should have mentioned as a comment that the only consumer being tested for invocation is the final result one (`assertFinalCounterfactualNotNull`), since it's the one that is guaranteed to be called. There is no guarantee that there will be intermediate results. But if `assertFinalCounterfactualNotNull` isn't called, the final assert `assertTrue(finalConsumerCalled.get());` will fail, if I'm not mistaken.

r00ta: I mean.. the wiring of the listeners could be broken and this test would not spot it

r00ta: got it, thanks

danielezonca (approved):

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java
```diff
@@ -116,14 +135,15 @@ public static Builder builder(List<Output> goal, List<Boolean> constraints, Data
 
         final UUID problemId = UUID.randomUUID();
 
+        Function<UUID, CounterfactualSolution> initial = uuid -> new CounterfactualSolution(entities, model, goal);
```
manstis, line 138: Is there value to create `CounterfactualSolution` with the `uuid`?

Or, put another way, is it better to retain the unique identifier for each solution than discard it?

In the UI we're going to need an identifier for each solution anyway.

@kelvah WDYT?

ruivieira: @manstis good point. I was planning on chaning the unique id handling (including CF results) on another PR, is that OK with you?

manstis: OK @ruivieira np... we'll come knocking on your door when we need them (if they're not provided) ;-)

manstis (approved):

r00ta (approved):

PR #719, opened 2021-03-19
KOGITO-4713 Fix URL handling

### LINKED ISSUE #713: data-index computes wrong serviceUrl
https://github.com/kiegroup/kogito-apps/blob/07160792a3f5bb9c36ccf0b66c6de745d5c410ee/data-index/data-index-service/data-index-service-common/src/main/java/org/kie/kogito/index/graphql/GraphQLSchemaManager.java#L174

should use `lastIndexOf()`, e.g.
```java
            return endpoint.contains("/" + context) ? endpoint.substring(0, endpoint.lastIndexOf("/" + context)) : null;
```

To see the problem using `Kogito-travel-agency` example, in `travels` app set `kogito.service.url=http://travels.example.com`.
Then start a new process instance.
When you query `data-index` service you will see the `serviceUrl` field as `http:/` which is wrong.

cristianonicolai:
Fixes #713


radtriste (approved):
lgtm

nmirasch (approved):

PR #794, opened 2021-05-03
FAI-471: Create smoke tests

sjamboro:
https://issues.redhat.com/browse/FAI-471

### DIFF
--- a/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
+++ b/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
@@ -28,7 +28,7 @@ describe('Basic Elements', () => {
       //toggle menu side bar - both directions
       if ($item.is(':visible')) {
         cy.get('button#nav-toggle').click();
-        cy.ouiaId('audit-item').should('be.visible');
+        cy.ouiaId('audit-item').should('not.be.visible');
         cy.get('button#nav-toggle').click();
         cy.ouiaId('audit-item').should('be.visible');
       } else {
@@ -41,25 +41,29 @@ describe('Basic Elements', () => {
   });
 
   it('Search is visible', () => {
-    cy.ouiaId("search-input").should('be.visible');
-    cy.ouiaId("search-input").type('someId');
-    cy.ouiaId("search-button").should('be.visible');
+    cy.ouiaId('search-input').should('be.visible');
+    cy.ouiaId('search-input').type('someId');
+    cy.ouiaId('search-button').should('be.visible');
   });
 
   it('Refresh is visible', () => {
-    cy.ouiaId("refresh-button").should('be.visible');
+    cy.ouiaId('refresh-button').should('be.visible');
   });
 
   it('Date inputs are visible', () => {
-    cy.get('div.flatpickr-wrapper>input#audit-from-date+input').should('be.visible');
-    cy.get('div.flatpickr-wrapper>input#audit-to-date+input').should('be.visible');
+    cy.get('div.flatpickr-wrapper>input#audit-from-date+input').should(
+      'be.visible'
+    );
+    cy.get('div.flatpickr-wrapper>input#audit-to-date+input').should(
+      'be.visible'
+    );
   });
 
   it('Top paging is visible', () => {
-    cy.ouiaId("top-pagination").should('be.visible');
+    cy.ouiaId('top-pagination').should('be.visible');
   });
 
   it('Decision result is available', () => {
-    cy.ouiaId("exec-table").find("tr>td>a:contains('1003')", { timeout: 5000 });
+    cy.ouiaId('exec-table').find("tr>td>a:contains('1003')", { timeout: 5000 });
   });
 });
--- a/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
+++ b/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("Basic Elements", () => {
+    beforeEach(() => {
+      cy.visit("/");
+    });
+  
+    it("Menu button is visible", () => {
+    });
+
+    it("Menu button shows/hides options", () => {
+    });
+  
+    it("Search is visible", () => {
+    });
+  
+    it("Refresh is visible", () => {
+    });
+
+    it("Date inputs are visible", () => {
+    });
+
+    it("Top paging is visible", () => {
+    });
+
+    it("Bottom paging is visible", () => {
+    });
+
+    it("Table is visible", () => {
+    });
+
+    it("Table is empty", () => {
+    });
+  });
--- a/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
+++ b/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
@@ -14,35 +14,57 @@
  * limitations under the License.
  */
 
-describe("Basic Elements", () => {
-    beforeEach(() => {
-      cy.visit("/");
-    });
-  
-    it("Menu button is visible", () => {
-    });
+describe('Basic Elements', () => {
+  beforeEach(() => {
+    cy.visit('/');
+  });
 
-    it("Menu button shows/hides options", () => {
-    });
-  
-    it("Search is visible", () => {
-    });
-  
-    it("Refresh is visible", () => {
-    });
+  it('Menu button is visible', () => {
+    cy.menuButton().should('be.visible');
+  });
 
-    it("Date inputs are visible", () => {
-    });
-
-    it("Top paging is visible", () => {
-    });
-
-    it("Bottom paging is visible", () => {
-    });
-
-    it("Table is visible", () => {
-    });
-
-    it("Table is empty", () => {
+  it('Menu button shows/hides options', () => {
+    cy.auditInvestigationItem().then($item => {
+      //toggle menu side bar - both directions
+      if ($item.is(':visible')) {
+        cy.menuButton().click();
+        cy.auditInvestigationItem().should('not.be.visible');
+        cy.menuButton().click();
+        cy.auditInvestigationItem().should('be.visible');
+      } else {
+        cy.menuButton().click();
+        cy.auditInvestigationItem().should('be.visible');
+        cy.menuButton().click();
+        cy.auditInvestigationItem().should('not.be.visible');
+      }
     });
   });
+
+  it('Search is visible', () => {
+    cy.searchInput().should('be.visible');
+    cy.searchInput().type('someId');
+    cy.searchButton().should('be.visible');
+  });
+
+  it('Refresh is visible', () => {
+    cy.refreshButton().should('be.visible');
+  });
+
+  it('Date inputs are visible', () => {
+    cy.fromInput().should('be.visible');
+    cy.toInput().should('be.visible');
+  });
+
+  it('Top paging is visible', () => {
+    cy.paginationTop().should('be.visible');
+  });
+
+  it('Bottom paging is visible', () => {
+    cy.get("main").scrollTo("bottom");
+    cy.paginationBottom().should('be.visible');
+  });
+
+  it('Decision result is available', () => {
+    cy.pageContent().find("tr>td>a:contains('1003')", { timeout: 5000 });
+  });
+});
--- a/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
+++ b/ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
@@ -20,51 +20,46 @@ describe('Basic Elements', () => {
   });
 
   it('Menu button is visible', () => {
-    cy.menuButton().should('be.visible');
+    cy.get('button#nav-toggle').should('be.visible');
   });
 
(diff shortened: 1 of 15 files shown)


ui-packages/packages/trusty/cypress/support/commands.ts
```diff
@@ -0,0 +1,114 @@
...
+  return cy.get("[data-ouia-component-id='refreshButton']");
+});
+
+Cypress.Commands.add('paginationTop', () => {
+  return cy.get("[data-ouia-component-id='OUIA-Generated-Pagination-top-1']");
+});
+
+Cypress.Commands.add('paginationBottom', () => {
+  return cy.get(
+    "[data-ouia-component-id='OUIA-Generated-Pagination-bottom-1']"
```
sjamboro: @jstastny-cz 
Is it save to use generated ouiaId? I am not sure how to define own ouiaId for Pagination in patternfly.

ui-packages/packages/trusty/cypress/support/commands.ts
```diff
@@ -0,0 +1,114 @@
...
+    paginationBottom(): Chainable<JQuery<HTMLBodyElement>>;
+
+    /**
+     * Return DOM Element for table which represents page content.
+     */
+    pageContent(): Chainable<JQuery<HTMLBodyElement>>;
+  }
+}
+
+Cypress.Commands.add('menuButton', () => {
```
jstastny-cz: Please see https://docs.cypress.io/api/cypress-api/custom-commands#Best-Practices . There's a nice description of what should and should not be a custom command. I personally prefer having nice selectors using OUIA which removes the need for such custom commands as we have here.
I don't think it's that problematic using in tests `cy.get("[data-ouid-component-id='searchInput'"]` instead of `cy.searchInput()`. Moreover for these OUIA specific attributes we can write general commads like `cy.ouidId("")` or `cy.ouiaType("")` which gives you more insight into what actual selector is being used.
This is not selenium where everything needs to be a PageObject or fragment.

ui-packages/packages/trusty/cypress/support/commands.ts
jstastny-cz: It's not safe, once you add another instance of same component into the page the ids are re-generated. Patternfly components are usually written so that it's possible to alter the value. It's just a matter of navigating within our code to find place where this component is instantiated and set respective attribute.

e.g. https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Pagination/Pagination.tsx#L272

ui-packages/packages/trusty/cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "sourceMap": false,
+    "types": ["cypress"]
```
jstastny-cz: I think there should be also the commands now to have correct ts integration.

ui-packages/packages/trusty/package.json
```diff
@@ -28,7 +28,11 @@
     "test:coverage": "yarn test --coverage",
     "test:watch": "jest --watch",
     "lint": "eslint './src/**/*.ts{,x}'",
-    "mock-server": "json-server --watch api-mock/db.js --routes api-mock/routes.json --port 1336 --delay 1000 --id executionId --middlewares api-mock/filterSingular.js"
+    "mock-server": "json-server --watch api-mock/db.js --routes api-mock/routes.json --port 1336 --delay 1000 --id executionId --middlewares api-mock/filterSingular.js",
+    "cypress:mock-and-run": "yarn run mock-server & yarn run cypress:run",
+    "cypress:open": "yarn run cypress open --project cypress",
+    "cypress:run": "yarn run cypress run -b chrome --project cypress",
+    "test:it": "start-server-and-test start http://0.0.0.0:9000 cypress:mock-and-run"
```
jstastny-cz: I would maybe rather combine in order 1. mock, 2. start, 3. cypress:run .... cause to me it seems `mock-and-run` command does not make sense on its own.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
```diff
@@ -88,12 +88,14 @@ export const AuditToolbarTop = (props: toolbarBaseProps & topToolbarProps) => {
...
                 onKeyDown={onSearchEnter}
                 placeholder="Search by ID"
+                data-ouia-component-id="searchInput"
               />
               <Button
                 id="audit-search"
                 variant={ButtonVariant.control}
                 aria-label="search button for search input"
                 onClick={onSearchSubmit}
+                ouiaId="searchButton"
```
jstastny-cz: Please add into readme what are the formatting rules for values of these properties (if camelcase, if dash-separated, etc).

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
```diff
@@ -88,12 +88,14 @@ export const AuditToolbarTop = (props: toolbarBaseProps & topToolbarProps) => {
                 aria-label="search executions"
                 onKeyDown={onSearchEnter}
                 placeholder="Search by ID"
+                data-ouia-component-id="searchInput"
```
jstastny-cz: Just FYI, this is also a possible complication, this depends on the component if it actually passes all the "received" attributes to the rendered element. Which does work for TextInput thanks to https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/TextInput/TextInput.tsx#L176 . 
Though some consider this approach too free-minded and discourage from it - it might happen that this will break in future.
But good to leave for now I'd say.

ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  it('Menu button is visible', () => {
+    cy.menuButton().should('be.visible');
+  });
+
+  it('Menu button shows/hides options', () => {
+    cy.auditInvestigationItem().then($item => {
+      //toggle menu side bar - both directions
+      if ($item.is(':visible')) {
```
jstastny-cz, line 29: This is kinda weird that we don't actually know what state the item should be. You're testing the widget here, not the app configuration - is that the intention?

ui-packages/packages/trusty/cypress/support/commands.ts
jstastny-cz: On second though I would encourage removing these commands and moving the selectors into the test. This way it's hard to deduce the cypress context in which the chain is, cause cy.get() and cy.find() have different behavior, which is again depending on outer context (different behavior in `within` context for example). All this information is here hidden.
Consider having a look at https://github.com/kiegroup/kogito-apps/blob/master/ui-packages/packages/management-console/cypress/support/commands.ts where there is different handling needed based on the context of the command.
Ideally we should unify the commands across the repo so that we use the same thing everywhere.

jstastny-cz (changes requested):
Please review my comments.

ui-packages/packages/trusty/package.json
sjamboro: @jstastny-cz it is about name of script or about order in which these scripts are performed?

Command 'start-server-and-test' supports only 2 commands (one for starting server and second one for running test). I am not sure how I can mock server before the server starts (for example: define command 'start-with-mock-data'). Existed command 'mock-server' is using port 1336. I have decided to present command 'cypress:mock-and-run' (meaning: mock data and run cypress test) because I am sure that port 1336 is listening at the moment.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
sjamboro: @jstastny-cz does it exist somewhere? It seems that everybody use dash-separated (for example: audit-search). So I am going to rewrite it.

ui-packages/packages/trusty/cypress/tsconfig.json
sjamboro: @jstastny-cz Can you share more information? My solution is based on https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfig-json. I have decited to include whole folders "[support](https://github.com/sjamboro/kogito-apps/tree/FAI-471/ui-packages/packages/trusty/cypress/support)" and "[integration](https://github.com/sjamboro/kogito-apps/tree/FAI-471/ui-packages/packages/trusty/cypress/integration)". File commands.ts is included by file index.ts. I have followed tsconfig.json in [kogito-tooling](https://github.com/kiegroup/kogito-tooling/blob/master/packages/online-editor/it-tests/tsconfig.json)

ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
sjamboro: @jstastny-cz yes, the purpose of BasicTest.ts is to verify that UI is available and it contains all desired elements (menu, search and filter options, list of items and possibility to go next page). I can remove this test. You are right that I tests widget not application. Unfortunately, it is only one widget which works, filters does not work with mocked data, pagination does not show another page or change a number of items on page. I can only click on item and show more details. Mocked data is generated randomly so I am not sure what to test. This is a reason why I want to find a way to build this application with real data.

ui-packages/packages/trusty/src/components/Organisms/AuditToolbar/AuditToolbar.tsx
sjamboro: @jstastny-cz README in kogito-apps contains contribution part which link https://github.com/kiegroup/kogito-runtimes/blob/master/CONTRIBUTING.md. IMHO the note about using dash-separated format of values should be addressed there. It is different github repo, so I have decided to create this issue https://issues.redhat.com/browse/FAI-481 and implement it by new PR.

ui-packages/packages/trusty/cypress/support/commands.ts
sjamboro: I removed this test

ui-packages/packages/trusty/cypress/support/commands.ts
sjamboro: I removed these functions and I have defined `cy.ouiaId("")`

ui-packages/packages/trusty/package.json
sjamboro: I have renamed `cypress:mock-and-run` for this PR. I have created https://issues.redhat.com/browse/FAI-482 to find a way how change the order of these command.

jstastny-cz (approved):
Thank you for the changes.

I think it's fine for the initial configuration and getting things together. We've discussed offline that there are rooms for improvement and utilization of what's already available in other packages, but that will itself require some refactoring which is out of scope of this task and will be solved as a separate effort.

ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it('Menu button shows/hides options', () => {
+    cy.ouiaId('audit-item').then($item => {
+      //toggle menu side bar - both directions
+      if ($item.is(':visible')) {
+        cy.get('button#nav-toggle').click();
+        cy.ouiaId('audit-item').should('be.visible');
+        cy.get('button#nav-toggle').click();
+        cy.ouiaId('audit-item').should('be.visible');
```
kelvah: I don't know if it's relevant but I think there's something wrong here. Maybe `audit-item` should not be visible the first time?

ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
```diff
@@ -0,0 +1,65 @@
...
+        cy.get('button#nav-toggle').click();
+        cy.ouiaId('audit-item').should('not.be.visible');
+      }
+    });
+  });
+
+  it('Search is visible', () => {
+    cy.ouiaId("search-input").should('be.visible');
```
kelvah: I see Prettier warnings because of the usage of both single and double quotes. Could you please make sure to enable Prettier support on your IDE? We have a configuration inside `ui-packages/.prettierrc`.
It's not documented anywhere, I know. It should be at least in the README. Sorry about that.

kelvah (approved):
Hi @sjamboro, approving and leaving a couple of minor comments.

I also have a question about linting. Should cypress folders be ignored inside the eslint configuration? @jstastny-cz wdyt?

jstastny-cz:
@kelvah the tests should not conflict with current settings, or do they? If not, we can configure later some cypress-specific linting rules, using for example https://www.npmjs.com/package/eslint-plugin-cypress .

kelvah:
> @kelvah the tests should not conflict with current settings, or do they? If not, we can configure later some cypress-specific linting rules, using for example https://www.npmjs.com/package/eslint-plugin-cypress .

I saw some linting errors in the cypress files, therefore I asked. Agree with providing some specific configuration later.

ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
sjamboro: @kelvah my mistake. I am going to fix it.

ui-packages/packages/trusty/cypress/integration/smoke/BasicTest.ts
sjamboro: @kelvah thanks, I am going to fix it and test it. Because this branch is not performed now. It may cause problem later.

sjamboro:
@kelvah can you tell me where you find linting errors in the cypress files? They were in PR checks?

kelvah:
> @kelvah can you tell me where you find linting errors in the cypress files? They were in PR checks?

hi @sjamboro. No, I was just opening the files with my editor, which should follow our eslint configuration.
I saw errors on `ui-packages/packages/trusty/cypress/support/commands.ts` and `ui-packages/packages/trusty/cypress/plugins/index.ts`.

sjamboro (commented):
@kelvah @jstastny-cz I have added next commit which should solve ESlint errors. I hope that commit message contains description of my steps. Otherwise feel free to ask me.

sjamboro:
Jenkins retest this

kelvah:
> @kelvah @jstastny-cz I have added next commit which should solve ESlint errors. I hope that commit message contains description of my steps. Otherwise feel free to ask me.

Hi @sjamboro, I just checked locally and the linting errors are gone. Thank you.

PR #820, opened 2021-05-14
FAI-475 - Automatically optimize LIME hyperparameters for stability

tteofili:
See https://issues.redhat.com/browse/FAI-475

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>


danielezonca:
@tteofili 
Can you please have a look to ?

> kiegroup_kogito_apps/explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/PrequalificationDmnLimeExplainerTest.java:123 | org.opentest4j.AssertionFailedError: Unexpected exception thrown: org.kie.kogito.explainability.utils.ValidationUtils$ValidationException: Expected negative stability score bigger than 0.4. Got:0.3 for LTV@k=1 	at org.junit.jupiter.api.AssertDoesNotThrow.createAssertionFailedError(AssertDoesNotThrow.java:83)

It is a failure in GHA build and it looks like a flaky test 🤔

tteofili:
now the build is green again

ruivieira (approved):
@tteofili lgtm, thank you!

tteofili:
thanks @ruivieira !

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntity.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.explainability.local.lime.optim;
+
+abstract class LimeConfigEntity {
```
danielezonca, line 18: Public?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntityFactory.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.kie.kogito.explainability.local.lime.LimeConfig;
+import org.kie.kogito.explainability.model.EncodingParams;
+import org.kie.kogito.explainability.model.PerturbationContext;
+
+class LimeConfigEntityFactory {
```
danielezonca, line 28: Public?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final boolean DEFAULT_NUMERIC_ENTITIES = true;
+
+    private long timeLimit;
+    private int tabuSize;
+    private int acceptedCount;
+    private boolean numericEntities;
+    private boolean booleanEntities;
+    private EasyScoreCalculator<LimeStabilitySolution, SimpleBigDecimalScore> scoreCalculator;
+
+    private LimeConfigOptimizer(long timeLimit, int tabuSize, int acceptedCount,
```
danielezonca: Why this constructor exist?
As far as I can see it is only used invoked by the other constructor and it is private

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,83 @@
...
+                    Thread.currentThread().interrupt();
+                } catch (TimeoutException e) {
+                    LOGGER.error("Timed out while waiting for saliency stability calculation");
+                }
+            }
+            if (succeededEvaluations > 0) {
+                stabilityScore /= succeededEvaluations;
+            }
+        }
+        return SimpleBigDecimalScore.parseScore(Double.toString(stabilityScore));
```
danielezonca: Why this double to string to double conversion?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -0,0 +1,168 @@
...
+        if (numericEntities) {
+            entities.addAll(LimeConfigEntityFactory.createNumericEntities(config));
+        }
+        if (booleanEntities) {
+            entities.addAll(LimeConfigEntityFactory.createBooleanEntities(config));
+        }
+
+        if (entities.isEmpty()) {
+            return config;
+        } else {
```
danielezonca: Please remove the else, the first if has a return so we can avoid this nesting

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilitySolution.java
```diff
@@ -0,0 +1,81 @@
...
+
+@PlanningSolution
+public class LimeStabilitySolution {
+
+    private LimeConfig config;
+
+    private List<Prediction> predictions;
+
+    @PlanningEntityCollectionProperty
+    private List<LimeConfigEntity> entities;
```
danielezonca: Do we really need this "flexibility"? In counterfactual scenario input is completely dynamic and based on model details while here the planning entity can be probably one with n fields based on the properties we have.
I don't know if this approach has performance implications.
@triceo @ge0ffrey 
Do you have any suggestion? Is a single planning entity with multiple fields preferable compared to this "dynamic" way to provide an arbitrary list of entities?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -0,0 +1,168 @@
...
+        this.tabuSize = tabuSize;
+        return this;
+    }
+
+    public LimeConfigOptimizer withAcceptedCount(int acceptedCount) {
+        this.acceptedCount = acceptedCount;
+        return this;
+    }
+
+    public LimeConfigOptimizer withNumericEntities(boolean numericEntities) {
```
danielezonca: As far as I can see numeric or boolean entities are just properties to enable, does it make sense to expose a specific "enable only numeric"?
Said differently, do we expect any use cases where only one (or even none) of them is enabled?
In any case I find the name of these methods quite obscure: what should user understand from it?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -0,0 +1,168 @@
...
+
+            try (SolverManager<LimeStabilitySolution, UUID> solverManager =
+                    SolverManager.create(solverConfig, new SolverManagerConfig())) {
+
+                UUID executionId = UUID.randomUUID();
+                SolverJob<LimeStabilitySolution, UUID> solverJob =
+                        solverManager.solve(executionId, initialSolution);
+                try {
+                    // Wait until the solving ends
+                    LimeStabilitySolution finalBestSolution = solverJob.getFinalBestSolution();
```
danielezonca: What if a solution cannot be found? Will the method return a null?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,83 @@
...
+        LimeConfig config = LimeConfigEntityFactory.toLimeConfig(solution);
+        double stabilityScore = 0;
+        List<Prediction> predictions = solution.getPredictions();
+        if (!predictions.isEmpty()) {
+            double succeededEvaluations = 0;
+            int topK = 2;
+            LimeExplainer limeExplainer = new LimeExplainer(config);
+            for (Prediction prediction : predictions) {
+                try {
+                    LocalSaliencyStability stability = ExplainabilityMetrics.getLocalSaliencyStability(solution.getModel(),
```
danielezonca: I expect this scoring method to be quite heavyweight, it should be nice in the future to canculate the complexity of the whole algorithm :thinking:

explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/FraudScoringDmnLimeExplainerTest.java
```diff
@@ -112,12 +91,69 @@ void testFraudScoringDMNExplanation() throws ExecutionException, InterruptedExce
...
+        LimeConfig optimizedConfig = limeConfigOptimizer.optimize(initialConfig, predictions, model);
+
+        LimeExplainer limeExplainer = new LimeExplainer(optimizedConfig);
+        PredictionInput testPredictionInput = getTestInput();
+        List<PredictionOutput> testPredictionOutputs = model.predictAsync(List.of(testPredictionInput))
+                .get(Config.INSTANCE.getAsyncTimeout(), Config.INSTANCE.getAsyncTimeUnit());
+        Prediction instance = new SimplePrediction(testPredictionInput, testPredictionOutputs.get(0));
+
+        assertDoesNotThrow(() -> ValidationUtils.validateLocalSaliencyStability(model, instance, limeExplainer, 1,
+                0.5, 0.5));
```
danielezonca, line 158: What are you testing in particular? I was expecting an assertion/comparison between Lime and optimized Lime to make sure the optimized one is better or the same compared to the "normal" 🤔 
Or at least that optimizedConf is different compared to initialConf
Am I missing something?
(same comment for all othe equivalent tests)

danielezonca (commented):
Sorry for delay with the review but it is a completely new features :)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
triceo: I agree with Daniele here. I really can not recommend using the easy calculator unless your problems are really trivial. Easy score calculators are not incremental, and therefore very slow.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,83 @@
...
+import org.slf4j.LoggerFactory;
+
+public class LimeStabilityScoreCalculator implements EasyScoreCalculator<LimeStabilitySolution, SimpleBigDecimalScore> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LimeStabilityScoreCalculator.class);
+
+    @Override
+    public SimpleBigDecimalScore calculateScore(LimeStabilitySolution solution) {
+        LimeConfig config = LimeConfigEntityFactory.toLimeConfig(solution);
+        double stabilityScore = 0;
```
triceo: Double should generally be avoided in scoring functions. It may lead to score corruptions due to numerical instability.
If you're using doubles to avoid the perf loss of BigDecimal, I suggest you use some kind of fixed-point arithmetic.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,83 @@
...
+                        succeededEvaluations++;
+                    }
+                } catch (ExecutionException e) {
+                    LOGGER.error("Saliency stability calculation returned an error {}", e.getMessage());
+                } catch (InterruptedException e) {
+                    LOGGER.error("Interrupted while waiting for saliency stability calculation {}", e.getMessage());
+                    Thread.currentThread().interrupt();
+                } catch (TimeoutException e) {
+                    LOGGER.error("Timed out while waiting for saliency stability calculation");
+                }
```
triceo: Having the scoring function block for some other long-running calculation is a bad thing to do to the solver.
Scoring functions are supposed to be stateless, fast and isolated.
Slow scoring functions will inevitably lead to poor quality solutions.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilitySolution.java
triceo: I think that having multiple entity types will lead to easier constraints - if you need to distinguish between the types (boolean, number etc.), doing it on a class level will perform the best - as you can limit your constraints to just that class type, reducing the number of entities OptaPlanner  has to iterate over.

That said, you really do not need to worry about this here, because your scoring function in its current implementation will be the bigger bottleneck.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
triceo: OptaPlanner will find *some* solution. It may not be feasible (hard constraints will be broken), but there will be a solution. That said, your score calculator uses SimpleScore, so there are no hard constraints - just one score level. Therefore you needn't worry about this - there will always be a solution.

triceo:
In general, I recommend running the solver for several minutes with `<environmentMode>FULL_ASSERT</>`.
At the moment, I am not entirely convinced that the scoring function is stable. (But having not run it myself, I may be totally wrong.)
If that's the case, performance does not matter, as the function would be fundamentally unsound.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntity.java
tteofili: I think this and other LIME optimization classes should not be public as they're meant to be used internally.
The only `public` class that is meant to be instantiated by the user is `LimeConfigOptimizer`.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntityFactory.java
tteofili: same as above

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
tteofili: ok, I'll just keep the `public` one.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
tteofili: the `SimpleBigDecimalScore` doesn't accept `double`, so it's either converted to `BigDeciamal` or `String`.
I'll move to `BigDecimal` as it sounds more convenient.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
tteofili: according to the javadoc it never returns `null`, if a good solution can't be found it return the initial state (in our case the starting `LimeConfig`).

explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/FraudScoringDmnLimeExplainerTest.java
tteofili: all the ITs have a lower bounded stability score in the LIME explanation test case, then there's another test demanding a higher bounded stability score when the optimized config is used.
Anyway I agree we should at least make sure the optimized config is different than the initial config.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilitySolution.java
tteofili: the reason for having the flexibility in defining entities according to their type is because there might be new LIME parameters to be handled in `LimeConfig` in the future and it sounded like it'd be easier to incorporate them by just modifying `LimeConfigEntityFactory`.

explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/DmnTestUtils.java
```diff
@@ -0,0 +1,139 @@
...
+        t2.put("Auth Code", "Denied");
+        transactions.add(t2);
+        Map<String, Object> map = new HashMap<>();
+        map.put("Transactions", transactions);
+        List<Feature> features = new ArrayList<>();
+        features.add(FeatureFactory.newCompositeFeature("context", map));
+        PredictionInput predictionInput = new PredictionInput(features);
+        predictionInputs.add(predictionInput);
+
+        Random random = new Random();
```
r00ta: Is this making the tests non-deterministic?

r00ta (commented):
Just surfing around :)

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
tteofili: the point about these configuration handles is mostly related to performance, however I agree that we might actually "group" entities by affected part of the LIME algorithm, so that it makes more sense also for the end user.

explainability/explainability-integrationtests/explainability-integrationtests-dmn/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/dmn/DmnTestUtils.java
tteofili: thanks, fixed now

tteofili:
Jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
tteofili: I've created https://issues.redhat.com/browse/FAI-514 to address this in a separate issue/PR.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+import org.slf4j.LoggerFactory;
+
+public class LimeStabilityScoreCalculator implements EasyScoreCalculator<LimeStabilitySolution, SimpleBigDecimalScore> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LimeStabilityScoreCalculator.class);
+
+    @Override
+    public SimpleBigDecimalScore calculateScore(LimeStabilitySolution solution) {
+        LimeConfig config = LimeConfigEntityFactory.toLimeConfig(solution);
+        BigDecimal stabilityScore = BigDecimal.valueOf(0d);
```
triceo:
```suggestion
        BigDecimal stabilityScore = BigDecimal.ZERO;
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+        List<Prediction> predictions = solution.getPredictions();
+        if (!predictions.isEmpty()) {
+            stabilityScore = getStabilityScore(solution, config, predictions);
+        }
+        return SimpleBigDecimalScore.of(stabilityScore);
+    }
+
+    private BigDecimal getStabilityScore(LimeStabilitySolution solution, LimeConfig config, List<Prediction> predictions) {
+        double succeededEvaluations = 0;
+        BigDecimal stabilityScore = BigDecimal.valueOf(0d);
```
triceo:
```suggestion
        BigDecimal stabilityScore = BigDecimal.ZERO;
```

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+            positiveStabilityScore = positiveStabilityScore.add(BigDecimal.valueOf(stability.getPositiveStabilityScore(decision, i)));
+            negativeStabilityScore = negativeStabilityScore.add(BigDecimal.valueOf(stability.getNegativeStabilityScore(decision, i)));
+        }
+        positiveStabilityScore = positiveStabilityScore.divide(BigDecimal.valueOf(topK), RoundingMode.CEILING);
+        negativeStabilityScore = negativeStabilityScore.divide(BigDecimal.valueOf(topK), RoundingMode.CEILING);
+        // TODO: FAI-495 - differentiate (or weight) between positive and negative
+        return positiveStabilityScore.add(negativeStabilityScore)
+                .divide(BigDecimal.valueOf(2d).multiply(BigDecimal.valueOf(stability.getDecisions().size())),
+                        RoundingMode.CEILING);
+    }
```
triceo: For performance, I suggest you cache the BigDecimals. This code is on the hot path, and in a loop.
You can reuse `BigDecimal.valueOf(topK)`.
For `BigDecimal.valueOf(2d)` and the zeros, you can turn them into constants.

This is some nice "free" performance.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public SimpleBigDecimalScore calculateScore(LimeStabilitySolution solution) {
+        LimeConfig config = LimeConfigEntityFactory.toLimeConfig(solution);
+        BigDecimal stabilityScore = BigDecimal.valueOf(0d);
+        List<Prediction> predictions = solution.getPredictions();
+        if (!predictions.isEmpty()) {
+            stabilityScore = getStabilityScore(solution, config, predictions);
+        }
+        return SimpleBigDecimalScore.of(stabilityScore);
```
triceo, line 48: If zero, you can simply return `SimpleBigDecimalScore.ZERO`, saving yourself some time and GC.

tteofili:
Jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeStabilityScoreCalculator.java
tteofili: resolving as I've opened another ticket to revisit this.

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlRegressionLimeExplainerTest.java
```diff
@@ -66,31 +70,12 @@ void testPMMLRegression() throws Exception {
         random.setSeed(0);
         PerturbationContext perturbationContext = new PerturbationContext(random, 1);
         LimeConfig limeConfig = new LimeConfig()
-                .withSamples(10)
+                .withSamples(100)
```
danielezonca, line 73: Why more sample? Does this impact test execution time?

danielezonca (approved):

explainability/explainability-integrationtests/explainability-integrationtests-pmml/src/test/java/org/kie/kogito/explainability/explainability/integrationtests/pmml/PmmlRegressionLimeExplainerTest.java
tteofili: I have noticed that this was a bit flaky, using more samples solves that, no noticeable difference in the test execution time.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntity.java
```diff
@@ -0,0 +1,39 @@
...
+
+    protected Object proposedValue;
+    protected String name;
+
+    protected LimeConfigEntity() {
+
```
jiripetrlik: Maybe there should be some default values. What do you think?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/BooleanLimeConfigEntity.java
```diff
@@ -0,0 +1,58 @@
...
+import org.optaplanner.core.api.domain.valuerange.ValueRange;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeFactory;
+import org.optaplanner.core.api.domain.valuerange.ValueRangeProvider;
+import org.optaplanner.core.api.domain.variable.PlanningVariable;
+
+@PlanningEntity
+public class BooleanLimeConfigEntity extends LimeConfigEntity {
+
+    public BooleanLimeConfigEntity() {
+        super();
```
jiripetrlik, line 28: Shouldn't there be any default values?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/BooleanLimeConfigEntity.java
```diff
@@ -0,0 +1,58 @@
...
+        return (Boolean) proposedValue;
+    }
+
+    public void setProposedValue(Boolean proposedValue) {
+        this.proposedValue = proposedValue;
+    }
+
+    @Override
+    double asDouble() {
+        return Double.NaN;
```
jiripetrlik: I would suggest to either throw some exception which says that it is not implemented or use abstract methods for this.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntity.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    protected LimeConfigEntity(String name, Object proposedValue) {
```
jiripetrlik, line 28: Does it make sense to protect this constructor? Class is abstract and it is made public in specific classes. I would suggest to keep it public.

jiripetrlik (commented):
I have few comments/suggestions.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntity.java
tteofili: I can put them there, but instance fields for `LimeConfigEntity` and subclasses are intended to be set (via reflection) by OptaPlanner and hence it doesn't really make much difference. However for the sake of cleaner code I think it's anyway better to give them some "dummy" initial values anyway.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/BooleanLimeConfigEntity.java
tteofili: same as above

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/BooleanLimeConfigEntity.java
tteofili: agree on throwing an exception

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigEntity.java
tteofili: ok

PR #854, opened 2021-05-31
KOGITO-5209 unify data-table to PF standards

AjayJagan:
unified data table according to PF standards

### DIFF
--- a/ui-packages/packages/components-common/src/components/DataTable/DataTable.tsx
+++ b/ui-packages/packages/components-common/src/components/DataTable/DataTable.tsx
@@ -26,8 +26,12 @@ import {
   sortable,
   ISortBy
 } from '@patternfly/react-table';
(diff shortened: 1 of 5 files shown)


pefernan (changes requested):
Really good @AjayJagan! would it be possible to also show the no-filters-selected empty state in the table? It should be easy to pass it using the ErrorComponent property and pass the error flag as true...

AjayJagan:
> Really good @AjayJagan! would it be possible to also show the no-filters-selected empty state in the table? It should be easy to pass it using the ErrorComponent property and pass the error flag as true...

@pefernan , in management console , when displaying the no status selected , we hide the table headers .. and in task console we have the same functionality .. so do we need to change it?

pefernan:
@AjayJagan you're right, my bad!

pefernan:
jenkins retest this

pefernan (approved):
+1

PR #858, opened 2021-06-02
FAI-502: Implement outcome match threshold for numerical features

ruivieira:
[FAI-502](https://issues.redhat.com/browse/FAI-502): 

A threshold (as a ratio of the value) is implemented, allowing for a customisable tolerance when matching numerical variables. Categorical variables will not use tolerance and still use a strict match.
The previous behaviour (strict match) is supported by setting the threshold to zero.
The default value is 0.01 and it can be passed to the `CounterfactualExplainerProducer` with the key `trusty.explainability.counterfactuals.goalThreshold` in the `explainability-services` properties file.

> 
> 
> 
> - [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
> - [ ] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
> - [x] Pull Request contains link to any dependent or related Pull Request

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>


manstis (approved):
From what I understand, it looks good to me.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterFactualScoreCalculator.java
```diff
@@ -110,8 +111,25 @@ public BendableBigDecimalScore calculateScore(CounterfactualSolution solution) {
...
+
+                    if (output.getType().equals(Type.NUMBER)) {
+
+                        final double outputValue = output.getValue().asNumber();
+                        final double goalValue = goalOutput.getValue().asNumber();
+                        final double difference = Math.abs(outputValue - goalValue);
+                        final double threshold = solution.getGoalThreshold();
+                        final double change = difference / Math.max(outputValue, goalValue);
+
+                        if ((goalValue == 0.0 && difference > threshold) || (goalValue != 0.0 && change > threshold)) {
```
danielezonca: Why this is different logic based on goalValue? Can you please clarify?

ruivieira: Hi @danielezonca, it would be useful for the tolerance to be scale invariant and unitless so that it can behave consistently for large and small values. However, the solutions we discussed offline had an edge case when the goal value was zero. The standard relative change is used for all cases where `goal!=0`. When `goal==0` we add a special case and check the  difference directly. (small rationale [here](https://ruivieira.dev/counterfactuals-with-constraint-solvers.html#tolerance))

danielezonca (approved):

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterFactualScoreCalculator.java
```diff
@@ -110,8 +111,25 @@ public BendableBigDecimalScore calculateScore(CounterfactualSolution solution) {
...
+                            distance += difference * difference;
+                        }
+
+                    } else {
+                        distance +=
+                                goalOutput.getValue().getUnderlyingObject().equals(output.getValue().getUnderlyingObject()) ? 0
+                                        : 1;
+                    }
+
```
r00ta, line 132: I have a more general consideration more related to the algorithm itself and not to the code.  
Let's assume I have two features: one numerical and one categorical (boolean for instance). The numerical one has a very large span (from 0 to 1M). The distance on the categorical feature will basically have no effect, because the distance on the numerical feature will always "cover" it, right? 
If this is correct, should we think about a kind of weighted average?

r00ta: I might be wrong, but I think the categorical features have basically no impact on the distance with this implementation. 
Let's say we have two features: one numerical and one categorical (boolean).
If the numerical feature has a large span (let's say `[-100, 100]`), then the `difference` might be `200`. The contribution to the `distance` of this feature would be `40000`. On the other hand, the contribution of the categorical feature is always `1` or `0`, a very small number respect to the numerical one. But actually both features should have the same importance. 
I might be wrong, but I suspect the numerical features have a big impact on the distance respect to the categorical feature and this might impact the quality of the solutions OP might find. wdyt?

ruivieira: Hi @r00ta, there are two different distances to consider. This can be a issue, but not for this particular distance.
- The outcomes `distance` above is used to calculate the `primaryHardScore`. Being a "hard" constraint, the actual magnitude will be less important (we are mostly interested in knowing if any proposed solution's outcomes are different from the goal, i.e. if `distance > 0`)
- The inputs `distance` (`primarySoftScore`) is more sensitive to this issue. It uses the `.distance()` method as defined by the `CounterfactualEntity` interface. A possible way to mitigate this problem is already implemented as scaling the distance (c.f. `AbstractNumericEntity::distance`), but this unfortunately requires having access to the original data.

r00ta: @ruivieira we calculate this distance only for those features for which the user sets the CF search domain right? If yes, don't we already have a maximum value that we can use to calculate an average weigth for the distance? 

I'm thinking about something like `finalDistance = sum_{i=1}^n 1/i featureDistance_i` where `featureDistance_i` is a number `[0,1]`. In this way `finalDistance` would be a number `[0,1]` where every feature has the same weight

ruivieira: @r00ta that is a possible way of scaling the distance, but not without problems.
The main one is that by making the distance dependent on the domain size you could get potential different counterfactuals for different bounds.
If you think this is an approach that we should implement, perhaps we could implement it as an alternative search method that can enabled via configuration, and with all the proper disclaimers (that different searches with the same inputs/goals might yeld different CFs depending on the search bounds)? wdyt?

r00ta: Honestly I'm not an expert of the algorithm and I think we should dig deeper into the problem and do some practical tests to decide what to implement. 
I think we have to understand what is the relationship between the `distance` and the partial matching of the `goals`. Our situation is that we might have a non-intuitive situation like
1) many primary constraints are satisfied with a big distance `X`
2) none primary constaints are satisfied but the distance `Y` is very small:  `Y <<<< X`. 

Can this confuse OP when it tries to find the solutions? Do you know somebody from OP that can provide us a consultancy? :)

manstis (approved):
LGTM (my simpleton view).. I'll let @ruivieira and @r00ta debate the algorithm itself.

ruivieira:
> LGTM (my simpleton view).. I'll let @ruivieira and @r00ta debate the algorithm itself.

Hi @manstis (cc @r00ta), many thanks for the review!

This PR is converted to draft, mostly due to offline discussions where another direction was chosen.
A new scoring/distance calculation is being implemented with [FAI-515](https://issues.redhat.com/browse/FAI-515) which will normalise the distances between `0.0` and `1.0`. When FAI-515's is ready, this PR will be either changed and removed from draft or deleted and included directly in FAI-515. 

The advantage is that the new score can be directly compared to the threshold (avoiding involving feature values and edge cases as it's the case now) and the weight of categorical and numerical variables will be the same (full explanation will be included in FAI-515's PR).

ruivieira:
Jenkins retest this please

tteofili (approved):
LGTM thanks!

jiripetrlik (approved):

ruivieira:
Jenkins retest this please

PR #980, opened 2021-08-10
Extend GH action timeout

cristianonicolai:


radtriste:
Jenkins is `180` => https://github.com/kiegroup/kogito-apps/blob/main/Jenkinsfile#L228
So I guess we could put the same value :)

radtriste (approved):

PR #991, opened 2021-08-20
KOGITO-5743 - Review native tests

cristianonicolai:


cristianonicolai:
@tiagodolphine native tests were using injection, so my guess is that it never actually worked. If jobs service is meant to run in native mode, I guess we need to do a follow up to implement these.

cristianonicolai:
Jenkins run native

tiagodolphine:
> @tiagodolphine native tests were using injection, so my guess is that it never actually worked. If jobs service is meant to run in native mode, I guess we need to do a follow up to implement these.

right, probably they were not working, TBH we never focused to make it run in native build, since it is supposed to be a long running service, if we wanna provide it to run in native mode probably we need to review a few points to make it work, but better to file a Jira for it.

tiagodolphine (approved):

cristianonicolai:
Jenkins run native

cristianonicolai:
jenkins please retest this

cristianonicolai:
Jenkins run native

radtriste (approved):

cristianonicolai:
merging this as it already improves/fixes some modules. @r00ta is helping to review the missing fix for trusty service

PR #1000, opened 2021-08-30
FAI-555 - make it possible to run LIME config optimization deterministically

tteofili:
See https://issues.redhat.com/browse/FAI-555

- [x] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>


explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -62,13 +65,20 @@ public LimeConfigOptimizer() {
         this.samplingEntities = DEFAULT_SAMPLING_ENTITIES;
         this.encodingEntities = DEFAULT_ENCODING_ENTITIES;
         this.weightingEntities = DEFAULT_WEIGHTING_ENTITIES;
+        this.seed = Long.MIN_VALUE;
```
jiripetrlik: Interesting. Just out of curiosity. Is there any reason for using MIN_VALUE as seed?

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -130,23 +140,32 @@ public LimeConfig optimize(LimeConfig config, List<Prediction> predictions, Pred
...
 
         TerminationConfig terminationConfig = new TerminationConfig();
-        terminationConfig.setSecondsSpentLimit(timeLimit);
+        if (timeLimit > 0) {
+            terminationConfig.setSecondsSpentLimit(timeLimit);
+        }
         solverConfig.setTerminationConfig(terminationConfig);
 
         LocalSearchPhaseConfig localSearchPhaseConfig = new LocalSearchPhaseConfig();
+        if (seed != Long.MIN_VALUE) {
```
jiripetrlik: Can we somehow allow user to set seed to a different value than MIN_VALUE and enable random option in some different way? For example make variable seed Optional<> etc?

manstis:
FWIW we took a different approach to make Counterfactual tests deterministic.

Allow for OP to be proxied:-

https://github.com/kiegroup/kogito-apps/blob/main/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainer.java#L124-L125

Mock use of OP in tests:-

https://github.com/kiegroup/kogito-apps/blob/main/explainability/explainability-core/src/test/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualExplainerTest.java#L658-L669

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
tteofili: this actually is just a placeholder for a kind of "unusual value to be set" to work as `seedNotSet`.
So basically if the seed is not set explicitly (`seed == Long.MIN_VALUE`), the seed won't be set in OP config.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
tteofili: I think it makes sense to set the instance field  `seed` as `Optional<Long>` so that if it's not set we don't need a special `Long` value for that.

tteofili:
Jenkins retest this please

tteofili:
@manstis thanks for your suggestions, I would rather not expose OP API from within the `LimeExplainer` API if possible.
The underlying technology for optimizing LIME parameters should not be constraining the API unless needed, for the sake of the evolution of the codebase (e.g. we move to a non back-compat version of OP in 2051 :smile: ).

For mocking, I like the idea for unit tests, thanks.

manstis:
> @manstis thanks for your suggestions, I would rather not expose OP API from within the `LimeExplainer` API if possible.
> The underlying technology for optimizing LIME parameters should not be constraining the API unless needed, for the sake of the evolution of the codebase (e.g. we move to a non back-compat version of OP in 2051  ).
> 
> For mocking, I like the idea for unit tests, thanks.

That's a very good point.

We could make [this](https://github.com/kiegroup/kogito-apps/blob/main/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualConfig.java#L41) `protected`; only overriding and using it from unit tests.. it'd keep the use of OP more _internal_. Or.. we could decouple _explainers_ from _solvers_ all together and create our own interface for the "solver". The default implementation would delegate to OP but we could use whatever in other implementations too...

@ruivieira @RobGeada WDYT re: keeping some consistency in the implementation of explainers!?

TBH I am sticking my (unwanted) nose in your work so you can also tell me to "move on, nothing to see here".

tteofili:
@manstis thx for your suggestions, please keep them going ;)

tteofili:
Jenkins retest this please

ruivieira:
> > @manstis thanks for your suggestions, I would rather not expose OP API from within the `LimeExplainer` API if possible.
> > The underlying technology for optimizing LIME parameters should not be constraining the API unless needed, for the sake of the evolution of the codebase (e.g. we move to a non back-compat version of OP in 2051  ).
> > For mocking, I like the idea for unit tests, thanks.
> 
> That's a very good point.
> 
> We could make [this](https://github.com/kiegroup/kogito-apps/blob/main/explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/counterfactual/CounterfactualConfig.java#L41) `protected`; only overriding and using it from unit tests.. it'd keep the use of OP more _internal_. Or.. we could decouple _explainers_ from _solvers_ all together and create our own interface for the "solver". The default implementation would delegate to OP but we could use whatever in other implementations too...
> 
> @ruivieira @RobGeada WDYT re: keeping some consistency in the implementation of explainers!?
> 
> TBH I am sticking my (unwanted) nose in your work so you can also tell me to "move on, nothing to see here".

@manstis, @tteofili I think these are all important points and I definitely agree that the OP details should be hidden as much as possible from the actual explainer API. Only exposing the configuration bits we need, decoupled from OP's config seems a good choice, we could perhaps continue the discussion. 

As for the method that allows to pass a complete OP config object to the CF explainer: since OP config is a complex one, this was as convenient way of passing a "bespoke" OP configuration without needing to implement rarely-used fields in a CF explainer config. It can be removed (or made protected as you suggest), since it's not essential.

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -62,13 +66,20 @@ public LimeConfigOptimizer() {
         this.samplingEntities = DEFAULT_SAMPLING_ENTITIES;
         this.encodingEntities = DEFAULT_ENCODING_ENTITIES;
         this.weightingEntities = DEFAULT_WEIGHTING_ENTITIES;
+        this.seed = Optional.empty();
```
danielezonca: What about reusing `PerturbationContext` also here?
I think you can:
- move this optional `seed` field to `PerturbationContext` and update the class to handle it as additional constructor
```java
    private Optional<Long> seed = Optional.empty();

    public PerturbationContext(Random random, long seed, int noOfPerturbations) {
        this.seed = Optional.of(seed);
        random.setSeed(seed);
        this(random, noOfPerturbations);
    }

    public Optional<Long> getSeed() {
        return seed;
    }
```
- remove all `random.setSeed(..)` done outside `PerturbationContext`
- update the code here to fetch the seed from `PerturbationContext`

In this way we have a single and common way to specify the seed in all the context (and also we can remove a lot of boilerplate in many tests).
Of course this does imply/impose to use the same `PerturbationContext` instance everywhere, we can instantiate multiple but the handling is the same

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
```diff
@@ -130,23 +141,32 @@ public LimeConfig optimize(LimeConfig config, List<Prediction> predictions, Pred
...
+        if (timeLimit > 0) {
+            terminationConfig.setSecondsSpentLimit(timeLimit);
+        }
         solverConfig.setTerminationConfig(terminationConfig);
 
         LocalSearchPhaseConfig localSearchPhaseConfig = new LocalSearchPhaseConfig();
+        if (seed.isPresent()) {
+            solverConfig.setRandomSeed(seed.get());
+            solverConfig.setEnvironmentMode(EnvironmentMode.REPRODUCIBLE);
+        }
```
danielezonca, line 166: Can you add an else branch with a `log.debug` information about the not reproducible env?

tteofili: ok, sure

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
tteofili: good idea, I'll give it a try to see how it looks like in terms of passing `PerturbationContexts` around :)

jiripetrlik (commented):
Thank you @tteofili looks good to me. It is always good idea to allow tests to be more deterministic! :+1:

jiripetrlik (approved):

tteofili:
Jenkins retest this please

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/local/lime/optim/LimeConfigOptimizer.java
danielezonca: Please use `debug` instead of `warn`: in general it might be not a problem to have no reproducible build, it is based on user decision

explainability/explainability-core/src/main/java/org/kie/kogito/explainability/model/PerturbationContext.java
```diff
@@ -25,15 +26,25 @@
...
     private final int noOfPerturbations;
 
     public PerturbationContext(Random random, int noOfPerturbations) {
+        this.seed = Optional.empty();
         this.random = random;
         this.noOfPerturbations = noOfPerturbations;
     }
 
+    public PerturbationContext(Long seed, Random random, int noOfPerturbations) {
+        this.seed = Optional.of(seed);
```
danielezonca: Just to be on the safe side
```suggestion
        this.seed = Optional.ofNullable(seed);
```

danielezonca (approved):
LGTM, just a minor comment and [a reply](https://github.com/kiegroup/kogito-apps/pull/1000/files#r715755543)

PR #1020, opened 2021-09-13
FAI-551: CF UI - Configuration of progress bar duration

manstis:

See https://issues.redhat.com/browse/FAI-551

This PR moves the configuration parameter for Counterfactual explanation timeouts from the `explainability-service` to the `trusty-service` now passing the parameter in the explanation request to `explainabilty-service` and returning the setting in the UI response payload.

@kelvah FYI. I've made a new [JIRA](https://issues.redhat.com/browse/FAI-574) to update the CF UI use the setting returned in the "request response payload".

@ruivieira FYI too.. I changed `CounterfactualExplainer` to use the setting provided in the `CounterfactualPrediction` removing it from `SolverConfigBuilder`. In essence the parameter is now "per explanation" and not (as previously) "per explainer". I didn't include you on the review as you're on PTO. 

- [x] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>


manstis:
Jenkins retest this

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/CounterfactualExplainabilityRequestDto.java
```diff
@@ -33,24 +33,34 @@
...
     private Map<String, TypedValue> goals;
 
     @JsonProperty(SEARCH_DOMAINS_FIELD)
     @NotNull(message = "searchDomains object must be provided.")
     private Map<String, CounterfactualSearchDomainDto> searchDomains;
 
+    @JsonProperty(MAX_RUNNING_TIME_SECONDS_FIELD)
+    @NotNull(message = "maxRunningTimeSeconds must be provided.")
+    private Long maxRunningTimeSeconds;
+
```
r00ta, line 63: I have only one functional concern about this PR: given that the user does not have (yet?) the ability to **stop** an analysis, I'm wondering if we should accept the request only if `maxRunningTimeSeconds < internalMaxTimeSecondConfiguration` (something like 1h? 2h?). 
Otherwise, the user might send an arbitrary long `maxRunningTimeSeconds` and potentially do suicide. In addition to that, if `maxRunningTimeSeconds > 120s` the kafka connector will timeout (https://github.com/kiegroup/kogito-apps/pull/1016).

danielezonca: This PR only moves ownership of timeout from explainable-service to trusty-service so we still have control of the duration. Btw I agree to add a check to verify that the requested timeout is lower than Kafka timeout

manstis: @r00ta @danielezonca I [added](https://github.com/kiegroup/kogito-apps/pull/1020/commits/12891fdb793b1a9d1549741a4bccb69b8d09a442) the check in `explainability-service` in the end. This is where the messaging sub-system is configured so I can more easily check that the Counterfactual request parameter is less that the SmallRye configuration. Otherwise I'd have had to either duplicate the property in `trusty-service` that would have meant Users would need to configure it twice.. and it all seemed a bit more error prone and messy to me.. However... if you'd still prefer the check in `trusty-service` let me know.

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/handlers/CounterfactualExplainerServiceHandler.java
```diff
@@ -85,14 +94,23 @@ public CounterfactualExplainerServiceHandler(CounterfactualExplainer explainer,
 
     @Override
     public CounterfactualExplainabilityRequest explainabilityRequestFrom(CounterfactualExplainabilityRequestDto dto) {
+        Long maxRunningTimeSeconds = dto.getMaxRunningTimeSeconds();
+        if (Objects.nonNull(maxRunningTimeSeconds)) {
+            if (maxRunningTimeSeconds > kafkaMaxRecordAgeSeconds) {
+                LOGGER.info(String.format("Maximum Running Timeout set to '%d's since the provided value '%d's exceeded the Messaging sub-system configuration '%d's.", kafkaMaxRecordAgeSeconds,
+                        maxRunningTimeSeconds, kafkaMaxRecordAgeSeconds));
+                maxRunningTimeSeconds = kafkaMaxRecordAgeSeconds;
+            }
```
r00ta, line 103: My only problem now is that the user is not aware that his timeout is ignored. The auditor/user of the auditUI probably does not even have access to logs (he's going to ask somebody more technical to investigate what's wrong). 
1) This current approach it's ok if the user knows it: we should write it somewhere in the UI. For example the user might use a slider.. where the max value is this one.
2) OR.. We return 400 when the CF request is made to the trusty service

In any case, this timeout conf would be duplicated in the trusty service or in the UI, we just have to decide where. 
I'm just wondering how to make this behaviour clear to the user that only access the auditUI

explainability/explainability-api/src/main/java/org/kie/kogito/explainability/api/CounterfactualExplainabilityRequestDto.java
r00ta: @manstis @danielezonca added another review below

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/handlers/CounterfactualExplainerServiceHandler.java
manstis: The User of the Audi UI cannot specify the value of the timeout at the moment.

It is read from a `ConfigProperty` [here](https://github.com/kiegroup/kogito-apps/pull/1020/files#diff-18d6c854a0bc1a7cf463a811a18f84dfac351c59fe77627ffa95cec09d2b466aR94-R95). It is passed to the `explainability-service` in every Counterfactual request. It is also returned to the UI to drive the "progress bar".

When we expose such a setting to the User then we can override the value, as I do now. We'd return a message with the solutions advising them that Counterfactual processing terminated early due to system configuration settings and to contact their administrator... Of course, by the time we do that, it may well be https://issues.redhat.com/browse/FAI-566 is resolved and we don't have to worry about SmallRye expiring messages before Counterfactuals complete and this discussion becomes mute.

r00ta: @manstis Thanks for the link to that config, I missed it. Sorry to insist on this, it's just for my understanding: if we already have this config property in the trusty service https://github.com/kiegroup/kogito-apps/pull/1020/files#diff-18d6c854a0bc1a7cf463a811a18f84dfac351c59fe77627ffa95cec09d2b466aR94-R95 and it is used to drive the user to the maximum timeout he can choose, why can't we use it to validate the CF request?

danielezonca (approved):

explainability/explainability-service/src/main/java/org/kie/kogito/explainability/handlers/CounterfactualExplainerServiceHandler.java
manstis: Hello @r00ta it's good to discuss your concerns as it makes me question my actions and we can then both be certain of the optimal solution being implemented. We have two timeouts: Messaging/Kafka/SmallRye that is set when the application deploys and boots; and Counterfactual execution that is set when a Counterfactual is requested.

The SmallRye timeout is static and cannot be changed once the application is deployed and running. The Counterfactual timeout is meant to be dynamic (one day in the future) and potentially be different for different Counterfactual requests for different models. It needs to be passed back to the UI to drive the "Progess bar". Additionally the Counterfactual timeout cannot be greater or equal to the SmallRye timeout due to the [ACK issue](https://issues.redhat.com/browse/FAI-562).

We could use the SmallRye [timeout](https://github.com/kiegroup/kogito-apps/blob/main/explainability/explainability-service-messaging/src/main/resources/application.properties#L14) for the Counterfactual timeout too and return the value in response to the Counterfactual request however we'd need to make other (more significant) changes.

**Current**
```
UI --[sync request]-----------> Trusty --[async request]-----------------> Explainability[*]
UI <-[response with timeout]--- Trusty 
                                Trusty <-[async response with solutions]-- Explainability
UI --[sync polling]-----------> Trusty
UI <-[response]---------------- Trusty

[*] SmallRye timeout is configured in Explainability
```
**Option A - Future (request to Explainability becomes synchronous)**
```
UI --[sync request]-----------> Trusty --[sync request]------------------> Explainability[*]
UI <-[response with timeout]--- Trusty <-[response with timeout]---------- Explainability 
                                Trusty <-[async response with solutions]-- Explainability
UI --[sync polling]-----------> Trusty
UI <-[response]---------------- Trusty

[*] SmallRye timeout is configured in Explainability
```
**Option B - Future (early response and initial poll)**
```
UI --[sync request]-----------> Trusty --[async request]-----------------> Explainability[*]
UI <-[response]---------------- Trusty 
UI --[sync polling]-----------> Trusty
                                Trusty <-[async response with timeout]---- Explainability
                                Trusty <-[async response with solutions]-- Explainability
UI <-[response]---------------- Trusty

[*] SmallRye timeout is configured in Explainability
```
- Option A is a change to make the "request" mechanism synchronous.
- Option B is to delay the UI from getting the "Progress bar" metadata until after the first async response is available.

When https://issues.redhat.com/browse/FAI-566 is resolved we should not need to worry about the SmallRye timeout and just pass whatever value we like for the Counterfactual timeout. That would mean the check I've put in Explainability could be removed entirely.

r00ta: Thanks @manstis for the very detailed recap :) fine for me :+1:

r00ta (approved):

manstis:
@jiripetrlik can you please review?

If you're not able can you please advise who in QE will.

jiripetrlik (approved):
LGTM

PR #1030, opened 2021-09-22
[FAI-591] Reorganize modules

gitgabrio:
@danielezonca @evacchi @jiripetrlik

See https://issues.redhat.com/browse/FAI-591

Needed by https://github.com/kiegroup/kogito-runtimes/pull/1621


gitgabrio:
@manstis @danielezonca 
See [here](https://github.com/kiegroup/kogito-runtimes/pull/1621#issuecomment-984818825) .
Only thing to still cleanup is the property "executionType" of some bean/jsons (e.g. [here](https://github.com/kiegroup/kogito-apps/blob/3bea57f763771836fe74d8c5585187e734ad7856/trusty/trusty-storage/trusty-storage-api/src/main/java/org/kie/kogito/trusty/storage/api/model/Execution.java#L67))
As said in referenced comment, it also resolve, at the end, to the executed type, so it is exactly the same as the enum I've introduced.
For backward compatibility I kept both (basically, two properties with different names are mapped to the same object), but I would prefer to get rid of it. Do you know if that property is used by final consumers?

manstis:
I suspect you'll need to review:
- https://github.com/kiegroup/kogito-apps/blob/main/trusty/trusty-service/trusty-service-infinispan/src/main/resources/reflection-config.json
- https://github.com/kiegroup/kogito-apps/blob/main/explainability/explainability-service-messaging/src/main/resources/reflection-config.json
- (Possibly the end of this, but I don't think so.... just mentioning so you're aware of its existence) https://github.com/kiegroup/kogito-apps/blob/main/jitexecutor/jitexecutor-dmn/src/main/resources/META-INF/native-image/org.kie.kogito/jitexecutor-dmn/reflect-config.json

manstis (approved):

manstis:
Jenkins run native

manstis:
Native will fail until https://github.com/kiegroup/kogito-apps/pull/1143 is merged.. but it's a good check to ensure Quakus is happy.

gitgabrio:
Jenkins retest this

jiripetrlik (approved):

danielezonca (approved):

gitgabrio:
Jenkins retest this

PR #1277, opened 2022-03-18
KOGITO-6886 Consolidate Event API between Data Index and Runtime

cristianonicolai:
Assembly:
- https://github.com/kiegroup/kogito-apps/pull/1277
- https://github.com/kiegroup/kogito-runtimes/pull/2090
- https://github.com/kiegroup/kogito-examples/pull/1191

PR #1278, opened 2022-03-18
FAI-732: E2E test Audit UI - "Traffic Violation open Audit Details – …

sjamboro:
…open Audit Details" failed in the kogito-apps-deploy job

I have added the clock function to synchronized application and test. Application should get same time which was set by test.

https://issues.redhat.com/browse/FAI-732

### DIFF
--- a/ui-packages/packages/trusty/cypress/e2e/TrafficViolation.ts
+++ b/ui-packages/packages/trusty/cypress/e2e/TrafficViolation.ts
@@ -45,6 +45,8 @@ describe('Traffic Violation', () => {
   });
 
   it('open Audit Details', () => {
+    const plus3DaysShift = new Date().getTime() + 3*60*60*24*1000; 
+    cy.clock(plus3DaysShift);
     cy.visit('/');
     cy.ouiaId(reqId, 'PF4/TableRow', {timeout: 15000}).within(() => {
       cy.ouiaId('status', 'execution-status').should('have.text', 'Completed');


ui-packages/packages/trusty/cypress/e2e/TrafficViolation.ts
```diff
@@ -45,6 +45,8 @@ describe('Traffic Violation', () => {
   });
 
   it('open Audit Details', () => {
+    cy.clock(new Date());
```
kelvah: Shouldn't you pass a timestamp? Like [docs show here](https://docs.cypress.io/api/commands/clock#Now)

sjamboro: @kelvah Maybe I misunderstend the idea:
1. I send a request in the before method: https://github.com/kiegroup/kogito-apps/blob/d78457fb8d795ca02b13ea11e0fedd8ea0c219c2/ui-packages/packages/trusty/cypress/e2e/TrafficViolation.ts#L21-L44
2. I set a current date (in test's point of view, several seconds later than the answer is recieved). Synchronized time with application (I follow [this example](https://docs.cypress.io/api/commands/clock#Create-a-clock-and-use-it-to-trigger-a-setInterval))
3. Addition of one minute: https://github.com/kiegroup/kogito-apps/blob/d78457fb8d795ca02b13ea11e0fedd8ea0c219c2/ui-packages/packages/trusty/cypress/e2e/TrafficViolation.ts#L49
4. Visit a page ... the application should read the synchronized time: https://github.com/kiegroup/kogito-apps/blob/d78457fb8d795ca02b13ea11e0fedd8ea0c219c2/ui-packages/packages/trusty/src/components/Templates/AuditOverview/AuditOverview.tsx#L27-L29

IMHO in case that I want to change date I should put it into the "before" method. Do you thing that `cy.clock()` will work for the sended request, too? Otherwise I will need to add some day (like `cy.clockl(new Date() + 1 day)`). `cy.tick()` should work similar. I can rise it from 60 000 to 15 000 000 (from 1 minute to 4 hours).

kelvah: Hi @sjamboro,
I was merely pointing out that you passed a Date object to `cy.clock`, while the docs say it should be a timestamp, like `new Date().getTime()` from [the link I shared](https://docs.cypress.io/api/commands/clock#Now).

Another thing I noticed while reading your explanation,  `cy.tick` only affects running timers, like `setInterval` or `setTimeout`. We don't use timers in the source code... so probably it's not what you want to use.

sjamboro: @kelvah thank you, I can see it now. I thought that you want to define the exact time like `new Date(2017, 3, 14)`. In case that `cy.tick()` has no effect on our UI than I need to add some hours in Date object or try to find out how to deal with EDT timezone.

sjamboro:
@kelvah @jstastny-cz Can you review this issue? I wanted to check this issue locally but I had a problem with running zookeeper via docker (I have to migrate to podman). Now, I can say that `cy.clock` really affect the values of the date filter (+ 3 days). 

@radtriste I saw https://github.com/kiegroup/kogito-apps/pull/1277, I am not sure if it could fix the problem (it contains lots of `DateTimeUtils.parseZonedDateTime` calls - IMPOV that changes look like refactoring). I can not see any run of nigthly build .

sjamboro:
There is a screenshot from my local run (The test request was sent the 1st of April and the filter contains the 4th of April ):
image

jstastny-cz (approved):
LGTM.

kelvah (approved):

PR #1386, opened 2022-06-15
KOGITO-7360 improve ui packages build and create profile

AjayJagan:
- JIRA : https://issues.redhat.com/browse/KOGITO-7360
 - Ignore certain packages in `ui-packages` for swf builds
 - create a productization profile in `ui-packages` pom

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,22 @@
       </modules>
     </profile>
     <profile>
+      <id>productized</id>
+      <activation>
+        <property>
+          <name>productized</name>
+        </property>
+      </activation>
+      <modules>
+        <module>kogito-apps-bom</module>
+        <module>kogito-apps-build-parent</module>
+        <module>persistence-commons</module>
+        <module>data-index</module>
+        <module>ui-packages</module>
+        <module>swf-quarkus-extension-parent</module>
+      </modules>
+    </profile>
+    <profile>
       <id>optaplanner-downstream</id>
       <activation>
         <property>
--- a/pom.xml
+++ b/pom.xml
@@ -1,96 +1,99 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xmlns="http://maven.apache.org/POM/4.0.0"
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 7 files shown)


ui-packages/pom.xml
```diff
@@ -219,4 +221,166 @@
...
+            </configuration>
+          </execution>
+          <execution>
+            <id>yarn build:swf</id>
+            <goals>
+              <goal>yarn</goal>
+            </goals>
+            <configuration>
+              <skip>${skip.ui.build}</skip>
+              <arguments>run build:prod</arguments>
```
cristianonicolai: build:prod:swf ?
```suggestion
              <arguments>run build:prod:swf</arguments>
```

ui-packages/pom.xml
```diff
@@ -219,4 +221,166 @@
...
+          </execution>
+          <execution>
+            <id>yarn test</id>
+            <goals>
+              <goal>yarn</goal>
+            </goals>
+            <phase>test</phase>
+            <configuration>
+              <skip>${skip.ui.test}</skip>
+              <arguments>run test:swf</arguments>
```
cristianonicolai: @AjayJagan it seems that only the commands that you're executing differ from one profile to another. Maybe instead of duplicating it all, I would suggest using maven properties to set the command, in that way you simply change the value of the property ( command to be executed ) depending on the profile.

ui-packages/pom.xml
AjayJagan: yea! sorry my bad ;)

ui-packages/pom.xml
AjayJagan: used profiling to set properties :+1:

pom.xml
```diff
@@ -80,5 +80,20 @@
...
+	<!--need to remove comment after merging/rebasing KOGITO-7321(Cristiano's PR) -->
+	<profile>
+	 <id>productized</id>
+	 <activation>
+	   <property>
+		 <name>productized</name>
+		</property>
+	 </activation>
+	 <modules>
+	  <module>kogito-apps-bom</module>
```
pefernan, line 108: I think we may need data-index here too!

@cristianonicolai @AjayJagan  wdyt?

cristianonicolai: yep, I'll fix the actual PR for prod build

cristianonicolai:
@AjayJagan main PR merged, please rebase

AjayJagan:
reopening in another pr

PR #1388, opened 2022-06-18
Kogito 7360 improve build

AjayJagan:
- JIRA : https://issues.redhat.com/browse/KOGITO-7360
 - Ignore certain packages in `ui-packages` for swf builds
 - create a productization profile in `ui-packages` pom

### DIFF
--- a/ui-packages/package.json
+++ b/ui-packages/package.json
@@ -17,8 +17,8 @@
     "locktt": "locktt",
     "test:e2e:trusty": "lerna run --scope @kogito-apps/trusty test:e2e:trusty --",
     "format:all": "lerna run format --stream --",
-    "build:prod:swf":"lerna run --ignore={@kogito-apps/common,@kogito-apps/task-console-webapp,@kogito-apps/management-console,@kogito-apps/management-console-webapp,@kogito-apps/cypress-ouia} build:prod --stream --",
-    "test:swf":"lerna run --ignore={@kogito-apps/common,@kogito-apps/task-console-webapp,@kogito-apps/management-console,@kogito-apps/management-console-webapp,@kogito-apps/cypress-ouia} test --stream --"
+    "build:prod:swf":"lerna run --scope='@kogito-apps/runtime-tools-dev-ui-webapp' build:prod --include-dependencies --stream --",
+    "test:swf":"lerna run --scope='@kogito-apps/runtime-tools-dev-ui-webapp' test --include-dependencies --stream --"
   },
   "husky": {
     "hooks": {
--- a/ui-packages/package.json
+++ b/ui-packages/package.json
@@ -16,7 +16,9 @@
     "test:it": "lerna run test:it --stream --concurrency=1 --no-bail --",
     "locktt": "locktt",
     "test:e2e:trusty": "lerna run --scope @kogito-apps/trusty test:e2e:trusty --",
-    "format:all": "lerna run format --stream --"
+    "format:all": "lerna run format --stream --",
+    "build:prod:swf":"lerna run --ignore={@kogito-apps/common,@kogito-apps/task-console-webapp,@kogito-apps/management-console,@kogito-apps/management-console-webapp,@kogito-apps/cypress-ouia} build:prod --stream --",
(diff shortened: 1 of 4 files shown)


ui-packages/pom.xml
```diff
@@ -44,6 +44,33 @@
       packages/process-definition-list/coverage/lcov.info
     </sonar.javascript.lcov.reportPaths>
   </properties>
+
+  <profiles>
+  <profile>
```
cristianonicolai: just review the format here, otherwise looks good.

cristianonicolai:
jenkins run quarkus-lts

ui-packages/pom.xml
AjayJagan: done.. please let me know if the formatting is okay.

AjayJagan:
jenkins run quarkus-lts

pefernan (approved):
+1 thanks @AjayJagan

cristianonicolai (approved):

ui-packages/package.json
```diff
@@ -16,7 +16,9 @@
     "test:it": "lerna run test:it --stream --concurrency=1 --no-bail --",
     "locktt": "locktt",
     "test:e2e:trusty": "lerna run --scope @kogito-apps/trusty test:e2e:trusty --",
-    "format:all": "lerna run format --stream --"
+    "format:all": "lerna run format --stream --",
+    "build:prod:swf":"lerna run --ignore={@kogito-apps/common,@kogito-apps/task-console-webapp,@kogito-apps/management-console,@kogito-apps/management-console-webapp,@kogito-apps/cypress-ouia} build:prod --stream --",
```
paulovmr: @AjayJagan I believe it would be better to use the scope + include-dependencies instead of the ignore to rely on the dependencies to execute the build, like: `lerna run --scope="@kogito-apps/runtime-tools-dev-ui-webapp" build:prod --include-dependencies --stream --`

ui-packages/package.json
```diff
@@ -16,7 +16,9 @@
...
+    "test:swf":"lerna run --ignore={@kogito-apps/common,@kogito-apps/task-console-webapp,@kogito-apps/management-console,@kogito-apps/management-console-webapp,@kogito-apps/cypress-ouia} test --stream --"
```
paulovmr: Same here.

paulovmr (changes requested):

ui-packages/pom.xml
```diff
@@ -44,6 +44,33 @@
...
+    <id>productized</id>
+    <activation>
+     <property>
+      <name>productized</name>
+     </property>
+    </activation>
+    <properties>
+     <yarn.build.script>yarn run build:prod:swf</yarn.build.script>
+     <yarn.test.script>yarn run test:swf</yarn.test.script>
+     <skip.ui.it>true</skip.ui.it>
```
jstastny-cz, line 69: Is there a plan on how to run those tests with the built packages ex-post? Is it so problematic to keep integration tests turned on by default?

AjayJagan: hello @jstastny-cz  - thanks for the review. The reason why I skipped the IT tests for this build is that this `productized` build profile only builds the packages required for the serverless workflow dev ui, which does not have any IT tests as of now. Even if we enable it, I think it won't make much difference I guess. So please let me know if you want to enable it. :)

ui-packages/package.json
AjayJagan: hey @paulovmr , initially I did that :) but the number of packages to be included in the scope was more , so I did the opposite(ignore them). :P. I shall change it to scope now. Thanks

AjayJagan: changed to scope

ui-packages/package.json
AjayJagan: changed to scope

paulovmr (approved):

ui-packages/pom.xml
jstastny-cz: I am fine with the proposed solution right now. When we have some integration tests to run, we need to figure out if enable here as part of productized build (easier) or figure out how to run ex-post with already built packages.

PR #1503, opened 2022-11-10
[KOGITO-7208] fix throughput issue under load in job-service

tiagodolphine:
I executed a load test with a process containing Async task + Timer Task. 
With 50 concurrent threads and a loop of 1000 iterations (creating process instances that trigger jobs for timers + async task), it was noticed a bunch of errors regarding the downstream when the jobs were executed, on the emitters on`JobStreams`.
After some investigations and looking at the smallrye doc, I added the Overflow annotation to disable backpressure worked well, which means before there was an in-memory buffer to control the jobs after they are triggered by the scheduler (vertx), and it was leading to exceptions in this buffer with high load. Disabling the buffer means now all jobs are sent to downstream without backpressure using the buffer, this increased the throughput and the load test run without errors. 

Related to
https://issues.redhat.com/browse/KOGITO-7208
https://issues.redhat.com/browse/KOGITO-7209

- [ ] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
</summary>

- for <b>pull request checks</b>  

- for a <b>specific pull request check</b>  
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] tests</b>

- for <b>quarkus branch checks</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins run quarkus-branch</b>

- for a <b>quarkus branch specific check</b>  
  Run checks against Quarkus current used branch  
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] quarkus-branch</b>

- for <b>quarkus main checks</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins run quarkus-main</b>

- for a <b>specific quarkus main check</b>  
  Run checks against Quarkus main branch  
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] quarkus-main</b>

- for <b>quarkus lts checks</b>  
  Run checks against Quarkus lts branch  
  Please add comment: <b>Jenkins run quarkus-lts</b>

- for a <b>specific quarkus lts check</b>  
  Run checks against Quarkus lts branch  
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] quarkus-lts</b>
 
- for <b>native checks</b>  
  Run native checks  

- for a <b>specific native check</b>  
  Run native checks 
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] native</b>

- for <b>mandrel checks</b>  
  Run native checks against Mandrel image
  Please add comment: <b>Jenkins run mandrel</b>

- for a <b>specific mandrel check</b>  
  Run native checks against Mandrel image  
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] mandrel</b>

- for <b>mandrel lts checks</b>  
  Run native checks against Mandrel image and quarkus lts branch
  Please add comment: <b>Jenkins run mandrel-lts</b>

- for a <b>specific mandrel lts check</b>  
  Run native checks against Mandrel image and quarkus lts branch
  Please add comment: <b>Jenkins (re)run [kogito-apps|kogito-examples] mandrel-lts</b>
 
</details>

### DIFF
--- a/jobs-service/jobs-service-common/src/main/java/org/kie/kogito/jobs/service/stream/JobStreams.java
+++ b/jobs-service/jobs-service-common/src/main/java/org/kie/kogito/jobs/service/stream/JobStreams.java
@@ -45,7 +45,7 @@ public class JobStreams {
      */
     @Inject
     @Channel(AvailableStreams.JOB_ERROR)
-    @OnOverflow(value = OnOverflow.Strategy.BUFFER, bufferSize = 10000)
+    @OnOverflow(value = OnOverflow.Strategy.NONE)
     Emitter<JobExecutionResponse> jobErrorEmitter;
 
     /**
@@ -53,6 +53,7 @@ public class JobStreams {
      */
     @Inject
     @Channel(AvailableStreams.JOB_SUCCESS)
+    @OnOverflow(value = OnOverflow.Strategy.NONE)
     Emitter<JobExecutionResponse> jobSuccessEmitter;
 
     /**
@@ -60,6 +61,7 @@ public class JobStreams {
      */
     @Inject
     @Channel(AvailableStreams.JOB_STATUS_CHANGE)
+    @OnOverflow(value = OnOverflow.Strategy.NONE)
     Emitter<JobDetails> jobStatusChangeEmitter;
 
     public JobExecutionResponse publishJobError(JobExecutionResponse response) {


wmedvede (commented):
Hi  @tiagodolphine , so the problem here was that at some point, no more jobs where executed because the some of this channels reached the bufferSize (defaultValue = 128) and thus all the stream chain was stuck?

tiagodolphine:
> Hi @tiagodolphine , so the problem here was that at some point, no more jobs where executed because the some of this channels reached the bufferSize (defaultValue = 128) and thus all the stream chain was stuck?

from my local tests yes, with the buffer at some point it reaches a limit, and after that exceptions start being thrown.. with the annotation to disable the buffer the events go directly downstream with no buffer for the backpressure thing on the emitters. In this case, the results were better, at least with the load I mentioned, with no exceptions.

wmedvede:
> > Hi @tiagodolphine , so the problem here was that at some point, no more jobs where executed because the some of this channels reached the bufferSize (defaultValue = 128) and thus all the stream chain was stuck?
> 
> from my local tests yes, with the buffer at some point it reaches a limit, and after that exceptions start being thrown.. with the annotation to disable the buffer the events go directly downstream with no buffer for the backpressure thing on the emitters. In this case, the results were better, at least with the load I mentioned, with no exceptions.

ok, thanks for the explanation.

wmedvede (approved):

radtriste:
this is a small test

MarianMacik (approved):
Looks good. Just a question what happens now if there is an overflow? Will the downstream part continue execution until the app runs out of resources or even emitter code will hang until there is enough resources on the downstream size to continue execution?

tiagodolphine:
> Looks good. Just a question what happens now if there is an overflow? Will the downstream part continue execution until the app runs out of resources or even emitter code will hang until there is enough resources on the downstream size to continue execution?

once the job is triggered by the scheduler it is sent downstream without the buffer, before the change there was a buffer to control the backpressure but with a higher load the issue started to happen.
The best way would be to control the backpressure from the scheduler using not a buffer but probably a queue (on kafka for instance) we can think about this when we change the architecture from job-service.
For now, I think this approach with no buffers performs better, at least compared to the current one.

PR #1573, opened 2022-12-21
KOGITO-8394 - Data Index attributes retrieved from runtime should ret…

cristianonicolai:
…urn null instead of 404 error


nmirasch (approved):

pefernan (approved):
+1 thanks @cristianonicolai !

radtriste (approved):

PR #1802, opened 2023-07-04
KOGITO-9523 - Move Data Index DevService test to kogito-apps

cristianonicolai:
Assembly:
- https://github.com/kiegroup/kogito-runtimes/pull/3100
- https://github.com/kiegroup/kogito-apps/pull/1802

### DIFF
--- a/apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus-devservice/pom.xml
+++ b/apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus-devservice/pom.xml
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>integration-tests-data-index-service</artifactId>
+    <groupId>org.kie.kogito</groupId>
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>integration-tests-data-index-service-quarkus-devservice</artifactId>
+  <name>Kogito Apps :: Integration Tests :: Data Index :: Quarkus DevService</name>
+
+  <properties>
+    <data-index-ephemeral.image.test>org.kie.kogito/data-index-service-inmemory:${project.version}</data-index-ephemeral.image.test>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-quarkus-serverless-workflow</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-jackson</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-quarkus-test-utils</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-junit5</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-junit5-internal</artifactId>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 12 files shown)


cristianonicolai:
jenkins run quarkus-lts

nmirasch (approved):

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus-devservice/src/test/java/org/kie/kogito/index/inmemory/KogitoDevServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+import static io.restassured.RestAssured.given;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+
+public class KogitoDevServiceTest {
+
+    private static final Duration TIMEOUT = Duration.ofMinutes(1);
+    private static final String DATA_INDEX_EPHEMERAL_IMAGE_TEST_PROPERTY = "data-index-ephemeral.image.test";
+    private static final String DATA_INDEX_EPHEMERAL_IMAGE_TEST_DEFAULT_VALUE = "use-default-image";
```
MarianMacik, line 41: I think we no longer need this. Tests will always use the specified image.
```suggestion
```

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus-devservice/src/test/java/org/kie/kogito/index/inmemory/KogitoDevServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+                + "quarkus.kogito.devservices.port=" + dataIndexHttpPort + "\n"
+                + getKogitoDevServicesImageName();
+    }
+
+    private static String getKogitoDevServicesImageName() {
+        String imageName = System.getProperty(DATA_INDEX_EPHEMERAL_IMAGE_TEST_PROPERTY);
+        if (imageName == null) {
+            throw new IllegalStateException("Property " + DATA_INDEX_EPHEMERAL_IMAGE_TEST_PROPERTY + " was not passed to tests!");
+        }
+        if (imageName.isEmpty() || DATA_INDEX_EPHEMERAL_IMAGE_TEST_DEFAULT_VALUE.equals(imageName)) {
```
MarianMacik, line 111:
```suggestion
        if (imageName.isEmpty()) {
```

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus-devservice/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    <groupId>org.kie.kogito</groupId>
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>integration-tests-data-index-service-quarkus-devservice</artifactId>
+  <name>Kogito Apps :: Integration Tests :: Data Index :: Quarkus DevService</name>
+
+  <properties>
+    <data-index-ephemeral.image.test>org.kie.kogito/data-index-service-inmemory:${project.version}</data-index-ephemeral.image.test>
```
MarianMacik, line 16: I like that for PR checks we will have an instant response now. Though for nightlies we should probably override this with the nightly image in our automation. This means that nightlies will run with the nightly image from the previous nightly but will check the cekit-built image. For the "current" changes we will still have the PR checks. 

Moreover, the release could be also checked with the latest nightly as before and committed here as it was until now. Afterwards, when the version is promoted, we can put it back to this internally-built image.

Or we can have a separate stage in the pipeline which will run just this test after the images are built.

What do you think?
cc @radtriste

MarianMacik (approved):

apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-quarkus-devservice/pom.xml
cristianonicolai: @MarianMacik agree with overriding for testing nightly with the previous image. As for the release Im not sure, would be best to use the image from the current build. I can maybe thing of something like, one stage to build artefacts from kogito-apps without tests, you then build images and back to apps to run full cycle? Not sure to be honest. Maybe @radtriste wold have a better idea.

MarianMacik: @cristianonicolai  Yes, that would be best. While I was thinking about it I realized that we should leave the option to use the `use-default-image` value because this will be used for the nightlies when we want to rely on what is predefined in the `data-index-devservice.properties`. Can you revert it, please?

cristianonicolai: @MarianMacik done

radtriste (approved):

PR #817, opened 2021-05-13
Fix snapshot for test failure

manstis:
Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](https://github.com/kiegroup/kogito-runtimes#contributing-to-kogito)
- [ ] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [ ] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Pull Request</b>  
  Please add comment: <b>Jenkins retest this</b>
 
* <b>Quarkus LTS checks</b>  
  Please add comment: <b>Jenkins run LTS</b>

* <b>Native checks</b>  
  Please add comment: <b>Jenkins run native</b>

</details>

PR #1968, opened 2024-01-23
#817 - removing oracle database dependency

porcelli:
This PR is related to https://github.com/apache/incubator-kie-issues/issues/817

Related PR:

- https://github.com/apache/incubator-kie-kogito-runtimes/pull/3369


pefernan (approved):

fjtirado:
@porcelli I see that rather than removing the Oracle driver, we are completely removing Oracle db support. 
To be honest, I would have been more conservative, because there is not actually source code depending on oracle stuff to support oracle (since we are using JPA). 
So, It would have been posible to leave references to Oracle and just remove the driver from the repo.

porcelli:
@fjtirado i just wrote this to ML:

I might be interpreting the Category X dependency way too strictly; but it’s my current understanding that we cannot have such dependency in our codebase at all.

The direction that we agreed is to avoid endless discussion and just remove for now; once we past the critical phase of our first release- we can try to explore and discuss different angles to tackle the Category X dependency; it’s just not worth spend time right now on it.

fjtirado:
@porcelli 
We spent quite amount of time on avoiding code duplication for Oracle and Postgresql and we will have to spent again when in future community ask for oracle support (as they already did in the past, the oracle support, a lot of copy paste from postgres that was later refactor, was added by community folks on their spare time) . I do not really see the problem on just removing the driver and test and keep the source code that does not depend on the driver, so people can include the addon and the driver in their projects.

porcelli:
note: community may ask for support and we might not be able to cover the request for legal reasons....

my perspective is to not have unused code left behind just because... that's the value of version control (we can always revert if different interpretation of category X is provided)

fjtirado:
Already implemented

PR #2090, opened 2024-08-23
Fix #3634 Update Client keycloak in Management Console & Task Console 

tru3d3v:
Hi There,

according to this ticket number 3634 I had already update for the client keycloak-js@25.0.4 with disable few configuration in crypto:

resolve: {
fallback: {
crypto: false // Disable polyfilling the 'crypto' module
}
}


pefernan:
Closing since code has been moved to kie-tools

PR #2199, opened 2025-02-27
Avoid potential concurrent exception

fjtirado:
If the list of events in the published (which global) gets modified by another test, then a concurrent exception will be thrown. In other words, the test is potentially flaky, so it is safer to copy the list to check it holds the expected data.


wmedvede (approved):

gmunozfe (approved):
LGTM, well done @fjtirado !

PR #2240, opened 2025-07-10
[kie-issues#2031] Set processId from ProcessInstanceStateDataEvent

### LINKED ISSUE #2031: ProcessInstanceStateDataEvent with updated processId not reflected in data-index JPA storage
When we execute a Process Instance Migration, the workflow runtime engine sends a `ProcessInstanceStateDataEvent`. This event has the information about the new (migrated) process: `processId`, `processName`, `version`. This event then gets processed by the data-index, however in the JPA module the existing entity is not updated with the new  `processId` .

martinweiler:
Closes https://github.com/apache/incubator-kie-issues/issues/2031


pefernan (approved):

jomarko (approved):
Thank you @martinweiler . 

Just for a record. This PR will solve the issue of `processId` propagation just for a process instance migration, that is backed by a MPF file stored in the business service project.

`ProcessInstanceStateDataEvent` is not published at all if user interact with process instance that was migrated without related MPF file.

ljmotta (approved):

martinweiler:
@jomarko you are right, thanks for pointing this out. This will be addressed in a separate PR, depending on what we decide regarding the time the event will be sent:
1. Keep it as-is, sending the event at the time the migrated process instance is deserialized. In this case, we just need to make sure the event is sent regardless of the existence of a mpf file.
2. Change the logic when the event is sent to the invocation of the `migrate` endpoint.

PR #2303, opened 2026-02-19
[incubator-kie-issues#2256] Fix Quarkus Job Exception Details Extractor injection

### LINKED ISSUE #2256: Quarkus WrappingConditionalJobExceptionDetailsExtractor fails to inject instances
The class has currently wrong annotations to collect instances, it mixes `@All` with `Instance<>` fields, which does not work.

It should be either `@All` + `List<>` or `@Any` + `Instance<>`.

This results in the `WrappingConditionalJobExceptionDetailsExtractor` using as delegate `DefaultJobExceptionDetailsExtractor` instance all the time.

jstastny-cz:
Fixes: apache/incubator-kie-issues#2256

The problem was mixup of 2 possible approaches.

Adding also tests for the injecting part.


martinweiler (approved):
Good catch @jstastny-cz

pefernan (approved):

PR #2325, opened 2026-04-23
incubator-kie-issues#2286: isolate and test data-index storage modules

### LINKED ISSUE #2286: Resolve data-index-storage-jpa-springboot dependency clash and test coverage
The module `data-index-storage-jpa-springboot` in kogito-apps does not have the same test coverage as `data-index-storage-jpa`.

Also, Spring Boot relevant modules in `data-index-springboot` import internal dependencies, that bring in quarkus libraries (at least for tests).

This task aims at removing framework-specific items from `data-index-storage-jpa` and use it as dependency to both  `data-index-storage-jpa-springboot` and (newly added) `data-index-storage-jpa-quarkus` modules to achieve isolation.

jstastny-cz:
Closes apache/incubator-kie-issues#2286

Extending test coverage for Spring Boot Data Index Storage JPA module by replicating the existing coverage from Quarkus variant.

To achieve that, existing modules which were closely aligned with Quarkus (especially annotations) needed to be refactored:
* quarkus-specific annotations, class references (e.g. `Instance<>`), tests and test dependencies moved out from `data-index-storage-jpa-common` and `data-index-storage-jpa` to newly added `data-index-storage-jpa-quarkus` module.
* `data-index-storage-jpa-common` is still dependency of both `data-index-storage-jpa` and `data-index-storage-postgresql`, but does NOT provide any quarkus annotated beans.
* `data-index-storage-jpa-quarkus` and `data-index-storage-jpa-springboot` now
  * use `data-index-storage-jpa` as dependency (still without framework specific beans).
  * take care of defining framework-specific beans by constructor injection in framework-specific classes, but instantiating the shared classes using regular constructor `super(...)`
  * contain equivalent test coverage by extending abstract test classes and providing framework-specific configuration and annotations.
* `data-index-storage-jpa` and `data-index-storage-postgresql` need both quarkus-specific beans, so there's some level of duplication, but it's a trivial class inheritance, injection and invocation of super-class constructor with injected values.
* The shared components and tests still use classes and annotations supported by both Quarkus and Spring Boot, e.g. `@Transactional`.


jstastny-cz:
There are still the abstract tests using Inject and ApplicationScoped annotations, need to refactor them too to pass instances via constructor from the framework-specific test subclasses.

pefernan (approved):

martinweiler (approved):

PR #2328, opened 2026-04-30
[Incubator-kie-issues#2323] Allow filtering schema on nested arrays

### LINKED ISSUE #2323: Allow filtering schema on nested arrays
### Describe the bug

Data Index GraphQL schema does define nested array fields for top-level types, e.g. nodes of type ProcessInstance queried by ProcessInstances query which allows filtering via input ProcessInstanceArgument
problem is, that nodes in ProcessInstanceArgument is implemented by NodeInstanceArgument, which is not though designed as an array type
As a result, when querying for process instances based on a filter over nodes field, it does not allow definition of predicates as contains, contains-all, contains-any, etc. Which is limiting in terms of being able to compose custom queries.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Output of `uname -a` or `ver`

_No response_

### Output of `java -version`

_No response_

### GraalVM version (if different from Java)

_No response_

### Kogito version or git rev (or at least Quarkus version if you are using Kogito via Quarkus platform BOM)

_No response_

### Build tool (ie. output of `mvnw --version` or `gradlew --version`)

_No response_

### Additional information

_No response_

GopikaReghunath:
closes: https://github.com/apache/incubator-kie-kogito-apps/issues/2323

New Array Argument Types: Created dedicated array argument types for nested collections

created new GraphQL input types specifically designed for filtering array fields. These are different from the regular input types used for scalar fields.
Each array argument type supports contains, containsAll, containsAny, isNull, and/or/not

Added mapArrayArgument() method to handle array-specific filtering in GraphQLQueryMapper.java
The mapArrayArgument() method is the core logic that translates GraphQL array operations into database queries.
Step 1: Detect Array Operation
Array operations need different SQL patterns than scalar operations.
Step 2: Extract Operation Type
Each operation requires a different SQL pattern.

Generates appropriate JPA Criteria API predicates for each operation
Working of different scenarios:

```
containsAll: "Has ALL of these"
Each JOIN finds instances with one specific element from the collection
AND combines them: only instances with ALL specified elements are returned
Database execution: Creates multiple JOINs (one per element), combines with AND

```
```
containsAny: "Has ANY of these"
Each JOIN finds instances with one specific element from the collection
OR combines them: instances with AT LEAST ONE of the specified elements are returned
Database execution: Creates multiple JOINs (one per element), combines with OR

```
```
NOT: "Doesn't have these"
Subquery finds instances that DO have the specified elements (using the positive condition)
NOT EXISTS inverts it: returns instances that DON'T have the specified elements
Original code applied NOT at row level (wrong for collections)
New code applies NOT at entity level using NOT EXISTS (correct for collections)

```
To summarise:
GraphQLQueryMapper.mapArrayArgument()
Location: data-index-graphql module

Role: GraphQL → Database-Agnostic Filters

What it does:

✅ Handles array operations (contains, containsAll, containsAny, isNull, and/or/not)
✅ Parses GraphQL queries with nested array filtering
✅ Converts to AttributeFilter objects

JPAQuery.java
Location: data-index-storage-jpa-common module

Role: Database-Agnostic Filters → SQL

What it does:

✅ Receives AttributeFilter objects from GraphQLQueryMapper
✅ Converts to JPA Criteria API predicates (which generate SQL)
✅ Handles ALL filter types (not just arrays - also strings, dates, numbers, etc.)
✅ CRITICAL FIX: Special handling for NOT operations on collections
❌ Old behavior: builder.not() caused row-level negation (wrong for collections)
✅ New behavior: NOT EXISTS subquery causes entity-level negation (correct!)

Testing: 
This can be tested with the help of GraphQL Queries. Verify functionalities like
Test 1: Basic contains
Test 2: containsAll (Multiple conditions with AND)
Test 3: containsAny (Multiple conditions with OR)
Test 4: NOT contains (Negation)
Test 5: isNull (Empty collection check)
Test 6: Mixed Filter (Regular + Collection)
Test 7: Cross-Collection (Multiple collections)
Test 8: Simple Array (containsAny)


Kusuma04-dev:
Thanks @GopikaReghunath  for the PR,  Can you please share the steps how to test this?

GopikaReghunath:
> Thanks @GopikaReghunath for the PR, Can you please share the steps how to test this?

done

martinweiler (approved):

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
```diff
@@ -82,9 +83,9 @@ public GraphQLQueryParser apply(GraphQLInputObjectType type) {
                         parser.mapAttribute(field.getName(), mapEnumArgument(field.getName()));
                     } else if (isListOfType(field.getType(), type.getName())) {
                         parser.mapAttribute(field.getName(), mapRecursiveListArgument(field.getName(), parser));
-                    } else if (((GraphQLNamedType) field.getType()).getName().equals(type.getName())) {
+                    } else if (field.getType() instanceof GraphQLNamedType && ((GraphQLNamedType) field.getType()).getName().equals(type.getName())) {
```
jstastny-cz: wouldn't this work too?

```suggestion
                    } else if (field.getType() instanceof GraphQLNamedType namedType && namedType.getName().equals(type.getName())) {
```

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
```diff
@@ -82,9 +83,9 @@ public GraphQLQueryParser apply(GraphQLInputObjectType type) {
...
                         parser.mapAttribute(field.getName(), mapRecursiveArgument(field.getName(), parser));
-                    } else {
+                    } else if (field.getType() instanceof GraphQLNamedType) {
                         String name = ((GraphQLNamedType) field.getType()).getName();
```
jstastny-cz:
```suggestion
                    } else if (field.getType() instanceof GraphQLNamedType namedType) {
String name = namedType.getName();
```

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
```diff
@@ -116,7 +117,13 @@ public GraphQLQueryParser apply(GraphQLInputObjectType type) {
                                 break;
                             default:
                                 if (field.getType() instanceof GraphQLInputObjectType) {
-                                    parser.mapAttribute(field.getName(), mapSubEntityArgument(field.getName(), new GraphQLQueryMapper().apply((GraphQLInputObjectType) field.getType())));
+                                    GraphQLInputObjectType inputType = (GraphQLInputObjectType) field.getType();
```
jstastny-cz: also here we could use pattern matching

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLInputObjectTypeMapper.java
```diff
@@ -108,7 +112,20 @@ private GraphQLInputType getInputTypeByField(GraphQLFieldDefinition field) {
                 return getInputObjectType("BooleanArgument");
             case "DateTime":
                 return getInputObjectType("DateArgument");
+            case "StringArray":
+                return getInputObjectType("StringArrayArgument");
             default:
+                // For array fields, try to use ArrayArgument type first
+                if (isArray) {
+                    String arrayTypeName = name + "ArrayArgument";
```
jstastny-cz: Seein ArrayArgument literal also further, this would deserve a constant somewhere.

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
```diff
@@ -229,6 +236,161 @@ private Function<Object, Stream<AttributeFilter<?>>> mapSubEntityArgument(String
...
+            GraphQLInputObjectType elementType = (GraphQLInputObjectType) unwrappedType;
+            GraphQLQueryParser elementParser = new GraphQLQueryMapper().apply(elementType);
+
+            // Separate array operations from backward-compatible fields
+            Map<String, Object> backwardCompatFields = new java.util.HashMap<>();
+            List<Stream<AttributeFilter<?>>> arrayOpStreams = new java.util.ArrayList<>();
+
+            for (Map.Entry<String, Object> entry : argMap.entrySet()) {
+                String operation = entry.getKey();
+                Object value = entry.getValue();
```
jstastny-cz, line 271: we should use as other methods:
```
            FilterCondition condition = FilterCondition.fromLabel(entry.getKey());
            if (entry.getValue() == null) {
                return null;
            }
            switch (condition) {
                case IN:
                    return filterValueList(entry.getValue(), value -> in(attribute, value));
                case EQUAL:
                    return equalTo(attribute, entry.getValue().toString());
                case IS_NULL:
                    return Boolean.TRUE.equals(entry.getValue()) ? isNull(attribute) : notNull(attribute);
                default:
                    return null;
            }
```
right? instead of string values.

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
```diff
@@ -229,6 +236,161 @@ private Function<Object, Stream<AttributeFilter<?>>> mapSubEntityArgument(String
         });
     }
 
+    private Function<Object, Stream<AttributeFilter<?>>> mapArrayArgument(String attribute, GraphQLInputObjectType arrayArgType) {
+        return argument -> {
+            Map<String, Object> argMap = (Map<String, Object>) argument;
+
+            // Get the element type from the "contains" field
```
jstastny-cz: what if there's no contains field? Why do we abort, the switch below has also other operators covered, shouldn't we return empty only in the default branch of that switch?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -156,17 +238,221 @@ protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E>
...
+    /**
+     * Checks if the filter is a collection operation (CONTAINS, CONTAINS_ALL, CONTAINS_ANY)
+     * or contains collection attributes that require special NOT handling with subqueries.
+     */
+    private boolean isCollectionNotOperation(AttributeFilter<?> filter) {
+        switch (filter.getCondition()) {
+            case CONTAINS:
+            case CONTAINS_ALL:
+            case CONTAINS_ANY:
+                return true;
```
jstastny-cz:
```suggestion
        switch (filter.getCondition()) {
            case CONTAINS, CONTAINS_ALL, CONTAINS_ANY ->
                return true;
```
Would that work here?

jstastny-cz (changes requested):
Perhaps there's no shortcut and this solution, quite complex, is necessary - I don't dispute that.
At the same time, I see room for making the code more readable, and more in sync with existing coding style.

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: comment addressed changes done

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
GopikaReghunath: changes done to resolve this

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
GopikaReghunath: done

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLInputObjectTypeMapper.java
GopikaReghunath: done

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
GopikaReghunath: comment addressed

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
GopikaReghunath: changes done as per the comment

data-index/data-index-graphql/src/main/java/org/kie/kogito/index/graphql/query/GraphQLQueryMapper.java
GopikaReghunath: changes done as per comment

data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls
```diff
@@ -252,6 +252,28 @@ input NodeInstanceArgument {
...
+input NodeInstanceArrayArgument {
+    # Array-specific operations
+    contains: NodeInstanceArgument
+    containsAll: [NodeInstanceArgument!]
+    containsAny: [NodeInstanceArgument!]
+    isNull: Boolean
+    and: [NodeInstanceArrayArgument!]
+    or: [NodeInstanceArrayArgument!]
+    not: NodeInstanceArrayArgument
+    # Backward compatibility:
```
jstastny-cz, line 264: Backward compatibility of what actually? Do we want to allow queries on arrays that define nested values, e.g. the id below, or? Would it then pick the first, all matching, or what are the semantics?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -156,17 +224,206 @@ protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E>
...
+     * Builds a NOT EXISTS subquery predicate for collection operations.
+     * This ensures entity-level negation instead of per-row negation.
+     */
+    private Predicate buildCollectionNotPredicate(AttributeFilter<?> filter, Root<E> root,
+            CriteriaBuilder builder, CriteriaQuery<?> criteriaQuery) {
+
+        // Special handling for AND/OR: Apply De Morgan's Law
+        // NOT (A AND B) = NOT A OR NOT B
+        // NOT (A OR B) = NOT A AND NOT B
+        if (filter.getCondition() == org.kie.kogito.persistence.api.query.FilterCondition.AND) {
```
jstastny-cz: why isn't this a switch on filter.getCondition() ? Should be possible, right?

data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls
GopikaReghunath: allows filtering on array elements without explicitly using the contains, containsAll, or containsAny operations. Behavior: Returns entities where ANY array element matches the condition
example: 
```
id: "Task-A"
"attachments": [
          {
            "id": "id-1",
            "name": "google"
          }
        ]

id: "Task-B"
"attachments": [
          {
            "id": "id-2",
            "name": "google"
          }
        ]

id: "Task-C"
"attachments": [
          {
            "id": "id-3",
            "name": "comment"
          }
        ]

Query : 
UserTaskInstances(
    where: {
      attachments: {
        name: { equal: "google" }
      }
    }
  )
Return Task-A and Task-B
```
@jstastny-cz let me know if you think this is not needed

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: comment addressed

data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls
jstastny-cz: So basically defaults to `containsAny`? IDK really, if "forcing" users to use proper operators now is a problem, @martinweiler wdyt?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -110,23 +121,80 @@ public List<T> execute() {
         return (List<T>) query.getResultList().stream().map(mapper).collect(toList());
     }
 
-    protected Function<AttributeFilter<?>, Predicate> filterPredicateFunction(Root<E> root, CriteriaBuilder builder) {
+    /**
+     * Determines if DISTINCT is needed based on whether the query uses collection attributes
+     * that would result in JOINs and potential duplicate rows.
+     */
+    private boolean needsDistinct() {
```
jstastny-cz: When/why do we need distinct at all?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -110,23 +121,80 @@ public List<T> execute() {
...
+    }
+
+    protected Function<AttributeFilter<?>, Predicate> filterPredicateFunction(Root<E> root, CriteriaBuilder builder, CriteriaQuery<?> criteriaQuery) {
         return filter -> jsonPredicateBuilder.filter(b -> filter.isJson()).map(b -> b.buildPredicate(filter, root, builder))
-                .orElseGet(() -> buildPredicateFunction(filter, root, builder));
+                .orElseGet(() -> buildPredicateFunction(filter, root, builder, criteriaQuery));
     }
 
-    protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E> root, CriteriaBuilder builder) {
+    protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E> root, CriteriaBuilder builder, CriteriaQuery<?> criteriaQuery) {
```
jstastny-cz, line 132: so basically, each array attribute results in a separate WHERE EXISTS (SELECT 1 FROM ... now the comparisons) ...
How are these EXISTS subqueries joined, is that an OR or AND?

If single query contains filters on multiple array arguments - they should be evaluated together, returning only top-level items matching all conditions - is that the case already?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -156,17 +224,206 @@ protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E>
...
+                filter.getAttribute() != null && isCollectionAttribute(filter.getAttribute());
+            default -> false;
+        };
+    }
+
+    /**
+     * Builds a NOT EXISTS subquery predicate for collection operations.
+     * This ensures entity-level negation instead of per-row negation.
+     */
+    private Predicate buildCollectionNotPredicate(AttributeFilter<?> filter, Root<E> root,
```
jstastny-cz: Why for negations we need whole new handling rather than running WHERE NOT EXISTS () with the relevant subqueries, would then be "just" matter of switching NOT and AND/OR for the combinations, or?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: Multiple filters on array/collection attributes are always combined with AND logic, ensuring only top-level entities matching ALL conditions are returned. Positive filters (CONTAINS, EQUAL) use INNER JOINs, while NOT operations use NOT EXISTS subqueries for proper entity-level negation. For example, the query filtering not: { nodes: { name: { equal: "Technical Interview" } } } and not: { nodes: { type: { equal: "ErrorNode" } } } generates:
`SELECT pie1_0.* 
FROM processes pie1_0 
WHERE NOT EXISTS (
    SELECT 1 FROM processes pie2_0 
    JOIN nodes n1_0 ON pie2_0.id = n1_0.process_instance_id 
    WHERE pie2_0.id = pie1_0.id AND n1_0.name = 'Technical Interview'
) 
AND NOT EXISTS (
    SELECT 1 FROM processes pie3_0 
    JOIN nodes n2_0 ON pie3_0.id = n2_0.process_instance_id 
    WHERE pie3_0.id = pie1_0.id AND n2_0.type = 'ErrorNode'
);
`

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: The handling was added to ensures entity-level negation, rather than incorrectly negating at the row level. Simply wrapping a positive predicate with NOT would fail because it would negate individual JOIN results rather than checking entity-level existence.

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: distinct is not need misunderstood on queries like the below:
 ```
ProcessInstances(
    where: {
      nodes: {
        type: { equal: "ActionNode" }
      }
    }
  )
```
when there are multiple nodes with same type works
Retested the scenario after removing distinct and is is working as expected

File: data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java (hunk 11 above)
Comment (jstastny-cz) [reply data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java:132]: Right, there are 2 subselects each joining the same tables - why not to merge the conditions into a single EXISTS query? And adjust AND/OR combinations accordingly.
Would be
`SELECT ... FROM processes p_0 WHERE NOT EXISTS (SELECT 1 FROM processes p1 JOIN nodes n1 ON p1.id=n1.id WHERE p_0.id=p1.id AND (n1.name='Technical Interview' OR n1.type='ErrorNode`)`?

so in simplified WHERE NOT EXISTS (both conditions via OR), similarly for positive filters, why isn't it the same approach, also single subselect (which results in the join anyways) and then combine the filters in a single set.

Having said that - can't actually be both positive and negative filters handled within the same single subquery using EXISTS by just inverting the conditions for the negative?

Comment (jstastny-cz) [reply data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java:132]: of course, perhaps if the joins are between various tables in each condition, it needs to be a separate subquery.

File: data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls (hunk 8 above)
Comment (martinweiler) [reply data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls:264]: @jstastny-cz as @GopikaReghunath showed in the example, the fallback to `containsAny` is how such queries were handled prior to this PR. The new operators give you more control, but I wouldn't say they are "forcing" you into anything.

**Review by martinweiler — APPROVED** [review]

File: data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java (hunk 11 above)
Comment (GopikaReghunath) [reply data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java:132]: @jstastny-cz I tried merging multiple NOT EXISTS subqueries into a single one but was causing issues with NOT CONTAINS ALL and was giving unexpected response for example:
Task has comment with id 243672d5-44a2-423b-901e-1f4130373ebe
Query: NOT CONTAINS_ALL [243672d5-44a2-423b-901e-1f4130373ebe, another-comment-id]
Expected: Return task (missing another-comment-id) 
this was resulting in a query : `NOT EXISTS(... JOIN c1_0 ... JOIN c2_0 WHERE (c1_0.id=? OR c2_0.id=?))` which return Returns 0 items hence followed `NOT EXISTS(... JOIN c1_0 WHERE c1_0.id=?) OR NOT EXISTS(... JOIN c2_0 WHERE c2_0.id=?)` and for maintaining the uniformity kept the same for other NOT conditions and for positive cases the current implementation reuses the existing getAttributePath(). Let me know if you have more thoughts

File: data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls (hunk 8 above)
Comment (jstastny-cz) [reply data-index/data-index-graphql/src/main/resources/graphql/basic.schema.graphqls:264]: @martinweiler I meant it the other way round actually - should we care about backward compatibility if the behavior was not correct? It brings increased complexity to code we need to maintain - all those extra fields below.

File: data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java (hunk 11 above)
Comment (jstastny-cz) [reply data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java:132]: This is all new code, right? so the fact it was not correctly translating query into SQL might be because of that and the actual logic. I expect that if we try to merge into a single subquery, we need to properly adjust logic in the inner scope of the logical condition. In your example, we'd basically need `NOT EXISTS(... JOIN c1_0 ... JOIN c2_0 WHERE (c1_0.id=? AND c2_0.id=?)` ? instead of OR?
I'd be interested in the resulting SQL Query plan really - if we merge EXISTS clauses into one or we combine several using AND/OR - would it be possible to attempt run EXPLAIN to get the query resulting query plans? If you have such queries from previous runs then you could use those, if not perhaps you could "simulate" only the SQL query without the need to actually implement the changes just to test SQL query plan.

Comment (GopikaReghunath) [reply data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java:132]: This has been addressed as discussed

File: data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java (hunk 12 above)
Comment (GopikaReghunath) [reply data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java]: comment addressed

File: data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java [hunk 13]
```diff
@@ -156,17 +185,145 @@ protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E>
...
+
+        Subquery<Integer> subquery = criteriaQuery.subquery(Integer.class);
+        Root<E> subRoot = subquery.from(entityClass);
+        Join<?, ?> collectionJoin = subRoot.join(collectionName);
+
+        List<Predicate> havingConditions = new ArrayList<>();
+
+        for (AttributeFilter<?> filter : filters) {
+            String property = filter.getAttribute().split("\\.")[1];
+            List<Object> values = (filter.getCondition() == FilterCondition.CONTAINS || filter.getCondition() == FilterCondition.EQUAL)
```
jstastny-cz: Could we perhaps alter AttributeFilter class by adding method that tells us if the value it holds is collection or single value? 
So that then we'd be able to call `filter.isMultipleValues()` or similar? I noticed we have this conditional in 2 places at least. Or perhaps at least add static method to AttributeFilter, that accepts FilterCondition and tells us if expects multiple values or not? With that we could also FilterCondition enum to hold for each operation the boolean and add a method, e.g. IN would expect multiple values, CONTAINS would expect single value, etc. The AttributeFilter's static method then can only invoke FilterCondition#expectsMultipleValues() (or similar) .

jstastny-cz: Perhaps even each multipleValues condition could carry the information of what is the "joining" operation, if AND or OR ... but that's just a tiny detail.

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -156,17 +185,145 @@ protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E>
...
+            }
+        }
+
+        List<Predicate> allPredicates = new ArrayList<>(otherPredicates);
+        for (List<AttributeFilter<?>> groupFilters : groups.values()) {
+            if (groupFilters.size() == 1) {
+                allPredicates.add(buildCollectionPredicate(groupFilters.get(0), root, builder, criteriaQuery, isNegated));
+            } else {
+                allPredicates.add(buildMultiFilterCollectionPredicate(groupFilters, root, builder, criteriaQuery, isAnd, isNegated));
+            }
```
jstastny-cz: what would happen if we call buildMultiFilterCollectionPredicate on groupFilters size=1 ? I mean, don't we have the "single-group" method while it could be done using the other?

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
```diff
@@ -156,17 +185,145 @@ protected final Predicate buildPredicateFunction(AttributeFilter filter, Root<E>
...
+            CriteriaBuilder builder, CriteriaQuery<?> criteriaQuery, boolean isAnd, boolean isNegated) {
+
+        List<AttributeFilter<?>> nestedFilters = (List<AttributeFilter<?>>) filter.getValue();
+
+        java.util.Map<String, List<AttributeFilter<?>>> groups = new java.util.HashMap<>();
+        List<Predicate> otherPredicates = new ArrayList<>();
+
+        for (AttributeFilter<?> f : nestedFilters) {
+            if (f.getAttribute() != null && isCollectionAttribute(f.getAttribute()) &&
+                    (f.getCondition() == FilterCondition.CONTAINS ||
```
jstastny-cz: If we're able to name a method nicely, I'd consider extending FilterCondition enum by a field and method that would return boolean and simplify this if to `f.getCondition().isXY()` .... is it `isCollection()` what we're checking here, i.e. filtering conditions based on fact if they are to be used on collections? Then I'd say it's an obvious candidate for such enum extension, cause it directly describes the value.

jstastny-cz: the same if condition is in the method above too, so I'd say it's a categorization that we want to have explicitly defined in a single place.

jstastny-cz (commented):
This is great progress @GopikaReghunath, inline I am asking if further refinements are feasible, seems to me they might be (they are less complex to what I asked previously) sort of finish touch I'd say. :-)

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: comment addressed

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: comment addressed

data-index/data-index-storage/data-index-storage-jpa-common/src/main/java/org/kie/kogito/index/jpa/storage/JPAQuery.java
GopikaReghunath: buildMultiFilterCollectionPredicate is reusable when groupFilters.size() == 1 hence refactored

jstastny-cz (approved):
Looks good to me, great work on the refactoring and simplification.

@martinweiler you might want to re-review after the changes.

martinweiler (approved):

PR #2333, opened 2026-05-13
[incubator-kie-issues#2311] Handle reconciliation of RETRY or ERROR jobs in a new transaction

### LINKED ISSUE #2311: VertxJobScheduler fails to retry jobs if current transaction is rolled back
Consider the following process:
```
                    --> async node 1
                   |
Start -> parallel gateway   --> async node 2
                   |
                    --> async node 3
```

The engine will create three jobs that trigger in 1ms. When these jobs execute (in different threads), it is possible that one of them fails:
```
2026-05-13 16:08:36.367|WARN||com.arjuna.ats.arjuna|ARJUNA012125: TwoPhaseCoordinator.beforeCompletion - failed for SynchronizationImple< 0:ffffac1b0004:91ff:6a04a1f8:21, org.hibernate.resource.transaction.backend.jta.internal.synchronization.RegisteredSynchronization@1ce8fe >: org.hibernate.exception.GenericJDBCException: could not execute statement [ERROR: current transaction is aborted, commands ignored until end of transaction block] [insert into job_execution_log (event_date,exception_details,exception_message,execution_counter,expiration_time,job_id,node_instance_id,priority,process_instance_id,repeat_interval,repeat_limit,retries,scheduled_id,status,id) values (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)]
...
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "event_types_pk"
  Detail: Key (process_instance_id, event_type)=(8416e04c-4296-42aa-a0b2-1c0395456abe, timerTriggered) already exists.
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2904)
...
	at io.agroal.pool.wrapper.PreparedStatementWrapper.executeUpdate(PreparedStatementWrapper.java:90)
	at org.kie.kogito.persistence.jdbc.GenericRepository.updateWithLock(GenericRepository.java:122)
	at org.kie.kogito.persistence.jdbc.JDBCProcessInstances.update(JDBCProcessInstances.java:90)
	at org.kie.kogito.process.impl.AbstractProcessInstance.syncPersistence(AbstractProcessInstance.java:783)
	at org.kie.kogito.process.impl.AbstractProcessInstance.lambda$executeInWorkflowProcessInstance$0(AbstractProcessInstance.java:636)
	at org.kie.kogito.process.impl.lock.ContextAwareProcessInstanceLockStrategy.lambda$executeOperation$0(ContextAwareProcessInstanceLockStrategy.java:63)
	at org.kie.kogito.process.impl.lock.ProcessInstanceAtomicLockStrategy.executeOperation(ProcessInstanceAtomicLockStrategy.java:98)
	at org.kie.kogito.process.impl.lock.ContextAwareProcessInstanceLockStrategy.executeOperation(ContextAwareProcessInstanceLockStrategy.java:60)
	at org.kie.kogito.process.impl.AbstractProcessInstance.executeInWorkflowProcessInstance(AbstractProcessInstance.java:615)
	at org.kie.kogito.process.impl.AbstractProcessInstance.executeInWorkflowProcessInstanceWrite(AbstractProcessInstance.java:588)
	at org.kie.kogito.process.impl.AbstractProcessInstance.send(AbstractProcessInstance.java:394)
	at org.kie.kogito.app.jobs.integrations.ProcessInstanceJobExecutor.lambda$execute$2(ProcessInstanceJobExecutor.java:58)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at org.kie.kogito.app.jobs.integrations.ProcessInstanceJobExecutor.lambda$execute$1(ProcessInstanceJobExecutor.java:55)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at org.kie.kogito.app.jobs.integrations.ProcessInstanceJobExecutor.lambda$execute$0(ProcessInstanceJobExecutor.java:54)
	at org.kie.kogito.services.uow.TransactionAwareUnitOfWorkExecutor.execute(TransactionAwareUnitOfWorkExecutor.java:36)
	at org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(UnitOfWorkExecutor.java:45)
	at org.kie.kogito.app.jobs.integrations.ProcessInstanceJobExecutor.execute(ProcessInstanceJobExecutor.java:53)
	at org.kie.kogito.app.jobs.impl.VertxJobScheduler.lambda$doExecute$1(VertxJobScheduler.java:547)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.kie.kogito.app.jobs.impl.VertxJobScheduler.doExecute(VertxJobScheduler.java:547)
	at org.kie.kogito.app.jobs.impl.VertxJobScheduler$3.call(VertxJobScheduler.java:416)
	at org.kie.kogito.app.jobs.impl.VertxJobScheduler$3.call(VertxJobScheduler.java:397)
	at io.quarkus.narayana.jta.QuarkusTransactionImpl.callInOurTx(QuarkusTransactionImpl.java:136)
```

As the transaction is marked for rollback, the retry logic in the job executor fails as well and the process gets stuck.

martinweiler:
Closes https://github.com/apache/incubator-kie-issues/issues/2311

Note that the PR does not attempt to resolve the duplicate key constraint violation reported in the issue. Instead, it improves the error handling if one of the job execution threads encounters an issue (which could also be caused by parallel updates of the same process instance). The solution here is to allow the failing transaction to be rolled back, while executing the retry logic in a new transaction.

### DIFF
--- a/jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
+++ b/jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
@@ -47,6 +47,8 @@ import org.kie.kogito.app.jobs.integrations.UserTaskInstanceJobDescriptorMerger;
 import org.kie.kogito.app.jobs.spi.JobContext;
 import org.kie.kogito.app.jobs.spi.JobContextFactory;
 import org.kie.kogito.app.jobs.spi.JobStore;
+import org.kie.kogito.app.jobs.spi.NoOpTransactionRollbackMarker;
+import org.kie.kogito.app.jobs.spi.TransactionRollbackMarker;
 import org.kie.kogito.app.jobs.spi.memory.MemoryJobContextFactory;
 import org.kie.kogito.app.jobs.spi.memory.MemoryJobStore;
 import org.kie.kogito.event.DataEvent;
@@ -114,6 +116,8 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
     private JobSynchronization jobSynchronization;
 
     private JobExceptionDetailsExtractor exceptionDetailsExtractor;
+
+    private TransactionRollbackMarker transactionRollbackMarker;
 
     public class VertxJobSchedulerBuilder implements JobSchedulerBuilder {
 
@@ -213,6 +217,12 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
             return this;
         }
 
+        @Override
+        public JobSchedulerBuilder withTransactionRollbackMarker(TransactionRollbackMarker transactionRollbackMarker) {
+            VertxJobScheduler.this.transactionRollbackMarker = transactionRollbackMarker;
+            return this;
+        }
+
     }
 
     public VertxJobScheduler() {
@@ -242,6 +252,7 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
         this.refreshJobsInterval = 1000L;
         this.retryInterval = 10 * 1000L; // ten seconds
         this.maxRefreshJobsIntervalWindow = 5 * 60 * 1000L; // every 5 minute
+        this.transactionRollbackMarker = new NoOpTransactionRollbackMarker();
     }
 
     @Override
@@ -422,22 +433,23 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
                 } catch (Exception exception) {
                     LOG.trace("Timeout {} with jobId {} will be retried if possible", timerId, jobId, exception);
 
-                    // CRITICAL: Mark the current transaction for rollback FIRST
-                    // This ensures user task JPA changes are rolled back
+                    // Mark the current transaction for rollback
                     markTransactionForRollback();
 
                     JobDetails nextJobDetails = doRetryOrError(jobDetails, exception);
 
-                    // For RETRY and ERROR status, we need to reconcile in a NEW transaction
-                    // because the current transaction is now marked for rollback.
-                    if (nextJobDetails.getStatus() == JobStatus.RETRY || nextJobDetails.getStatus() == JobStatus.ERROR) {
+                    // For RETRY and ERROR status, check if we need to reconcile in a NEW transaction
+                    // Only do this if there's an active transaction that was marked for rollback
+                    if ((nextJobDetails.getStatus() == JobStatus.RETRY || nextJobDetails.getStatus() == JobStatus.ERROR)
+                            && transactionRollbackMarker != null && transactionRollbackMarker.isTransactionActive()) {
                         scheduleReconciliationInNewTransaction(timerId, nextJobDetails);
                     } else {
-                        // For other statuses, reconcile normally
+                        // For other statuses or when no transaction is active, reconcile normally
                         reconcileScheduling(timerId, jobContext, nextJobDetails);
                     }
 
                     jobSchedulerListeners.forEach(l -> l.onFailure(jobDetails));
+
                     return new JobTimeoutExecution(nextJobDetails, exception);
                 }
 
@@ -457,13 +469,10 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
         String jobId = jobDetails.getId();
 
         // Schedule the reconciliation to run asynchronously in a new transaction
-        // Use vertx executeBlocking to run in a worker thread with a new transaction context
         vertx.executeBlocking(promise -> {
             try {
-                // Create a new job context for the new transaction
                 JobContext newJobContext = jobContextFactory.newContext();
 
-                // Create a callable that will be wrapped by transaction interceptors
                 Callable<Void> reconciliationTask = () -> {
                     reconcileScheduling(timerId, newJobContext, jobDetails);
                     return null;
@@ -474,7 +483,6 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
                 for (JobTimeoutInterceptor interceptor : interceptors) {
                     final Callable<Void> currentTask = wrappedTask;
                     wrappedTask = () -> {
-                        // Wrap in a callable that returns Void instead of JobTimeoutExecution
                         interceptor.chainIntercept(() -> {
                             currentTask.call();
                             return new JobTimeoutExecution(jobDetails);
@@ -483,7 +491,6 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
                     };
                 }
 
-                // Execute the wrapped task
                 wrappedTask.call();
 
                 promise.complete();
@@ -713,45 +720,13 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
     }
 
     /**
-     * Marks the current transaction for rollback using reflection to support both Quarkus and Spring Boot.
-     * This ensures that any changes made in the current transaction (like user task JPA operations)
-     * are rolled back, while allowing retry logic to execute in a new transaction.
+     * Marks the current transaction for rollback using the injected TransactionRollbackMarker.
      */
     private void markTransactionForRollback() {
-        try {
-            // Try Quarkus/Narayana first (com.arjuna.ats.jta.TransactionManager)
-            Class<?> tmClass = Class.forName("com.arjuna.ats.jta.TransactionManager");
-            java.lang.reflect.Method getTransactionManagerMethod = tmClass.getMethod("transactionManager");
-            Object tm = getTransactionManagerMethod.invoke(null);
-
-            if (tm != null) {
-                java.lang.reflect.Method getTransactionMethod = tm.getClass().getMethod("getTransaction");
-                Object transaction = getTransactionMethod.invoke(tm);
-
-                if (transaction != null) {
-                    java.lang.reflect.Method setRollbackOnlyMethod = tm.getClass().getMethod("setRollbackOnly");
-                    setRollbackOnlyMethod.invoke(tm);
-                    LOG.debug("Marked transaction for rollback due to exception in job execution (Quarkus/Narayana)");
-                    return;
-                }
-            }
-        } catch (ClassNotFoundException e) {
-            // Narayana not available, try Spring Boot
-            LOG.trace("Narayana TransactionManager not found, trying Spring Boot");
-        } catch (Exception e) {
-            LOG.warn("Failed to mark transaction for rollback using Narayana", e);
-        }
-
-        try {
-            // Try Spring Boot (org.springframework.transaction.support.TransactionSynchronizationManager)
-            Class<?> tsmClass = Class.forName("org.springframework.transaction.support.TransactionSynchronizationManager");
-            java.lang.reflect.Method setRollbackOnlyMethod = tsmClass.getMethod("setRollbackOnly");
-            setRollbackOnlyMethod.invoke(null);
-            LOG.debug("Marked transaction for rollback due to exception in job execution (Spring Boot)");
-        } catch (ClassNotFoundException e) {
-            LOG.trace("Spring TransactionSynchronizationManager not found");
-        } catch (Exception e) {
-            LOG.warn("Failed to mark transaction for rollback using Spring Boot", e);
+        if (transactionRollbackMarker != null) {
+            transactionRollbackMarker.markForRollback();
+        } else {
+            LOG.warn("No TransactionRollbackMarker configured. Transaction rollback marking is not supported.");
         }
     }
 
--- a/jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
+++ b/jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
@@ -437,11 +437,8 @@ public class VertxJobScheduler implements JobScheduler, Handler<Long> {
                     markTransactionForRollbackWhenEnabled();
 
                     JobDetails nextJobDetails = doRetryOrError(jobDetails, exception);
-
                     reconcileScheduling(timerId, jobContext, nextJobDetails);
-
                     jobSchedulerListeners.forEach(l -> l.onFailure(jobDetails));
-
                     return new JobTimeoutExecution(nextJobDetails, exception);
                 }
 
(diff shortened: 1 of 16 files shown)


jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -448,12 +519,36 @@ private void reconcileScheduling(Long timerId, JobContext jobContext, JobDetails
             case RETRY:
                 LOG.trace("Timeout {} with jobId {} will be updated and scheduled", timerId, jobId);
                 addOrUpdateTxTimer(nextJobDetails);
-                jobStore.update(jobContext, nextJobDetails);
+                // Use persist instead of update to handle case where transaction was rolled back
+                // and job record no longer exists in database
+                JobDetails existing = jobStore.find(jobContext, jobId);
+                if (existing != null) {
+                    jobStore.update(jobContext, nextJobDetails);
```
jstastny-cz: just FYI, update is implemented as merge.

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -416,14 +427,29 @@ public JobTimeoutExecution call() throws Exception {
...
+
+                    // Mark the current transaction for rollback
+                    markTransactionForRollback();
+
                     JobDetails nextJobDetails = doRetryOrError(jobDetails, exception);
-                    reconcileScheduling(timerId, jobContext, nextJobDetails);
+
+                    // For RETRY and ERROR status, check if we need to reconcile in a NEW transaction
+                    // Only do this if there's an active transaction that was marked for rollback
+                    if ((nextJobDetails.getStatus() == JobStatus.RETRY || nextJobDetails.getStatus() == JobStatus.ERROR)
```
jstastny-cz: I'd suggest creating boolean method named accordingly instead of complex if conditions and additional comments.

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -416,14 +427,29 @@ public JobTimeoutExecution call() throws Exception {
...
                     JobDetails nextJobDetails = computeAndScheduleNextJobIfAny(executeJobDetails);
-                    reconcileScheduling(timerId, jobContext, nextJobDetails);
+                    reconcileScheduling(timerId, jobContext, nextJobDetails, false);
                     jobSchedulerListeners.forEach(l -> l.onExecution(jobDetails));
                     return new JobTimeoutExecution(nextJobDetails);
                 } catch (Exception exception) {
                     LOG.trace("Timeout {} with jobId {} will be retried if possible", timerId, jobId, exception);
+
+                    // Mark the current transaction for rollback
+                    markTransactionForRollback();
```
jstastny-cz:
```suggestion
                    markTransactionForRollbackWhenEnabled();
```
or
```suggestion
                    markTransactionForRollbackWhenConfigured();
```

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -435,7 +461,52 @@ public JobTimeoutExecution call() throws Exception {
...
+                promise.fail(e);
+            }
+        }, false, result -> {
+            if (result.failed()) {
+                LOG.error("Async reconciliation failed for job {}", jobId, result.cause());
+            }
+        });
+    }
+
+    private void reconcileScheduling(Long timerId, JobContext jobContext, JobDetails nextJobDetails, boolean afterRollback) {
```
jstastny-cz: why not switch the logic and call `scheduleReconciliationInNewTransaction` from here - it would improve encapsulation and reconcileScheduling would still be the main place to tweak scheduling. We're again spreading the scheduling all over the place.
scheduleReconciliationInNewTransaction is not doing the reconciliation anyways, because it is only invoked for subset of states, so IMHO the semantics don't match.

perhaps we need an operation like "scheduleRetry" which would be invoked in both `RETRY` and `ERROR` case in reconcileScheduling switch(they're copy&paste anyways already). Then inside this new method, we'd check if we have the valid case:
```
(nextJobDetails.getStatus() == JobStatus.RETRY || nextJobDetails.getStatus() == JobStatus.ERROR) && transactionRollbackMarker != null && transactionRollbackMarker.isTransactionActive()
```
Now, the JobStatus check would be redundant in the call stack, so much simpler:
```
transactionRollbackMarker != null && transactionRollbackMarker.isTransactionActive()
```
If the condition is true, we'd need to handle the vertx new transaction scheduling, if not, we simply run what we need to. Even further - "what we need to" is your original
```
                JobDetails existingError = jobStore.find(jobContext, jobId);
                if (existingError != null) {
                    jobStore.update(jobContext, nextJobDetails);
                } else {
                    jobStore.persist(jobContext, nextJobDetails);
                }
                // Fire events only if we're reconciling after a rollback
                // (events from original transaction were lost)
                if (afterRollback) {
                    fireEvents(nextJobDetails);
                }
```
And the `if (afterRollback)` part would then be obvious from within the `scheduleRetry`, right? Based on `transactionRollbackMarker != null && transactionRollbackMarker.isTransactionActive()` again.
So in my naive perspective 
```
 Callable<void> scheduleRetryCall = () -> {
                JobDetails existingError = jobStore.find(jobContext, jobId);
                if (existingError != null) {
                    jobStore.update(jobContext, nextJobDetails);
                } else {
                    jobStore.persist(jobContext, nextJobDetails);
                }
          };
          if(transactionRollbackMarker != null && transactionRollbackMarker.isTransactionActive()) {
//                  vertx.executeBlocking with wrap callable from scheduleRetryCall var above.
                     fireEvents(nextJobDetails); // this now is needed unconditionally, right?
          } else {
                     scheduleRetryCall.call();
          }

}
```
Does it make sense?

martinweiler:
@jstastny-cz thank you for your feedback and the improvement suggestions! Much appreciated!

I think they all made sense and I have implemented them accordingly. To me the code is much cleaner now, but if you spot anything else, please let me know.

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -435,6 +453,84 @@ public JobTimeoutExecution call() throws Exception {
...
+    private void scheduleRetry(JobContext jobContext, JobDetails jobDetails) {
+        String jobId = jobDetails.getId();
+        JobStatus status = jobDetails.getStatus();
+
+        // Check if we need to execute in a new transaction
+        if (transactionRollbackMarker != null && transactionRollbackMarker.isTransactionActive()) {
+            // Current transaction will be rolled back, so schedule retry in a new transaction
+            LOG.trace("Job {} with status {} will be scheduled in new transaction", jobId, status);
+
+            vertx.executeBlocking(promise -> {
```
pefernan: I think `executeBlocking(Handler, boolean, AsyncResult)` is deprecated, we should be using 
`executeBlocking(Callable, boolean)`

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -435,6 +453,84 @@ public JobTimeoutExecution call() throws Exception {
...
+                        wrappedTask = () -> {
+                            interceptor.chainIntercept(() -> {
+                                currentTask.call();
+                                return new JobTimeoutExecution(jobDetails);
+                            }).call();
+                            return null;
+                        };
+                    }
+
+                    wrappedTask.call();
```
pefernan: If we switch the callables to  `Callable<JobTimeoutExecution>`

this way we could simplify all this code to something like:
```
                  Callable<JobTimeoutExecution> wrappedTask = () -> {
                        jobStore.update(newJobContext, jobDetails);
                        fireEvents(jobDetails); // Fire events after rollback
                        // Schedule timer in new transaction after persistence
                        if (status == JobStatus.RETRY) {
                            addOrUpdateTxTimer(jobDetails);
                        } else {
                            removeTxTimer(jobDetails);
                        }
                        return new JobTimeoutExecution(jobDetails);
                    };

                    // Apply transaction interceptors to ensure this runs in a new transaction
                    for (JobTimeoutInterceptor interceptor : interceptors) {
                        wrappedTask = interceptor.chainIntercept(wrappedTask);
                    }

                    wrappedTask.call();
```

pefernan (changes requested):

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
pefernan: we could still have the `newTransactionTask ` for better clarity in the code

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -435,6 +453,84 @@ public JobTimeoutExecution call() throws Exception {
...
+                } else {
+                    String mapKey = getMapKey(jobDetails);
+                    TimerInfo timerInfo = jobsScheduled.remove(mapKey);
+                    if (timerInfo != null) {
+                        removeTimerInfo(timerInfo);
+                    }
+                }
+            } catch (Exception e) {
+                LOG.error("Failed to schedule retry for job {}", jobId, e);
+            }
```
jstastny-cz, line 486: How did we get to this? addTimerInfo and removeTimerInfo?
Originally this "part of code" was done via:
```
            case RETRY:
                LOG.trace("Timeout {} with jobId {} will be updated and scheduled", timerId, jobId);
                addOrUpdateTxTimer(nextJobDetails);
                jobStore.update(jobContext, nextJobDetails);
                break;
            case ERROR:
                LOG.trace("Timeout {} with jobId {} will be set to error", timerId, jobId);
                removeTxTimer(nextJobDetails);
                jobStore.update(jobContext, nextJobDetails);
                break;
```
right?
So addOrUpdateTxTimer and removeTxTimer, what's the reason to inline?
What I though was we'd reuse the callable from the other if branch 
```
                    Callable<Void> newTransactionTask = () -> {
                        jobStore.update(newJobContext, jobDetails);
                        fireEvents(jobDetails); // Fire events after rollback
                        // Schedule timer in new transaction after persistence
                        if (status == JobStatus.RETRY) {
                            addOrUpdateTxTimer(jobDetails);
                        } else {
                            removeTxTimer(jobDetails);
                        }
                        return null;
                    };
```
in here. Isn't it possible?

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
martinweiler: Good catch, I don't know how I missed the deprecation warning in my IDE.

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
martinweiler: Good points, thanks!

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
martinweiler: You are right, both branches of the if should be aligned. I guess I went back and forth a couple of times to get the tests working correctly and must have changed this part as well. Thanks for spotting this!

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
```diff
@@ -467,63 +464,38 @@ private void scheduleRetry(JobContext jobContext, JobDetails jobDetails) {
...
                 if (status == JobStatus.RETRY) {
-                    addTimerInfo(jobDetails);
+                    addOrUpdateTxTimer(jobDetails);
                 } else {
-                    String mapKey = getMapKey(jobDetails);
-                    TimerInfo timerInfo = jobsScheduled.remove(mapKey);
-                    if (timerInfo != null) {
-                        removeTimerInfo(timerInfo);
-                    }
+                    removeTxTimer(jobDetails);
```
jstastny-cz: What I actually meant was, reuse the `newTransactionTask` callable in here (after moved to higher level context and renamed perhaps), so that we define this list of steps in a single place?
So in here I'd expect sth like `newTransactionTask.call()`; What do you think? Would we have problems catching the exception in such case? Would it then be more appropriate to create a reusable method out of 
```
                jobStore.update(newJobContext, jobDetails);
                fireEvents(jobDetails); // Fire events after rollback
                // Schedule timer in new transaction after persistence
                if (status == JobStatus.RETRY) {
                    addOrUpdateTxTimer(jobDetails);
                } else {
                    removeTxTimer(jobDetails);
                }
                return new JobTimeoutExecution(jobDetails);
```
and call it both in here an in the `newTransactionTask` ?

I mean, the logic of diverting the execution based on if status is RETRY or not, makes the procedure non-trivial, so my worry is if we'd update appropriately in both places in future.

pefernan (approved):

jobs/jobs-common-embedded/src/main/java/org/kie/kogito/app/jobs/impl/VertxJobScheduler.java
martinweiler: Thanks @jstastny-cz ! I moved the duplicate steps into a new `persistAndSchedule` method, to avoid potential issues in the future in case we have to change this logic. I did not implement your suggestion of reusing `newTransactionTask.call()`  as it would have complicated the exception handling.

jstastny-cz (approved):
Looks good, thank you @martinweiler

PR #2348, opened 2026-07-09
incubator-kie-issues#2307 add process id and version into data-index …

jstastny-cz:
…graphql schema

Follows-up on
* #2337 

Adding missing process version and id relevant fields into data-index graphql schema. Trageting not only type itself, but also input  arguments and orderBy.
This should allow retrieval of the values and filtering based on those on the client side.


martinweiler (approved):
LGTM @jstastny-cz

pefernan (approved):

